PR #463, opened 2022-05-19
fix: correct behavior for CreateApiKey config

alharris-at:
#### Description of changes
Update the behavior where we look at the old CreateAPIKey flag from transformer v1. We had implemented this in v1 to just look for truthy behavior, which isn't consistent with the previous implementation. Explicitly updating this to only disable for value 0.

#### Issue #, if available
https://github.com/aws-amplify/amplify-category-api/issues/80

Ran in my local CLI to verify the 0/1 behavior a few times, and added unit tests for this behavior to ensure we have consistency in how we're handling these parameters.

#### Checklist
- [x] PR description included
- [x] `yarn test` passes


marcvberg (approved):
LGTM, only thing is looking at the old implementation conditional should we add a test or two for literal true/false values?

AaronZyLee (approved):
LGTM 🍨

PR #467, opened 2022-05-23
fix: update cleanup script to correctly use assumed creds in child accounts

alharris-at:
#### Description of changes
There are 2 keys changes in this PR, then a bunch of general cleanup.

1. in `configureAws` which is called before searching for staleResources in child stacks, we were setting our aws creds to...

```
accessKeyId: process.env.AWS_ACCESS_KEY_ID,
secretAccessKey: process.env.AWS_SECRET_ACCESS_KEY,
sessionToken: process.env.AWS_SESSION_TOKEN,
...(process.env.AWS_SESSION_TOKEN ? { sessionToken: process.env.AWS_SESSION_TOKEN } : {}),
```

despite passing in the accessKeyId, etc. So we were never actually using the assumed creds.

2. inside the account->region nested loop we had some code like.
```typescript
for (const region of AWS_REGIONS_TO_RUN_TESTS) {
  amplifyApps.push(...(await getAmplifyApps(region)));
  stacks.push(...(await getStacks(region)));
}
```

Because these are parallel network calls, I rewrote the loop to return a list of promises, then yield all promises at once. This could be taken a lot further, since we have nested awaits inside those top-level functions, but that'd require a more significant refactor of the code.

#### Issue #, if available
N/A, but CI workflows are failing w/ 'Too Many Buckets' or some related error.

Ran locally w/ my own API key for Circle, and verified the output made sense.

#### Checklist
- [x] PR description included

### DIFF
--- a/packages/amplify-e2e-tests/src/cleanup-e2e-resources.ts
+++ b/packages/amplify-e2e-tests/src/cleanup-e2e-resources.ts
@@ -1,3 +1,4 @@
+/* eslint-disable spellcheck/spell-checker, camelcase, jsdoc/require-jsdoc, @typescript-eslint/no-explicit-any */
 import { CircleCI, GitType, CircleCIOptions } from 'circleci-api';
 import { config } from 'dotenv';
 import yargs from 'yargs';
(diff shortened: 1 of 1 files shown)


sundersc (approved):
LGTM.

PR #511, opened 2022-06-06
feature: add support for default generation of query and gsi names for @index directive behind a feature flag

alharris-at:
#### Description of changes
Add support for default generation of query and gsi names for @index directive behind a feature flag.

For customers with the feature flag disabled the following behavior will exist:
* The `name` field on the `@index` directive will now be optional, and default to generating a GSI with name like `by{FieldName}{forEach SortKeyField}And{sortKeyField}{/forEach}.
* query will not be generated unless a `queryField` field is provided on the directive by the user.

For customers with the feature flag enabled:
* The `name` field on the `@index` directive will now be optional, and default to generating a GSI with name like `by{FieldName}{forEach SortKeyField}And{SortKeyField}{/forEach}.
* query will be generated unless a `queryField` field is provided on the directive by the user, either as a string or `null` value, otherwise we will auto-generate one using the following scheme - `{ModelName}By{FieldName}{forEach SortKeyField}And{SortKeyField}{/forEach}.

_N.B. This won't enable me to reference an auto-generated index from an @hasMany, but that seems like an edge case, since you'll probably want to just set the name explicitly if you're defining a hasMany for a secondary index anyway, rather than trying to guess what the generated name will be. That technically WILL be possible once we cut over to the new GQL Compilation pipeline process though, and this becomes a mutate step in the pipeline._

#### Issue #, if available
N/A

Unit tested, and generated in a sample app, also went ahead and added a new e2e test which runs through our existing index tests with the feature flag enabled, and verifies that I can query the auto-generated index.

https://app.circleci.com/pipelines/github/aws-amplify/amplify-category-api/411/workflows/46cb2529-d343-4733-965e-fd9708ef7dc1

#### Checklist
- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-graphql-index-transformer/src/__tests__/amplify-graphql-index-transformer.test.ts
+++ b/packages/amplify-graphql-index-transformer/src/__tests__/amplify-graphql-index-transformer.test.ts
@@ -2,10 +2,21 @@ import { ModelTransformer } from '@aws-amplify/graphql-model-transformer';
 import {
   ConflictHandlerType, GraphQLTransform, SyncConfig, validateModelSchema,
 } from '@aws-amplify/graphql-transformer-core';
+import Template from '@aws-amplify/graphql-transformer-core/lib/transformation/types';
 import { FeatureFlagProvider } from '@aws-amplify/graphql-transformer-interfaces';
 import { expect as cdkExpect, haveResourceLike } from '@aws-cdk/assert';
-import { parse } from 'graphql';
+import { DocumentNode, parse } from 'graphql';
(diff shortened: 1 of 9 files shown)


marcvberg (approved):
LGTM, tests look good

phani-srikar (approved):
Changes look good to me. Do we need the changes to `yarn.lock`? I don't see any dependencies being changed in `package.json`.

alharris-at:
> Changes look good to me. Do we need the changes to `yarn.lock`? I don't see any dependencies being changed in `package.json`.

I'd just rebuilt the file as a normal course to update stale deps, etc.

PR #517, opened 2022-06-06
feat: add error when using an owner field as a sort key field

danielleadams:
#### Description of changes
As discussed, there's a breaking change when using owner `@auth` field as the `@primaryKey`'s sort key. Therefore, the transformer will raise an error and abort the transform process when using the `sub::username` identity claim in which the owner attribute in the database is being used as the sort key.

Docs PR to follow.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-graphql-index-transformer/src/graphql-primary-key-transformer.ts
```diff
@@ -156,6 +159,12 @@ function validate(config: PrimaryKeyDirectiveConfiguration, ctx: TransformerCont
       throw new InvalidDirectiveError(`The primary key on type '${object.name.value}' must reference non-null fields.`);
     }
 
+    if (!validateNotOwnerAuth(sortKeyFieldName, config, ctx)) {
+      throw new InvalidDirectiveError(
+        `The primary key's sort key type '${sortKeyFieldName}' cannot be used as an owner @auth field too. Please user another field for the sort key.`
+      );
+    }
+
```
sundersc, line 167: Could this validation affect any existing customer schema?

packages/amplify-graphql-index-transformer/src/utils.ts
```diff
@@ -50,3 +51,58 @@ export function validateNotSelfReferencing(config: IndexDirectiveConfiguration |
...
+
+      if (name === 'allow' && value === 'owner') {
+        isOwner = true;
+      }
+
+      if (name === 'identityClaim') {
+        hasIdentityClaimField = true;
+
+        if (value === 'sub::username') {
+          usesMultiClaim = true;
```
sachscode, line 89: nit: If possible, can we replace these checks with a switch-case to reduce complexity.

sachscode (approved):
LGTM, just a few comments

packages/amplify-graphql-index-transformer/src/utils.ts
danielleadams: I think a switch case makes this more confusing because sometimes it needs 2 values in the conditional, and sometimes it needs 1.

sachscode (approved):
LGTM

danielleadams:
(Just rebased, so ready for rereview)

packages/amplify-graphql-auth-transformer/src/__tests__/owner-auth.test.ts
```diff
@@ -937,4 +937,171 @@ describe('owner based @auth', () => {
...
+        } as FeatureFlagProvider),
+      });
+
+      expect(() => {
+        transformer.transform(schema);
+      }).toThrow(
+        "The primary key's sort key type 'myOwnerField' cannot be used as an owner @auth field too. Please use another field for the sort key.",
+      );
+    });
+  });
```
danielleadams, line 1106: Originally I had put these tests in the index transformer because the validation seemed to fit better in the sort key validations that run. There was a problem with circular dependencies though during the build, so I moved these tests to the auth transformer to prevent the circular dependency (I believe it was index transformer -> auth -> relational -> index). I can create a follow up task to organize these tests within their module, or move them to the velocity tests.

packages/amplify-graphql-index-transformer/src/utils.ts
```diff
@@ -50,3 +51,60 @@ export function validateNotSelfReferencing(config: IndexDirectiveConfiguration |
...
+  const hasOwnerFieldAsSortKey = dirRules.some(rule => {
+    let isOwner = false;
+    let identityClaimIsSet = false;
+    let ownerFieldIsSet = false;
+    let usesMultiClaim = false;
+    let sortKeyFieldIsAuthField = false;
+
+    (rule as ObjectValueNode).fields?.forEach(field => {
+      const name: string = field.name.value;
+      const { value } = field?.value as StringValueNode;
```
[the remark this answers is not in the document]
alharris-at: If this is really desireable we may need to do `(field?.value as (StringValueNode | undefined))?.value;`

packages/amplify-graphql-index-transformer/src/utils.ts
```diff
@@ -50,3 +51,60 @@ export function validateNotSelfReferencing(config: IndexDirectiveConfiguration |
...
+
+  if (!authDir) return true;
+
+  const dirRules = (authDir.arguments?.find(arg => arg.name.value === 'rules')?.value as ListValueNode)?.values || [];
+  const hasOwnerFieldAsSortKey = dirRules.some(rule => {
+    let isOwner = false;
+    let identityClaimIsSet = false;
+    let ownerFieldIsSet = false;
+    let usesMultiClaim = false;
+    let sortKeyFieldIsAuthField = false;
```
alharris-at, line 73: I'll OOS for this PR, since this isn't yet an established pattern, but I prefer to avoid let wherever possible, and rely instead on returning consistent shapes, and reducing state where necessary. In general for something like this I'd probably refactor to have your inner loop (i.e. `rule.fields?.forEach` turn into a `rule.fields.map(doesRuleHaveOwnerForSortKeyField)`, then just return a `.some` for that mapped value. As written, I'm not sure if it's possible to have `sortKeyFieldIsAuthField` be set to `true` for a single rule, and `usesImplicitOwnerField` for another unrelated rule, and result in a false positive here.

alharris-at: As written, there's some chance for state interaction between rules, which I don't think is by design. If we need rules to 'see' each other, we should make that super explicit, and either preprocess the necessary cross-rule state, or otherwise allow the rule checker to look at it's neighbors rather than relying on global state.

packages/amplify-graphql-index-transformer/src/utils.ts
```diff
@@ -50,3 +51,60 @@ export function validateNotSelfReferencing(config: IndexDirectiveConfiguration |
...
+      if (name === 'ownerField') {
+        ownerFieldIsSet = true;
+
+        if (value === sortKeyField) {
+          sortKeyFieldIsAuthField = true;
+        }
+      }
+    });
+
+    const featureFlagEnabled = ctx.featureFlags.getBoolean('useSubUsernameForDefaultIdentityClaim');
```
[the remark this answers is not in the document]
danielleadams: If this value isn't set in the project, `.getBoolean` will return the default value as configured between "existing" and "new" Amplify apps. (https://github.com/aws-amplify/amplify-cli/blob/master/packages/amplify-cli-core/src/feature-flags/featureFlags.ts#L218)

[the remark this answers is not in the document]
danielleadams: (Also, the feature flags module always errors out if the feature flag is missing from the instance because it acts as a singleton, so it always needs to be set in tests too.)

packages/amplify-graphql-index-transformer/src/utils.ts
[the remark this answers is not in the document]
danielleadams: I'll change this to set `value` as a string and have handle undefined.

packages/amplify-graphql-index-transformer/src/utils.ts
danielleadams: I'm fine with that. This was the best way I could configure for my own readability since the GQL objects can be messy to parse (I always wanted to work on some GQL adapters to integrate the code into the transformers), but I can change the pattern. I'll refactor this.

alharris-at (dismissed):

sachscode (approved):
LGTM

packages/amplify-graphql-relational-transformer/src/__tests__/test-helpers.ts
```diff
@@ -4,11 +4,11 @@ import { FeatureFlagProvider } from '@aws-amplify/graphql-transformer-interfaces
  * Merges passed in feature flags with default feature flags for tests
  */
 export const featureFlags: FeatureFlagProvider = {
-  getBoolean: (value: string): boolean => {
+  getBoolean: (value: string, defaultValue: boolean): boolean => {
     if (value === 'useSubUsernameForDefaultIdentityClaim') {
       return true;
     }
```
sundersc, line 10: Can we move the default values to a new hash map and use that here? In that way, if we have to check for any new flags in the future, we don't have to add multiple if statements in the method.

packages/amplify-graphql-schema-test-library/src/__tests__/index.test.ts
```diff
@@ -66,6 +66,17 @@ function expectToFail(name: string, schema: TransformerSchema, transformer: Grap
 
 function createV2Transformer(options: Partial<Writeable<GraphQLTransformOptions>> = {}): GraphQLTransform {
   options.transformers ??= getV2DefaultTransformerList();
+  options.featureFlags = {
+    getBoolean: (value: string, defaultValue: boolean): boolean => {
+      if (value === 'useSubUsernameForDefaultIdentityClaim') {
+        return true;
+      }
+      return defaultValue;
+    },
```
sundersc, line 75: Same comment as the previous. If possible, please move this to a hash map.

packages/graphql-transformers-e2e-tests/src/__tests__/IndexWithAuthV2WithFF.e2e.test.ts
```diff
@@ -314,22 +309,23 @@ async function createFamilyMember(client: GraphQLClient, parent: string, child:
 async function listFamilyMembers(client: GraphQLClient, args?: Record<string, any>) {
   const result = await client.query(
     `query ListFamilyMembers(
+      $id: ModelIDKeyConditionInput
       $parent: ID
-      $child: ModelIDKeyConditionInput
```
sundersc, line 318: Is there a reason to remove this field from an existing test?

alharris-at (approved):

packages/amplify-graphql-relational-transformer/src/__tests__/test-helpers.ts
alharris-at: We can potentially OOS for this change. I actually added a simple createFeatureFlagProviderWithOverrides util in another PR that can just become our standard to simplify behavior and keep us from reimplementing on a per-test basis. Your call if you think it's blocking or not.

packages/graphql-transformers-e2e-tests/src/__tests__/IndexWithAuthV2WithFF.e2e.test.ts
danielleadams: Yes, `$child` is no longer part of the primary key's query.

packages/amplify-graphql-relational-transformer/src/__tests__/test-helpers.ts
danielleadams: I agree with this in theory and am happy to follow up, but since this isn't an established pattern (and there's a mix of using these mock objects in the current tests) is it okay if consolidating the tests is a follow up? If this is blocking I can fix it, but the reason I had to add this feature flag to each test is because this is how the other tests do it.

sundersc (approved):

PR #532, opened 2022-06-11
fix: create owner field resolvers for all operations

danielleadams:
#### Description of changes
Bug fix: when operations don't include `read` operation with the `sub::username` identity claim, the sub prefix does not get stripped out from the return value, so clients would see `sub::username`.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-graphql-auth-transformer/src/graphql-auth-transformer.ts
+++ b/packages/amplify-graphql-auth-transformer/src/graphql-auth-transformer.ts
@@ -349,7 +349,6 @@ export class AuthTransformer extends TransformerAuthBase implements TransformerA
       // if there is a role that does not have read access on the field then we create a field resolver
       // or there is a relational directive on the field then we should protect that as well
       const readRoles = [...new Set(...READ_MODEL_OPERATIONS.map(op => acm.getRolesPerOperation(op)))];
-      const roleDefinitions = readRoles.map(role => this.roleMap.get(role)!);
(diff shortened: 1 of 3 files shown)


ammarkarachi (approved):

sachscode (approved):
LGTM

alharris-at:
It's not clear to me how moving that declaration line addresses this bug. I don't see anywhere we're mutating the `roles` or `acm` in the intervening code, and the only code to take advantage of it was already behind a feature flag.

packages/amplify-graphql-auth-transformer/src/graphql-auth-transformer.ts
```diff
@@ -402,6 +401,8 @@ export class AuthTransformer extends TransformerAuthBase implements TransformerA
       });
 
       if (context.featureFlags.getBoolean('useSubUsernameForDefaultIdentityClaim')) {
+        const roleDefinitions = acm.getRoles().map(role => this.roleMap.get(role)!);
```
danielleadams, line 404: @alharris-at the code is now getting all the roles instead of only the `readRoles` (see line 352). (variable moving is unrelated) We found a bug that only affected mutation operations because this method was only applied to read/query roles.

alharris-at: ahh yes, it turns out I AM blind. thanks for clarifying.

alharris-at (approved):

PR #573, opened 2022-06-20
Feat: Single Source Read

marcvberg:
#### Description of changes
This adds the capability for but does not enable single source reading on GraphQL schemas
This is intended for GraphQL Transformer V2 without support for V1
This also required pulling apart some of the logic in schema transformation so I can actually instantiate the transformer elsewhere

No validation yet, runtime change that hasn't been enabled. Running tests to make sure I haven't broken any existing functionality when refactoring the code

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/packages/amplify-category-api/src/category-utils/context-util.ts
+++ b/packages/amplify-category-api/src/category-utils/context-util.ts
@@ -1,0 +1,55 @@
+import {
+  $TSAny,
+  $TSContext,
+  AmplifyCategories,
+  pathManager,
+} from 'amplify-cli-core';
+import path from 'path';
+import { PROVIDER_NAME } from '../graphql-transformer/constants';
+
+/**
+ * ContextUtil
+ * Some values are calculated on the basis of the context and options that come
+ * from the Amplify CLI, this class/singleton help calculate and cache those values
+ * for reference
+ */
+export class ContextUtil {
+  private resourceDir: string;
+
+  /**
+   * Get the resource directory as used by the API category for GraphQL
+   * @param context the context from the CLI
+   * @param options the options from the CLI
+   */
+  getResourceDir = async (
+    context: $TSContext,
+    options: $TSAny,
+  ): Promise<string> => {
+    if (this.resourceDir) {
+      return this.resourceDir;
+    }
+    let { resourceDir } = options;
+    const backEndDir = pathManager.getBackendDirPath();
+    const { resourcesToBeCreated, resourcesToBeUpdated } = await context.amplify.getResourceStatus(AmplifyCategories.API);
+    const resources = resourcesToBeCreated.concat(resourcesToBeUpdated);
+    if (!resourceDir) {
+      // There can only be one appsync resource
+      if (resources.length > 0) {
+        const resource = resources[0];
+        if (resource.providerPlugin !== PROVIDER_NAME) {
+          return undefined;
+        }
+        const { category } = resource;
+        const { resourceName } = resource;
+        resourceDir = path.normalize(path.join(backEndDir, category, resourceName));
+      } else {
+        // No appsync resource to update/add
+        return undefined;
+      }
+    }
+    this.resourceDir = resourceDir;
+    return resourceDir;
+  }
+}
+
+export const contextUtil = new ContextUtil();
--- a/packages/amplify-category-api/src/category-utils/context-util.ts
+++ b/packages/amplify-category-api/src/category-utils/context-util.ts
@@ -34,6 +34,10 @@ export class ContextUtil {
     const resources = resourcesToBeCreated.concat(resourcesToBeUpdated);
     if (!resourceDir) {
       // There can only be one appsync resource
+      if (!resources.length) {
+        // No appsync resource to update/add
+        return undefined;
+      }
       if (resources.length > 0) {
         const resource = resources[0];
         if (resource.providerPlugin !== PROVIDER_NAME) {
@@ -42,9 +46,6 @@ export class ContextUtil {
         const { category } = resource;
         const { resourceName } = resource;
         resourceDir = path.normalize(path.join(backEndDir, category, resourceName));
-      } else {
-        // No appsync resource to update/add
-        return undefined;
       }
     }
     this.resourceDir = resourceDir;
--- a/packages/amplify-category-api/src/category-utils/schema-reader.ts
+++ b/packages/amplify-category-api/src/category-utils/schema-reader.ts
@@ -81,12 +81,10 @@ export class SchemaReader {
       this.schemaDocument = parse(fullSchema);
     }
 
-    if (preProcessSchema) {
-      if (!this.preProcessedSchemaDocument) {
-        const transformerOptions = await generateTransformerOptions(context, options);
-        const transform = await buildGraphQLTransformV2(transformerOptions);
-        this.preProcessedSchemaDocument = transform.preProcessSchema(this.schemaDocument);
-      }
+    if (preProcessSchema && !this.preProcessedSchemaDocument) {
+      const transformerOptions = await generateTransformerOptions(context, options);
+      const transform = await buildGraphQLTransformV2(transformerOptions);
+      this.preProcessedSchemaDocument = transform.preProcessSchema(this.schemaDocument);
     }
 
     return preProcessSchema ? this.preProcessedSchemaDocument : this.schemaDocument;
--- a/packages/amplify-category-api/src/category-utils/schema-reader.ts
+++ b/packages/amplify-category-api/src/category-utils/schema-reader.ts
@@ -1,0 +1,96 @@
+import * as fs from 'fs-extra';
+import path from 'path';
+import {
+  DocumentNode,
+  parse,
+} from 'graphql';
+import {
+  $TSAny,
+  $TSContext,
+  ApiCategoryFacade,
+} from 'amplify-cli-core';
+import { buildGraphQLTransformV2 } from '../graphql-transformer/transformer-factory';
+import {
+  SCHEMA_DIR_NAME,
+  SCHEMA_FILENAME,
+} from '../graphql-transformer/constants';
+import { generateTransformerOptions } from '../graphql-transformer/transformer-options-v2';
+import {
+  contextUtil,
+} from './context-util';
+
+/**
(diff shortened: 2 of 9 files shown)


packages/amplify-category-api/src/graphql-transformer/transformer-options-v2.ts
```diff
@@ -0,0 +1,300 @@
...
+  try {
+    const parametersPath = path.join(previouslyDeployedBackendDir, `build/${PARAMETERS_FILENAME}`);
+    const parametersExists = fs.existsSync(parametersPath);
+    if (parametersExists) {
+      const parametersString = await fs.readFile(parametersPath);
+      parameters = JSON.parse(parametersString.toString());
+    }
+    return parameters.S3DeploymentRootKey;
+  } catch (err) {
+    return undefined;
```
sundersc, line 268: Should we handle the file access errors? This seems to be okay to return `undefined` if the file doesn't exist but how about (1) file is locked by some other process or (2) file is not a valid JSON.

sundersc (commented):
few nits, otherwise LGTM.

packages/amplify-category-api/src/graphql-transformer/transformer-options-v2.ts
marcvberg: Possibly, I'm not sure exactly what we'd want to do differently in terms of throwing different errors/error handling, since it would be a DX change. This chunk of code isn't something I wrote, I only moved it, so it won't change existing behavior

alharris-at:
I think this needs to be rebased against main due to the release work that's ongoing. Makes this tough to review.

packages/amplify-category-api/src/category-utils/schema-reader.ts
```diff
@@ -0,0 +1,96 @@
...
+      this.schemaPath = schemaFilePath;
+    } else if (fs.pathExistsSync(schemaDirPath)) {
+      this.schemaPath = schemaDirPath;
+    } else {
+      this.schemaPath = null;
+    }
+    return this.schemaPath;
+  };
+
+  invalidateCachedSchema = (): void => {
```
alharris-at, line 54: Is this method necessary? Do we have a strong usecase to ever invalidate here?

packages/amplify-category-api/src/category-utils/schema-reader.ts
```diff
@@ -0,0 +1,96 @@
...
+        fileContentsList.push(fs.readFile(schemaPath));
+      }
+
+      const bufferList = await Promise.all(fileContentsList);
+      const fullSchema = bufferList.map(buff => buff.toString()).join('\n');
+      this.schemaDocument = parse(fullSchema);
+    }
+
+    if (preProcessSchema) {
+      if (!this.preProcessedSchemaDocument) {
```
alharris-at, line 85: This can just be a single statement.

packages/amplify-category-api/src/category-utils/schema-reader.ts
```diff
@@ -0,0 +1,96 @@
...
+  contextUtil,
+} from './context-util';
+
+/**
+ * SchemaReader is a utility point to consolidate and abstract GraphQL Schema reading
+ * The readSchema method provides a flag to read the un-processed (original) schema
+ * if desired, but by default the intent of the SchemaReader is to use the preProcess
+ * utility of the V2 transformer
+ */
+export class SchemaReader {
```
alharris-at, line 28: Does this class need to be exported if the const is down below?

alharris-at (commented):
Mostly minor comments.

packages/amplify-category-api/src/category-utils/schema-reader.ts
marcvberg: Not necessary at the moment, was just a small add I put in because it was easy and I thought it might be useful. Would you rather just leave it out for now?

packages/amplify-category-api/src/category-utils/schema-reader.ts
marcvberg: Strictly speaking maybe not, but this is just how I've seen the singleton pattern done in TS in general, as well as some other points in the CLI. I think exporting the class as well as the instance might be helpful for linting

packages/amplify-category-api/src/category-utils/schema-reader.ts
marcvberg: Fixed

packages/amplify-category-api/src/category-utils/context-util.ts
```diff
@@ -0,0 +1,56 @@
...
+  private resourceDir: string;
+
+  /**
+   * Get the resource directory as used by the API category for GraphQL
+   * @param context the context from the CLI
+   * @param options the options from the CLI
+   */
+  getResourceDir = async (
+    context: $TSContext,
+    options: $TSAny,
```
phani-srikar, line 26: If you only need `resourceDir` from this, can we avoid starting off with $TSAny and accept only what you need for this method?

packages/amplify-category-api/src/category-utils/schema-reader.ts
```diff
@@ -0,0 +1,94 @@
...
+  private schemaPath: string;
+  private schemaDocument: DocumentNode;
+  private preProcessedSchemaDocument: DocumentNode;
+
+  getSchemaPath = async (
+    context: $TSContext,
+    options: $TSAny,
```
phani-srikar, line 35: same as above. Better typed options would be nice here.

packages/amplify-category-api/src/category-utils/schema-reader.ts
```diff
@@ -0,0 +1,94 @@
...
+    usePreProcessing = true,
+  ): Promise<DocumentNode> => {
+    const preProcessSchema = usePreProcessing && (await ApiCategoryFacade.getTransformerVersion(context) === 2);
+    if (!this.schemaDocument) {
+      const fileContentsList = new Array<Promise<Buffer>>();
+      const schemaPath = await this.getSchemaPath(context, options);
+
+      const stats = fs.statSync(schemaPath);
```
phani-srikar, line 70: There is a possibility that the `schemaPath` could be `null` from the definition of `getSchemaPath`. Wouldn't this method throw in that case?

packages/amplify-category-api/src/graphql-transformer/transformer-options-types.ts
```diff
@@ -0,0 +1,60 @@
...
+/**
+ * Arguments passed into a TransformerFactory
+ * Used to determine how to create a new GraphQLTransform
+ */
+export type TransformerFactoryArgs = {
+  addSearchableTransformer: boolean;
+  authConfig: $TSAny;
+  storageConfig?: $TSAny;
+  adminRoles?: Array<string>;
+  identityPoolId?: string;
```
phani-srikar, line 59: is this not part of the `authConfig`?

packages/amplify-category-api/src/graphql-transformer/transformer-options-v2.ts
```diff
@@ -0,0 +1,300 @@
...
+    .filter(r => {
+      const buildDir = path.normalize(path.join(backEndDir, AmplifyCategories.API, r.resourceName, 'build'));
+      return !fs.existsSync(buildDir);
+    });
+  resources = resources.concat(resourceNeedCompile);
+
+  if (forceCompile) {
+    resources = resources.concat(allResources);
+  }
+  resources = resources.filter(resource => resource.service === 'AppSync');
```
phani-srikar, line 78: might want to convert `AppSync` to a const here

packages/amplify-category-api/src/graphql-transformer/transformer-options-v2.ts
```diff
@@ -0,0 +1,300 @@
...
+  if (flags['no-gql-override']) {
+    return undefined;
+  }
+
+  let { parameters } = options;
+  const { forceCompile } = options;
+
+  // Compilation during the push step
+  const { resourcesToBeCreated, resourcesToBeUpdated, allResources } = await context.amplify.getResourceStatus(AmplifyCategories.API);
+  let resources = resourcesToBeCreated.concat(resourcesToBeUpdated);
```
phani-srikar, line 61: what if there are no resources to be created? Does this handles that case?

phani-srikar (changes requested):

packages/amplify-category-api/src/graphql-transformer/transformer-options-types.ts
marcvberg: unsure, I only moved some existing code into new files as I was trying to make it more manageable. This isn't a change

packages/amplify-category-api/src/graphql-transformer/transformer-options-v2.ts
marcvberg: yeah the utility function returns an empty list in that case, but it's still a list and we can use concat

packages/amplify-category-api/src/category-utils/context-util.ts
marcvberg: unfortunately no, the underlying types are not exported and are managed by the CLI team. Cleaning up the use of context/options is probably a project in itself, so for the time being I'd prefer the approach of centralizing the usage of context/options like this

packages/amplify-category-api/src/category-utils/schema-reader.ts
```diff
@@ -0,0 +1,91 @@
...
+  invalidateCachedSchema = (): void => {
+    this.schemaPath = null;
+    this.schemaDocument = null;
+    this.preProcessedSchemaDocument = null;
+  };
+
+  readSchema = async (
+    context: $TSContext,
+    options: $TSAny,
+    usePreProcessing = true,
```
AaronZyLee, line 59: Just want to confirm that the `schemaReader` is not being used somewhere else as this `usePreProcessing` is true by default.

AaronZyLee (approved):
The refactor LGTM 🍨 . Have some comments below.

packages/amplify-category-api/src/category-utils/schema-reader.ts
marcvberg: it is not, this is all new and I'm adding it without any usage currently

Issue #37, opened 2022-05-03
Owner field not being set automatically with transformer v2


CodySwannGT:
### Before opening, please confirm:

- [X] I have installed the latest version of the Amplify CLI (see above), and confirmed that the issue still persists.
- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-cli/issues?q=is%3Aissue+).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-cli/blob/master/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.
- [X] I have removed any sensitive information from my code snippets and submission.

### How did you install the Amplify CLI?

npm

### Amplify CLI Version

8.1.0

### What operating system are you using?

Mac

### Did you make any manual changes to the cloud resources managed by Amplify? Please describe the changes made.

no

### Amplify Categories

api

### Amplify Commands

Not applicable

### Describe the bug

If I am signed in as a user in a group with read/write/update privileges, the owner field is not automatically set. The specific code (see "MINE"):

````
#if( $util.authType() == "User Pool Authorization" )
  #if( !$isAuthorized )
    #set( $staticGroupRoles = [{"claim":"cognito:groups","entity":"Admins","allowedFields":["id","createdAt","updatedAt","legacyId","odds","oddsCalculated","photoUrl","betType","eventDescription","subject","winAmount","betAmount","collectAmount","additionalInfo","ticketId","bookName","isLive","views","isMobile","creatorId","ticketEventId","ticketOptionId","ticketSportsbookId","ticketUserId","ticketSportId","ticketOriginalOwnerId","ticketLocaleId","hasShipped","providerTicketId","providerId","disableValues","source","result","inPropswapsPossession","reviewed","sport","user","originalOwner","locale","option","event","sportsbook","provider","listings","ticketViews","bets","cachedBets","userTicketsId","userOriginalTicketsId","providerTicketsId","sportTicketsId","localeTicketsId"],"isAuthorizedOnAllFields":true}] )
    #foreach( $groupRole in $staticGroupRoles )
      #set( $groupsInToken = $util.defaultIfNull($ctx.identity.claims.get($groupRole.claim), []) )
      #if( $groupsInToken.contains($groupRole.entity) )
        #if( $groupRole.isAuthorizedOnAllFields )
[...]

PR #580, opened 2022-06-21
Auto-Populate owner field if user is authorized via static auth checks

phani-srikar:
#### Description of changes
The Auth checks in a `create` mutation currently do not set the `ownerField` if the `@model` type has both static group auth and owner auth rules specified on it and the user is successfully authorized via static auth checks. This is reported by customers as a regression from `V1` behavior. 

This PR fixes this issue behind a FF `populateOwnerFieldForStaticGroupAuth` which is set to `false` for existing projects, `true` otherwise. FF PR in CLI repo: https://github.com/aws-amplify/amplify-cli/pull/10630

**Behavioral changes**
When the above FF is `true`, we automatically set the value of `ownerField` if the user is authorized via static group checks in a `create` mutation. This means that if the user is removed from the group, he will still continue to have access to the records he created since he is the owner.

#### Issue #, if available

https://github.com/aws-amplify/amplify-category-api/issues/37

Unit tests: Updated screenshots indicate the change
E2e tests: [Successful run for changed e2e test](https://app.circleci.com/pipelines/github/aws-amplify/amplify-category-api/509/workflows/09283767-1b0d-4a4d-91db-37f4db502917/jobs/13794)

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-graphql-auth-transformer/src/resolvers/helpers.ts
```diff
@@ -251,6 +252,48 @@ export const generateOwnerClaimExpression = (ownerClaim: string, refName: string
   return compoundExpression(expressions);
 };
 
+/**
+ * Sets the value of owner field if the user is already Authorized
+ */
+export const populateOwnerField = (
```
danielleadams, line 258: I think that this field should be something like `generatePopulateOwnerField` so that it's clear that it's generating VTL code.

danielleadams (changes requested):
@phani-srikar Looks good - I think it would be good to add a test scenario with the `AllThree` model too to test `[String]` type in the e2e.

danielleadams (dismissed):

AaronZyLee (commented):
Overall LGTM. Just have one question on the missing scenario in the tests where a user is not authorized by a group but creates an entry where the `owner` is not specified, which results in `sub::username` as the `owner` field.

phani-srikar:
@AaronZyLee added a check in an e2e test

AaronZyLee (commented):
Could you also check the `AllThree` schema with scenario where the owner is not defined in input and the mutation is done by none-admin owner? This case is special because the owner is defined as `[String]`. The current cases only include explicit `owner` in input or `owner = null`. Not sure the latter one is similar to the implicit input case.

danielleadams (dismissed):

alharris-at (approved):

danielleadams:
@phani-srikar I think this needs a rebase. There are commits in this PR that I think are from a previous version of `main`

Issue #628, opened 2022-07-07
Schema with 'Custom Lambda' Conflict Resolution Strategy does not deploy through cloudformation.


alharris-at:
### Before opening, please confirm:

- [X] I have installed the latest version of the Amplify CLI (see above), and confirmed that the issue still persists.
- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-cli/issues?q=is%3Aissue+).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-cli/blob/master/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.
- [X] I have removed any sensitive information from my code snippets and submission.

### How did you install the Amplify CLI?

npm

### If applicable, what version of Node.js are you using?

16.14.0

### Amplify CLI Version

9.1.0

### What operating system are you using?

Mac

### Did you make any manual changes to the cloud resources managed by Amplify? Please describe the changes made.

No

### Amplify Categories

api

### Amplify Commands

add

### Describe the bug

Lambda-based conflict resolution fails to create lambdas correctly with a cfn issue.

```
CREATE_FAILED Todo                                                            AWS::CloudFormation::Stack Wed Jul 06 2022 17:16:26 GMT-0700 (Pacific Daylight Time) Template format error: Unresolved resource dependencies [env] in the Resources block of the template
CREATE_FAILED amplify-repro-dev-171301-apirepro-13QNESAG9IHNU AWS::CloudFormation::Stack Wed Jul 06 2022 17:16:28 GMT-0700 (Pacific Daylight Time) The following resource(s) failed to create: [Todo].                                                 
⠏ Updating resources in the cloud. This may take a few minutes...

CREATE_FAILED               apirepro                 AWS::CloudFormation::Stack Wed Jul 06 2022 17:16:44 GMT-0700 (Pacific Daylight Time) Embedded stack arn:aws:cloudformation:us-west-2:411040628600:stack/amplify-repro-dev-171301-apirepro-13QNESAG9IHNU/dc976920-fd89-11ec-95a3-0a47c5cc8a1f was not successfully created: The following resource(s) failed to create: [Todo]. 
[...]

PR #633, opened 2022-07-08
fix: update api should allow setting NEW lambda conflict resolver

alharris-at:
#### Description of changes
* When a NEW lambda conflict resolver is set up, we're not appropriately reflecting the generated state of the lambda back to cli-inputs.json, leading to a push failure. This refactor allows us to correctly update state.

#### Issue #, if available
* https://github.com/aws-amplify/amplify-category-api/issues/628

Unit test for this use-case, and manually tested with an `amplify-dev update api` .

#### Checklist
- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-category-api/src/__tests__/provider-utils/awscloudformation/cfn-api-artifact-handler.test.ts
+++ b/packages/amplify-category-api/src/__tests__/provider-utils/awscloudformation/cfn-api-artifact-handler.test.ts
@@ -357,4 +357,36 @@ describe('update artifacts', () => {
       },
     ]);
   });
+
+  it('correctly updates the cli-inputs on an update that sets a NEW lambda for conflict resolution', async () =>{
+    jest.spyOn(AppsyncApiInputState.prototype, 'saveCLIInputPayload');
+    jest.spyOn(AppsyncApiInputState.prototype, 'cliInputFileExists').mockReturnValueOnce(true);
+
+    updateRequestStub.serviceModification.conflictResolution = {
+      defaultResolutionStrategy: {
+        type: 'LAMBDA',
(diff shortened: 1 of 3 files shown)


packages/amplify-category-api/src/__tests__/provider-utils/awscloudformation/cfn-api-artifact-handler.test.ts
```diff
@@ -357,4 +357,36 @@ describe('update artifacts', () => {
...
+
+    const { objectContaining, stringContaining } = expect;
+
+    expect(AppsyncApiInputState.prototype.saveCLIInputPayload).toHaveBeenCalledWith(objectContaining({
+      serviceConfiguration: objectContaining({
+        conflictResolution: {
+          defaultResolutionStrategy: {
+            type: 'LAMBDA',
+            resolver: {
+              type: 'EXISTING',
```
phani-srikar, line 384: I see that we're passing in the type as `NEW` in the update request and it's getting stored here as `EXISTING`, is this intentional?

phani-srikar (approved):

packages/amplify-category-api/src/__tests__/provider-utils/awscloudformation/cfn-api-artifact-handler.test.ts
alharris-at: Yeah, the behavior is slightly confusing, but in the call to `updateArtifacts` we actually check if the value is NEW, generate a new lambda, and then update the value to EXISTING w/ the generated values.

marcvberg (approved):
LGTM

PR #673, opened 2022-07-18
Add checks for invalid/non-existent identity claims in owner auth

phani-srikar:
#### Description of changes

This PR adds stricter checks to make sure the `identityClaim`s specified in the `owner` auth rules actually exist in the customer's JWT token before going ahead with authorizing the request. 
Even in the case of default identity claims, we make sure that the customer's JWT is valid by checking for existence of `sub` and (`username` or `cogntio:username`) claims before constructing the multi claim using the format `sub::username` as noted [here](https://docs.amplify.aws/cli/migration/identity-claim-changes/).

#### Behavioral changes
Before these changes, we allowed non-existent custom claims and set the default `__xamznone__` for the owner claim. This is not secure by default as ideally such requests should be `unauthorized`. 
After this change, we will check for existence of valid claims for owner auth rules which means that the records that have owner field set to the default `__xamznone__` will not be inaccessible to the customers with invalid claims.

Relevant Docs PR is [here](https://github.com/aws-amplify/docs/pull/4498).

- Added e2e tests to check for behavior with invalid custom claims
- The default claims should work as expected with valid `sub` and `username`.
- Full suite of e2e tests running successfully can be found [here](https://app.circleci.com/pipelines/github/aws-amplify/amplify-category-api/738/workflows/1f0f0779-917e-4048-ac53-9c0a4a4d2eab).

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-graphql-auth-transformer/src/resolvers/helpers.ts
```diff
@@ -252,6 +248,58 @@ export const generateOwnerClaimExpression = (ownerClaim: string, refName: string
...
+ * Concatenates multiple owner claims if any
+ */
+export const generateOwnerMultiClaimExpression = (ownerClaim: string, refName: string): Expression => {
+  let expression: Expression;
+  const identityClaims = ownerClaim.split(IDENTITY_CLAIM_DELIMITER);
+  const hasMultiIdentityClaims = identityClaims.length > 1;
+
+  if (hasMultiIdentityClaims) {
+    const additionalClaims: string[] = [];
+    identityClaims.forEach((claim, idx) => {
```
alharris-at, line 261: Rather than a foreach, this is potentially more undertstandable if we just write as a map. i.e.

```js
const additionalClaims = [...Array(identityClaims.length).keys()].splice(1).map((i) => `currentClaim${i}`);
return set(ref(refName), raw(`"$${[refName, ...additionalClaims].join(IDENTITY_CLAIM_DELIMITER)}"`));
```

packages/amplify-graphql-auth-transformer/src/resolvers/helpers.ts
```diff
@@ -252,6 +248,58 @@ export const generateOwnerClaimExpression = (ownerClaim: string, refName: string
...
+/**
+ * Generates a check for invalid owner claims
+ */
+export const generateInvalidClaimsCondition = (ownerClaim: string, refName: string): Expression => {
+  const expressions: Expression[] = [];
+  const identityClaims = ownerClaim.split(IDENTITY_CLAIM_DELIMITER);
+  const hasMultiIdentityClaims = identityClaims.length > 1;
+
+  if (hasMultiIdentityClaims) {
+    identityClaims.forEach((claim, idx) => {
```
alharris-at, line 283: Similar to previous comments, but I'd prefer to rewrite this as a map and get rid of the side effecting `push` statements.

packages/amplify-graphql-auth-transformer/src/resolvers/helpers.ts
```diff
@@ -252,6 +248,58 @@ export const generateOwnerClaimExpression = (ownerClaim: string, refName: string
...
+        );
+      }
+    });
+  } else {
+    expressions.push(
+      not(methodCall(ref('util.isNull'), ref(refName)))
+    );
+  }
+
+  return and(expressions);
```
alharris-at, line 300: Does this generate an empty and statement if we don't have multiple claims?

alharris-at (commented):
Overall looks good, just a few comments on the two new utility methods.

packages/amplify-graphql-auth-transformer/src/resolvers/helpers.ts
```diff
@@ -252,6 +248,58 @@ export const generateOwnerClaimExpression = (ownerClaim: string, refName: string
...
+export const generateOwnerMultiClaimExpression = (ownerClaim: string, refName: string): Expression => {
+  let expression: Expression;
+  const identityClaims = ownerClaim.split(IDENTITY_CLAIM_DELIMITER);
+  const hasMultiIdentityClaims = identityClaims.length > 1;
+
+  if (hasMultiIdentityClaims) {
+    const additionalClaims: string[] = [];
+    identityClaims.forEach((claim, idx) => {
+      if (idx > 0) {
+        additionalClaims.push(`$currentClaim${idx}`);
```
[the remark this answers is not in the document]
phani-srikar: I believe initially we only had `ownerClaim` which was then extended to accommodate multiple claims and these are named as `currentClaim`s. The `ownerClaim` is different from `currentClaim`s because when constructing the multi-claim, we update `ownerClaim` to `ownerClaim :: currentClaim0 :: currentClaim1` etc.

packages/amplify-graphql-auth-transformer/src/resolvers/helpers.ts
phani-srikar: For `expressions` of length 1, the `and` just returns the single expression. Updated this method.

alharris-at (dismissed):

alharris-at (approved):

marcvberg (approved):

PR #689, opened 2022-07-20
chore: add support for warning on owner who can reassign owner field

alharris-at:
#### Description of changes
Show a warning on gql-compile when an `owner` has permissions to `update` their referenced owner field. This will apply to userPools and oidc providers, and should work regardless of whether the user is using a generated owner field, provided one, or list of owners.

As a followup I'll be adding a metric to this as well, so we can determine how many customers are encountering this condition, and whether or not we should consider changing the default behavior of ownership.

#### Issue #, if available
N/A

Unit tests, ran `setup-dev` target and verified against a few schemas that I was seeing/not seeing this warning when expected.

#### Checklist
- [x] PR description included
- [x] `yarn test` passes


phani-srikar (approved):
LGTM!

dpilch (approved):

alharris-at:
Confirmed messaging w/ product.

PR #702, opened 2022-07-28
Avoid scan on base table when sync expression is specified on index field

phani-srikar:
#### Description of changes

Adds a check to identify if the current query operation falls outside the delta table TTL window.
Avoids scanning the entire base table but query it when `syncExpression` returns a predicate that maps to a query expression. Relevant [doc](https://docs.amplify.aws/lib/datastore/sync/q/platform/js/#advanced-use-case---query-instead-of-scan).
This improves the performance of the query significantly as called out in the above doc section.

The Delta Sync Table TTL is set to 30 mins by Amplify CLI when conflict detection is enabled by the developer.
Developers however can override this value following the guide [here](https://docs.amplify.aws/cli/graphql/override/#customize-amplify-generated-resources-for-model-directive).
If the delta sync table TTL is overridden, we will use the overridden value in the Sync Query resolver function.

#### Issue #, if available

https://github.com/aws-amplify/amplify-category-api/issues/672

[Successful e2e test run](https://app.circleci.com/pipelines/github/aws-amplify/amplify-category-api/784/workflows/c4d07ec6-0dbb-49eb-a1ad-97db3552189a)

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-e2e-tests/schemas/schema_with_index.graphql
+++ b/packages/amplify-e2e-tests/schemas/schema_with_index.graphql
@@ -1,0 +1,9 @@
+input AMPLIFY {
+  globalAuthRule: AuthRule = { allow: public }
+}
+
+type Song @model {
+    id: ID!
+    name: String!
+    genre: String! @index(name : "byGenre", queryField: "songInfoByGenre")
+}
--- a/packages/amplify-e2e-tests/src/__tests__/graphql-v2/sync_query_datastore.test.ts
+++ b/packages/amplify-e2e-tests/src/__tests__/graphql-v2/sync_query_datastore.test.ts
@@ -1,0 +1,211 @@
+import {
+    initJSProjectWithProfile,
+    deleteProject,
+    amplifyPush,
+} from 'amplify-category-api-e2e-core';
+import { addApiWithBlankSchemaAndConflictDetection, updateApiSchema, getProjectMeta } from 'amplify-category-api-e2e-core';
+import { createNewProjectDir, deleteProjectDir } from 'amplify-category-api-e2e-core';
+import gql from 'graphql-tag';
+import AWSAppSyncClient, { AUTH_TYPE } from 'aws-appsync';
+(global as any).fetch = require('node-fetch');
+
+const projectName = 'syncquerytest';
+describe('Sync query V2 resolver tests', () => {
+    let projRoot: string;
+    let appSyncClient = undefined;
+  
+    beforeEach(async () => {
+      projRoot = await createNewProjectDir(projectName);
+      await initJSProjectWithProfile(projRoot, {
+        name: projectName,
+      });
+      // await addAuthWithDefault(projRoot, {});
+    });
+  
+    afterEach(async () => {
+      await deleteProject(projRoot);
+      deleteProjectDir(projRoot);
+    });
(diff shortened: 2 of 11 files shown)


alharris-at:
Is there any way for us to leverage the configured ttl instead of just estimating? I'd prefer us not to ask customers to use custom resolvers if at all possible.

phani-srikar:
> Is there any way for us to leverage the configured ttl instead of just estimating? I'd prefer us not to ask customers to use custom resolvers if at all possible.

Custom overrides are applied centrally in transformer core [after all the stacks and resolvers are generated](https://github.com/aws-amplify/amplify-category-api/blob/7d288436ac4141f122790096f84cd5aff723d718/packages/amplify-graphql-transformer-core/src/transformation/transform.ts#L302) in memory. Tweaking this process to re-run/update the transformers might overwrite the user specified overrides.

marcvberg (commented):
Re: Al's comment above - it might make more sense given the context to just pull out those values for TTL into regular CFN parameters, then we could reference them from the resolvers and CFN would put the correct value into the resolvers at deploy time

If a customer configures a higher TTL than default then this might be sub-optimal but no necessarily harmful apart from cost, but if they configure a lower TTL than default without updating this then they'd potentially be working with stale data and could get into more serious issues

At a minimum we would need to document this as a new requirement for customers configuring base table TTL

alharris-at:
> Re: Al's comment above - it might make more sense given the context to just pull out those values for TTL into regular CFN parameters, then we could reference them from the resolvers and CFN would put the correct value into the resolvers at deploy time
> 
> If a customer configures a higher TTL than default then this might be sub-optimal but not necessarily harmful apart from cost, but if they configure a lower TTL than default without updating this then they'd potentially be working with stale data and could get into more serious issues
> 
> At a minimum we would need to document this as a new requirement for customers configuring base table TTL. Possibly even a feature flag considering the impact to existing customers

marcvberg (approved):
LGTM

naveenkoduri:
> 
> If a customer configures a higher TTL than default then this might be sub-optimal but not necessarily harmful apart from cost, 
Just want to put a perspective as we were bitten by this in April 2022. We originally had default of 30min and ended up with lot of scans despite having the right index for Selective Sync. The scan cost was exponential, and is proportional to the no. of customers using our product. So, at some point, cost would be been a deal breaker and also not consider Amplify Datastore for other products. Our solution was to bump from 30 min to 27 hours for _DeltaSync_ table. Part of our problem was likely due to our configuration of _syncInterval_ of 15 seconds which we commented out falling back to the default of 86400 seconds.

In short, we cannot afford to Scan as our data is huge and devices only want to sync the data that is applicable to them.

alharris-at (approved):

naveenkoduri:
Hi,  I am working with AWS support, but, thought to leave a comment as a heads up. The changes made with this PR might be missing one use case.  Per the changes made, lastSync is included in the Sync operation if it is not a scan. We do a Selective Sync and our initial sync will use an Index. So, we will not have a value for lastSync for our initial sync to be passed to Sync operation. Hence, it is failing at the client with an error indicating that lastSync must be a number. We do see the transformed template has lastSync=null.

PR #551, opened 2022-06-14
feat(graphql): add runtime filtering support for subscriptions

sundersc:
#### Description of changes
This PR adds support for subscriptions runtime filtering. 

List of changes:
- Adds new filter subscription input type for each model type.
- Adds a new argument `filter` for the generated subscriptions.
- Uses AppSync enhanced filtering to support server side runtime filtering.
- Adds @auth subscriptions support for multiple owners field (`owners: [String]`)

#### Issue #, if available

Fixes https://github.com/aws-amplify/amplify-category-api/issues/372

#### Description of how you validated changes
- Manual test
- Unit test
- e2e test

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are [changed or added](https://github.com/aws-amplify/amplify-cli/blob/master/CONTRIBUTING.md#tests)
- [ ] Relevant documentation is changed or added (and PR referenced)
- [ ] New AWS SDK calls or CloudFormation actions have been added to relevant test and service IAM policies

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #709, opened 2022-08-04
fix(graphql): revert server-side filtering

sundersc:
Revert subscriptions server-side filtering.


packages/amplify-graphql-auth-transformer/src/graphql-auth-transformer.ts
```diff
@@ -397,8 +397,8 @@ export class AuthTransformer extends TransformerAuthBase implements TransformerA
       const subscriptionRoles = acm
         .getRolesPerOperation('listen')
         .map(role => this.roleMap.get(role)!)
-        // for subscriptions we only use static rules or owner rule
-        .filter(roleDef => roleDef.strategy === 'owner' || roleDef.static);
+        // for subscriptions we only use static rules or owner rule where the field is not a list
```
phani-srikar, line 400: There is a call to `removeAuthFieldsFromSubscriptionFilter`  method in line 276. Is that still necessary? [Refer](https://github.com/aws-amplify/amplify-category-api/pull/551/files#diff-0a38964c5525097a10b63bbbbea2f24278afce76a68e59353eaeb7d805372314R275)

packages/amplify-graphql-auth-transformer/src/graphql-auth-transformer.ts
```diff
@@ -397,8 +397,8 @@ export class AuthTransformer extends TransformerAuthBase implements TransformerA
...
+        .filter(roleDef => (roleDef.strategy === 'owner' && !fieldIsList(def.fields ?? [], roleDef.entity!)) || roleDef.static);
       subscriptionFieldNames.forEach(subscription => {
         this.protectSubscriptionResolver(context, subscription.typeName, subscription.fieldName, subscriptionRoles);
```
phani-srikar, line 403: Can you please double-check if the other changes to this file in [the original PR](https://github.com/aws-amplify/amplify-category-api/pull/551/files#diff-0a38964c5525097a10b63bbbbea2f24278afce76a68e59353eaeb7d805372314R955) need to be reverted?

phani-srikar (changes requested):
Looks Good. Just want to double-check if we're reverting everything to not cause other issues in these files: [diff1](https://github.com/aws-amplify/amplify-category-api/pull/551/files#diff-7774187e24faf546f86ef27b23655e41e800c281e8a5347c869c165e37800b2cR66), [diff2](https://github.com/aws-amplify/amplify-category-api/pull/551/files#diff-42344140b2da23dfec0ac9201d277c867ebd0e1e174f692b8ec3d65f89195093R21), [diff3](https://github.com/aws-amplify/amplify-category-api/pull/551/files#diff-23119c362d4c4e9a901c1ec8adcf7b1e3e6f63c48dce25f0297c10abe001ad8eR31). Thanks!

packages/amplify-graphql-auth-transformer/src/graphql-auth-transformer.ts
sundersc: We are keeping the generated types but it would be harmless. We will completely revert the resolver code back.

packages/amplify-graphql-auth-transformer/src/graphql-auth-transformer.ts
sundersc: No need to revert those changes, they don't impact the functionality.

sundersc:
> Looks Good. Just want to double-check if we're reverting everything to not cause other issues in these files: [diff1](https://github.com/aws-amplify/amplify-category-api/pull/551/files#diff-7774187e24faf546f86ef27b23655e41e800c281e8a5347c869c165e37800b2cR66), [diff2](https://github.com/aws-amplify/amplify-category-api/pull/551/files#diff-42344140b2da23dfec0ac9201d277c867ebd0e1e174f692b8ec3d65f89195093R21), [diff3](https://github.com/aws-amplify/amplify-category-api/pull/551/files#diff-23119c362d4c4e9a901c1ec8adcf7b1e3e6f63c48dce25f0297c10abe001ad8eR31). Thanks!

We are completely reverting back the resolvers to a state before server-side filtering but keeping the generated types for now. There is no change in functionality due to this. It will work the same as it was before server-side filtering.

phani-srikar (approved):

AaronZyLee (approved):

PR #712, opened 2022-08-04
fix(graphql): revert subscriptions server-side filtering

sundersc:
Reverts subscriptions server-side filtering feature.


AaronZyLee (approved):
🚢 
I think we need a `create merge commit` to sync the release with new `main`

PR #714, opened 2022-08-05
Fix: detect when sort key added to primary key

marcvberg:
#### Description of changes
We aren't currently detecting and warning users that a GraphQL schema change which adds a sort key to a primary key is a destructive update, or even performing the destructive update.

This adds the sanity check warning, and should go with a amplify-cli PR to actually perform the destructive update when `--allow-destructive-graphql-schema-updates` is provided

#### Issue #, if available
N/A

Debugged to see what a sort key addition looks like, ran `amplify-dev` against a project with a sort key change and without the change

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/packages/amplify-e2e-tests/schemas/iterative-push/add-sort-to-primary-key/final-schema.graphql
+++ b/packages/amplify-e2e-tests/schemas/iterative-push/add-sort-to-primary-key/final-schema.graphql
@@ -1,0 +1,5 @@
+type Blog @model {
+  id: ID! @primaryKey(sortKeyFields: ["name"]
+  name: String!
+  posts: [Post] @connection(keyName: "byBlog1", fields: ["id"])
+}
--- a/packages/amplify-e2e-tests/schemas/iterative-push/add-sort-to-primary-key/final-schema.graphql
+++ b/packages/amplify-e2e-tests/schemas/iterative-push/add-sort-to-primary-key/final-schema.graphql
@@ -1,5 +1,4 @@
 type Blog @model {
-  id: ID! @primaryKey(sortKeyFields: ["name"]
+  id: ID! @primaryKey(sortKeyFields: ["name"])
   name: String!
-  posts: [Post] @connection(keyName: "byBlog1", fields: ["id"])
 }
--- a/packages/amplify-e2e-tests/src/__tests__/schema-iterative-update-5.test.ts
+++ b/packages/amplify-e2e-tests/src/__tests__/schema-iterative-update-5.test.ts
@@ -30,7 +30,7 @@ describe('Schema iterative update - add sort key to primary key', () => {
     const apiName = 'iterativetest1';
 
     const initialSchema = path.join('iterative-push', 'add-sort-to-primary-key', 'initial-schema.graphql');
-    await addApiWithoutSchema(projectDir, { apiKeyExpirationDays: 7, transformerVersion: 1 });
+    await addApiWithoutSchema(projectDir, { apiKeyExpirationDays: 7, transformerVersion: 2 });
     await updateApiSchema(projectDir, apiName, initialSchema);
     await amplifyPush(projectDir);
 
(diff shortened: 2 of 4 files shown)


alharris-at:
Could you add some unit tests around this case as well?

marcvberg:
We don't have any testing around this right now, I'll set up some E2E testing since this code is dependent on diffing between an updated project and its cloud backend

packages/amplify-e2e-tests/schemas/iterative-push/add-sort-to-primary-key/final-schema.graphql
```diff
@@ -0,0 +1,5 @@
+type Blog @model {
+  id: ID! @primaryKey(sortKeyFields: ["name"]
+  name: String!
+  posts: [Post] @connection(keyName: "byBlog1", fields: ["id"])
```
phani-srikar, line 4: Is there any reason you're using V1 (`@connection`) and V2 (`@primaryKey`) directives on the same schema? Can we make it fully V2?

packages/graphql-transformer-core/src/util/sanity-check.ts
```diff
@@ -75,7 +75,9 @@ export const sanityCheckDiffs = (
  */
 export const getCantEditKeySchemaRule = (iterativeUpdatesEnabled = false) => {
   const cantEditKeySchemaRule = (diff: Diff): void => {
-    if (diff.kind === 'E' && diff.path.length === 8 && diff.path[5] === 'KeySchema') {
+    const sortKeyAdded = diff.kind === 'A' && diff.path.length === 6 && diff.path[5] === 'KeySchema' && diff.index === 1;
+    const keySchemaModified = diff.kind === 'E' && diff.path.length === 8 && diff.path[5] === 'KeySchema';
```
phani-srikar, line 79: Does this cover removing the existing sort key?

packages/amplify-e2e-tests/src/__tests__/schema-iterative-update-5.test.ts
```diff
@@ -0,0 +1,41 @@
...
+    const apiName = 'iterativetest1';
+
+    const initialSchema = path.join('iterative-push', 'add-sort-to-primary-key', 'initial-schema.graphql');
+    await addApiWithoutSchema(projectDir, { apiKeyExpirationDays: 7, transformerVersion: 1 });
+    await updateApiSchema(projectDir, apiName, initialSchema);
+    await amplifyPush(projectDir);
+
+    const finalSchema = path.join('iterative-push', 'add-sort-to-primary-key', 'final-schema.graphql');
+    await updateApiSchema(projectDir, apiName, finalSchema);
+    await expect(amplifyPushUpdate(projectDir)).resolves.toThrowError();
```
phani-srikar, line 39: I believe this should work after the changes to CLI?

phani-srikar (changes requested):

packages/amplify-e2e-tests/src/__tests__/schema-iterative-update-5.test.ts
marcvberg: It won't work unless destructive updates are enabled, otherwise it should error

packages/graphql-transformer-core/src/util/sanity-check.ts
marcvberg: it does, I adjusted the variable naming appropriately

packages/amplify-e2e-tests/schemas/iterative-push/add-sort-to-primary-key/final-schema.graphql
marcvberg: woops... bad copy, fixed

AaronZyLee (approved):
LGTM. Nit: A e2e test case for removing the sortKeyField in primary key can be added (The current one is adding the field)

AaronZyLee (approved):

phani-srikar (approved):
LGTM!

PR #744, opened 2022-08-18
Single source schema

marcvberg:
#### Description of changes
re-add capability for single source read

temporarily not validated, adding as a no-op

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/packages/amplify-category-api/src/category-utils/context-util.ts
+++ b/packages/amplify-category-api/src/category-utils/context-util.ts
@@ -1,0 +1,56 @@
+import {
+  $TSAny,
+  $TSContext,
+  AmplifyCategories,
+  pathManager,
+} from 'amplify-cli-core';
+import path from 'path';
+import { PROVIDER_NAME } from '../graphql-transformer/constants';
+
+/**
+ * ContextUtil
+ * Some values are calculated on the basis of the context and options that come
+ * from the Amplify CLI, this class/singleton help calculate and cache those values
+ * for reference
+ */
+export class ContextUtil {
+  private resourceDir: string;
+
+  /**
+   * Get the resource directory as used by the API category for GraphQL
+   * @param context the context from the CLI
+   * @param options the options from the CLI
+   */
+  getResourceDir = async (
+    context: $TSContext,
+    options: $TSAny,
+  ): Promise<string> => {
+    if (this.resourceDir) {
+      return this.resourceDir;
+    }
+    let { resourceDir } = options;
+    const backEndDir = pathManager.getBackendDirPath();
+    const { resourcesToBeCreated, resourcesToBeUpdated } = await context.amplify.getResourceStatus(AmplifyCategories.API);
+    const resources = resourcesToBeCreated.concat(resourcesToBeUpdated);
+    if (!resourceDir) {
+      // There can only be one appsync resource
+      if (!resources.length) {
+        // No appsync resource to update/add
+        return undefined;
+      }
+      if (resources.length > 0) {
+        const resource = resources[0];
+        if (resource.providerPlugin !== PROVIDER_NAME) {
+          return undefined;
+        }
+        const { category } = resource;
+        const { resourceName } = resource;
+        resourceDir = path.normalize(path.join(backEndDir, category, resourceName));
+      }
+    }
+    this.resourceDir = resourceDir;
+    return resourceDir;
+  }
+}
+
+export const contextUtil = new ContextUtil();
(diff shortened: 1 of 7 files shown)


packages/amplify-category-api/src/graphql-transformer/transformer-options-v2.ts
```diff
@@ -0,0 +1,302 @@
...
+
+export const APPSYNC_RESOURCE_SERVICE = 'AppSync';
+
+/**
+ * Use current context to generate the Transformer options for generating
+ * a GraphQL Transformer V2 object
+ * @param context The $TSContext from the Amplify CLI
+ * @param options The $TSAny options config coming from the Amplify CLI
+ */
+export const generateTransformerOptions = async (
```
alharris-at, line 47: Is this logic cloning in from somewhere? This reads like it's a dup of something, but it's not obvious to me where we're copying it from, and how we're keeping them in-sync until the migration is finished.

marcvberg: Yeah this is a bunch of extracted logic from the transform schema method for V2, to be able to produce a transformer we have to get all this data together and we have to be able to instantiate a transformer object for the single source read. I'll look at the refactor again, maybe I can just get this to work

packages/amplify-category-api/src/category-utils/schema-reader.ts
```diff
@@ -0,0 +1,91 @@
...
+      if (stats.isDirectory()) {
+        fs.readdirSync(schemaPath).forEach((fileName) => {
+          fileContentsList.push(fs.readFile(path.join(schemaPath, fileName)));
+        });
+      } else {
+        fileContentsList.push(fs.readFile(schemaPath));
+      }
+
+      const bufferList = await Promise.all(fileContentsList);
+      const fullSchema = bufferList.map((buff) => buff.toString()).join('\n');
```
phani-srikar, line 77: This could potentially still be an empty schema if the `schemaPath` is a directory with no `.graphql` files or empty files right?

packages/amplify-category-api/src/graphql-transformer/transformer-factory.ts
```diff
@@ -231,3 +235,36 @@ const importTransformerModule = (transformerName: string) => {
     throw error;
   }
 };
+
+/**
+ * Use the options generated from 'transformer-options' to create a V2 GraphQL Transform instance
+ * @param opts The options produced by 'transformer-options'
+ * @returns GraphQLTransform A brand new instance of the GraphQL Transform
+ */
+export const buildGraphQLTransformV2 = async (opts: TransformerProjectOptions<TransformerFactoryArgs>): Promise<GraphQLTransform> => {
```
phani-srikar, line 244: nit: Is this method actually building or in other words applying the transformers on the schema? Or just constructing a `GraphQLTransform` object which has another `transform` method inside it. Would consider renaming if it's the later case.

packages/amplify-category-api/src/category-utils/context-util.ts
```diff
@@ -0,0 +1,78 @@
...
+    }
+    resources = resources.filter((resource) => resource.service === APPSYNC_RESOURCE_SERVICE);
+    if (!resourceDir) {
+      // There can only be one appsync resource
+      if (!resources.length) {
+        // No appsync resource to update/add
+        return undefined;
+      }
+      if (resources.length > 0) {
+        const resource = resources[0];
```
phani-srikar, line 64: is this the behavior previously too that we silently pick the first AppSync resource if multiple are available, given that multiple should not exist.

phani-srikar (changes requested):

packages/amplify-category-api/src/category-utils/context-util.ts
marcvberg: Yes, this is the existing logic

packages/amplify-category-api/src/graphql-transformer/transformer-factory.ts
marcvberg: Renamed

packages/amplify-category-api/src/category-utils/schema-reader.ts
marcvberg: Added an error in the case that no file is found

phani-srikar (approved):

packages/amplify-category-api/src/graphql-transformer/transformer-options-v2.ts
marcvberg: Update: I did refactor it and got E2E to pass this time

alharris-at (dismissed):

alharris-at (approved):

phani-srikar (approved):

PR #747, opened 2022-08-18
Fix broken method call from CLI if FF provider not given

marcvberg:
#### Description of changes
Make FF provider input optional in `getAuthDirectiveRules` method to make the call from CLI to not break existing flows

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-graphql-auth-transformer/src/utils/definitions.ts
+++ b/packages/amplify-graphql-auth-transformer/src/utils/definitions.ts
@@ -1,4 +1,5 @@
 import { AppSyncAuthConfiguration } from '@aws-amplify/graphql-transformer-interfaces';
+import { GetArgumentsOptions } from '@aws-amplify/graphql-transformer-core/lib/utils/directive-wrapper';
 
 /**
  * AuthStrategy
@@ -29,6 +30,8 @@ export interface SearchableConfig {
     search: string;
   };
 }
+
+export type GetAuthRulesOptions = GetArgumentsOptions;
 
 /**
  * AuthTransformerConfig
--- a/packages/amplify-graphql-auth-transformer/src/utils/index.ts
+++ b/packages/amplify-graphql-auth-transformer/src/utils/index.ts
@@ -35,7 +35,7 @@ export const splitRoles = (roles: Array<RoleDefinition>): RolesByProvider => ({
 /**
  * returns @auth directive rules
  */
-export const getAuthDirectiveRules = (authDir: DirectiveWrapper, isField = false, featureFlags: FeatureFlagProvider = null): AuthRule[] => {
+export const getAuthDirectiveRules = (authDir: DirectiveWrapper, featureFlags?: FeatureFlagProvider, isField = false): AuthRule[] => {
   const splitReadOperation = (rule: AuthRule): void => {
     const operations: (ModelOperation | 'read')[] = rule.operations ?? [];
     const indexOfRead = operations.indexOf('read', 0);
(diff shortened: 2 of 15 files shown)


packages/amplify-graphql-transformer-core/package.json
```diff
@@ -27,6 +27,7 @@
     "test": "jest"
   },
   "dependencies": {
+    "amplify-cli-core": "^2.9.0",
```
phani-srikar, line 30: This should be in `peerDependencies` ?

phani-srikar (changes requested):

packages/amplify-graphql-auth-transformer/src/utils/index.ts
```diff
@@ -35,7 +35,7 @@ export const splitRoles = (roles: Array<RoleDefinition>): RolesByProvider => ({
 /**
  * returns @auth directive rules
  */
-export const getAuthDirectiveRules = (authDir: DirectiveWrapper, featureFlags: FeatureFlagProvider, isField = false): AuthRule[] => {
+export const getAuthDirectiveRules = (authDir: DirectiveWrapper, isField = false, options: GetAuthRulesOptions = {}): AuthRule[] => {
```
phani-srikar, line 38: I think it's cleaner to have `isField` as part of the options too!

packages/amplify-graphql-auth-transformer/src/utils/index.ts
```diff
@@ -51,7 +51,7 @@ export const getAuthDirectiveRules = (authDir: DirectiveWrapper, featureFlags: F
     }
   };
 
-  const { rules } = authDir.getArguments<{ rules: Array<AuthRule> }>({ rules: [] }, featureFlags);
+  const { rules } = authDir.getArguments<{ rules: Array<AuthRule> }>({ rules: [] }, null, options);
```
phani-srikar, line 54: given the recommendation above, we could just remove the `null` argument from here and include `isField` in options

packages/amplify-graphql-transformer-core/src/utils/directive-wrapper.ts
```diff
@@ -34,15 +38,19 @@ export class DirectiveWrapper {
       arguments: this.arguments.map(arg => arg.serialize()),
     };
   };
-  public getArguments = <T>(defaultValue: Required<T>, featureFlags: FeatureFlagProvider): Required<T> => {
+  public getArguments = <T>(defaultValue: Required<T>, featureFlags?: FeatureFlagProvider, options?: GetArgumentsOptions): Required<T> => {
```
phani-srikar, line 41: do we still need the entire `featureFlags`?

packages/amplify-graphql-transformer-core/src/utils/directive-wrapper.ts
```diff
@@ -34,15 +38,19 @@ export class DirectiveWrapper {
...
     const argValues = this.arguments.reduce(
       (acc: Record<string, any>, arg: ArgumentWrapper) => ({
         ...acc,
         [arg.name.value]: valueFromASTUntyped(arg.value),
       }),
       {},
     );
-    if (featureFlags.getBoolean('shouldDeepMergeDirectiveConfigDefaults', false)) {
+    const deepMergeFlag = 'shouldDeepMergeDirectiveConfigDefaults';
+    const useDeepMerge = featureFlags != null
```
phani-srikar, line 50: I would refactor this to just use the options instead of having 2 sources of truth here

phani-srikar (changes requested):

packages/amplify-graphql-auth-transformer/src/utils/definitions.ts
```diff
@@ -1,4 +1,5 @@
 import { AppSyncAuthConfiguration } from '@aws-amplify/graphql-transformer-interfaces';
+import { GetArgumentsOptions } from '@aws-amplify/graphql-transformer-core/lib/utils/directive-wrapper';
```
phani-srikar, line 2: should not do this again!

PR #755, opened 2022-08-22
test(graphql): add subscription tests with variable syntax

sundersc:
Add subscription tests with variable syntax.

- Pass owner argument with variable syntax.
- Use variable syntax but do not pass optional argument value.


phani-srikar:
I see a lot of commits unrelated to the change in this PR. Can you please update your fork with latest remote `main`? Thanks.

phani-srikar (approved):

alharris-at (approved):

PR #772, opened 2022-08-30
Deep directive merge

marcvberg:
#### Description of changes
This allows public subscriptions to be generated on models, and adds use of a new feature flag to deep merge directive defaults. For a description of the behavior, see https://github.com/aws-amplify/amplify-cli/pull/10597

#### Issue #, if available
internally tracked

Unit testing and sample application

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/packages/amplify-graphql-auth-transformer/src/graphql-auth-transformer.ts
+++ b/packages/amplify-graphql-auth-transformer/src/graphql-auth-transformer.ts
@@ -9,6 +9,7 @@ import {
   TransformerResolver,
   getTable,
   getKeySchema,
+  generateGetArgumentsInput,
 } from '@aws-amplify/graphql-transformer-core';
 import {
   DataSourceProvider,
@@ -31,6 +32,7 @@ import {
   ListValueNode,
   StringValueNode,
 } from 'graphql';
+import { merge } from 'lodash';
 import { SubscriptionLevel, ModelDirectiveConfiguration, removeSubscriptionFilterInputAttribute } from '@aws-amplify/graphql-model-transformer';
 import {
   getBaseType,
@@ -170,9 +172,8 @@ export class AuthTransformer extends TransformerAuthBase implements TransformerA
     if (context.metadata.has('joinTypeList')) {
       isJoinType = context.metadata.get<Array<string>>('joinTypeList')!.includes(typeName);
     }
-    this.rules = getAuthDirectiveRules(new DirectiveWrapper(directive), {
-      deepMergeArguments: context.featureFlags.getBoolean('shouldDeepMergeDirectiveConfigDefaults'),
-    });
+    const getAuthRulesOptions = merge({ isField: true }, generateGetArgumentsInput(context.featureFlags));
+    this.rules = getAuthDirectiveRules(new DirectiveWrapper(directive), getAuthRulesOptions);
 
     // validate rules
     validateRules(this.rules, this.configuredAuthProviders, def.name.value);
@@ -227,10 +228,8 @@ export class AuthTransformer extends TransformerAuthBase implements TransformerA
     const modelDirective = parent.directives?.find(dir => dir.name.value === 'model');
     const typeName = parent.name.value;
     const fieldName = field.name.value;
-    const rules: AuthRule[] = getAuthDirectiveRules(new DirectiveWrapper(directive), {
-      isField: true,
-      deepMergeArguments: context.featureFlags.getBoolean('shouldDeepMergeDirectiveConfigDefaults'),
-    });
+    const getAuthRulesOptions = merge({ isField: true }, generateGetArgumentsInput(context.featureFlags));
+    const rules: AuthRule[] = getAuthDirectiveRules(new DirectiveWrapper(directive), getAuthRulesOptions);
     validateFieldRules(new DirectiveWrapper(directive), isParentTypeBuiltinType, modelDirective !== undefined, field.name.value, context.featureFlags);
     validateRules(rules, this.configuredAuthProviders, field.name.value);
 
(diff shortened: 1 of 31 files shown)


phani-srikar (changes requested):
Two comments:
- The build is failing because the def of `getArguments` accepts `options` but not FF provider  and you're padding the later from several places. Please check build-test and also cloud e2e. 
- We should either accept the `options` or the FF provider (I prefer the first one) from changed methods in V2 core, to avoid 2 sources of truth. In order to make the construction of these options easier & centralized, you could create a util function in V2 core that accepts the FF provider and gives back the options. This util function is meant to be used from V2 transformers and not from CLI, so from CLI invocation, we will still pass in the `options`.

marcvberg:
@phani-srikar Running tests against this now, I thought I had fixes for these things in the other branch I tried to merge in locally but I guess the fixes didn't make it for some reason

packages/amplify-graphql-auth-transformer/src/utils/schema.ts
```diff
@@ -65,20 +66,20 @@ export const getModelConfig = (directive: DirectiveNode, typeName: string, isDat
       createdAt: 'createdAt',
       updatedAt: 'updatedAt',
     },
-  });
+  }, { deepMergeArguments: featureFlags.getBoolean('shouldDeepMergeDirectiveConfigDefaults', false) });
```
phani-srikar, line 69: It would be nice to have a utility function in the V2 core that generates defaults for these `GetArgumentsOptions` so that we can refactor this to something like `getDefaultArgumentsOptions(featureFlags)`. This utility method will be only used by V2 transformers and provides a way to centrally handle this. You can even specify the name of FF i.e `shouldDeepMergeDirectiveConfigDefaults` as a const there.

packages/amplify-graphql-auth-transformer/src/graphql-auth-transformer.ts
```diff
@@ -170,7 +170,9 @@ export class AuthTransformer extends TransformerAuthBase implements TransformerA
     if (context.metadata.has('joinTypeList')) {
       isJoinType = context.metadata.get<Array<string>>('joinTypeList')!.includes(typeName);
     }
-    this.rules = getAuthDirectiveRules(new DirectiveWrapper(directive));
+    this.rules = getAuthDirectiveRules(new DirectiveWrapper(directive), {
+      deepMergeArguments: context.featureFlags.getBoolean('shouldDeepMergeDirectiveConfigDefaults'),
```
phani-srikar, line 174: default to false, if doesn't exist?

phani-srikar (changes requested):

packages/amplify-graphql-auth-transformer/src/graphql-auth-transformer.ts
marcvberg: Fixed

packages/amplify-graphql-auth-transformer/src/utils/schema.ts
marcvberg: Done

packages/amplify-graphql-auth-transformer/src/utils/definitions.ts
```diff
@@ -1,4 +1,5 @@
 import { AppSyncAuthConfiguration } from '@aws-amplify/graphql-transformer-interfaces';
+import {GetArgumentsOptions} from "@aws-amplify/graphql-transformer-core/lib/utils/directive-wrapper";
```
phani-srikar, line 2: Can you please change this to a non-lib import?

packages/amplify-graphql-auth-transformer/src/utils/schema.ts
```diff
@@ -28,6 +29,7 @@ import {
 import md5 from 'md5';
 import { RELATIONAL_DIRECTIVES } from './constants';
 import { RelationalPrimaryMapConfig, RoleDefinition, SearchableConfig } from './definitions';
+import { generateGetArgumentsInput } from '@aws-amplify/graphql-transformer-core/lib/utils/directive-wrapper';
```
phani-srikar, line 32: change to non-lib import

packages/amplify-graphql-model-transformer/src/graphql-model-transformer.ts
```diff
@@ -398,13 +428,15 @@ export class ModelTransformer extends TransformerModelBase implements Transforme
...
               throw new Error('Unknown subscription field type');
           }
-          resolver.addToSlot(
-            'postAuth',
-            MappingTemplate.s3MappingTemplateFromString(
-              generateAuthExpressionForSandboxMode(context.sandboxModeEnabled),
-              `${subscription.typeName}.${subscription.fieldName}.{slotName}.{slotIndex}.req.vtl`,
-            ),
-          );
+          if (subscriptionLevel === SubscriptionLevel.on) {
```
phani-srikar, line 431: is there a reason we're checking `equals to on` here and in the above `if` conditional is changed to check for `not equals to off`? Isn't this always `on` based on the above check?

phani-srikar (changes requested):
Mostly looks good. Just called out few left out lib imports and a question.

packages/amplify-graphql-model-transformer/src/graphql-model-transformer.ts
marcvberg: public is also an option in addition to off/on, which is why I specify to avoid getting authorization on those public subscriptions

packages/amplify-graphql-auth-transformer/src/utils/definitions.ts
```diff
@@ -1,4 +1,5 @@
 import { AppSyncAuthConfiguration } from '@aws-amplify/graphql-transformer-interfaces';
+import { GetArgumentsOptions } from '@aws-amplify/graphql-transformer-core/lib/utils/directive-wrapper';
```
phani-srikar, line 2: nit: there is one more lib import here

phani-srikar (approved):

alharris-at (approved):

PR #784, opened 2022-09-07
chore(amplify-graphql-searchable-transformer): refactor the expression to construct aggregate values

phani-srikar:
#### Description of changes

These changes refactor a step in the Request Mapping Template that is generated by the Searchable V2 transformer.
The step is responsible for constructing a Map that contains the Aggregate values that will eventually be used in the construction of the Search Query. The current refactor dynamically generates this Map instead of a raw JSON representation. 
This is necessary for the Searchable Mocking to work since the mock VTL compiler does not read raw JSON representations.

The test snapshot updates show the change in the Request Mapping Template.
Added explicit checks as well.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


AaronZyLee (approved):
The factor LGTM. Just one question on the raw JSON output. What the form will be after the refactor if it is not raw JSON?

phani-srikar:
> The factor LGTM. Just one question on the raw JSON output. What the form will be after the refactor if it is not raw JSON?

It'll be the same map object but generated dynamically.

marcvberg:
Can we run E2E against this just for sanity?

marcvberg (approved):
LGTM

AaronZyLee (approved):

Issue #806, opened 2022-09-17
DataStore initial selective sync of models with GSI fail when created with Amplify CLI 10.0.0


oogatta:
### Before opening, please confirm:

- [X] I have installed the latest version of the Amplify CLI (see above), and confirmed that the issue still persists.
- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-cli/issues?q=is%3Aissue+).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-cli/blob/master/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.
- [X] I have removed any sensitive information from my code snippets and submission.

### How did you install the Amplify CLI?

npm

### If applicable, what version of Node.js are you using?

v16.17.0

### Amplify CLI Version

10.0.0

### What operating system are you using?

Mac

### Did you make any manual changes to the cloud resources managed by Amplify? Please describe the changes made.

no

### Amplify Categories

api

### Amplify Commands

add

### Describe the bug

After upgrading Amplify CLI to 10.0.0, initial selective sync of models with GSI fail. I can confirm this behavior on a sample Todo model generated (and pushed) with `amplify add api`.

Also I can confirm that this does not happen on a Todo model generated with Amplify CLI 9.2.1.

**AppSync /graphql response**

```json
{
    "data": {
        "syncTodos": null
    },
    "errors": [
        {
            "path": [
                "syncTodos"
            ],
            "data": null,
            "errorType": "MappingTemplate",
            "errorInfo": null,
            "locations": [
                {
                    "line": 2,
                    "column": 3,
                    "sourceName": null
                }
            ],
            "message": "Value for field '$[lastSync]' must be a number."
        }
    ]
}
```

### Expected behavior

Syncs succeed and DataStore's `isReady` event emits.

### Reproduction steps

1. amplify init (with 10.0.0 CLI)
2. amplify add api with the "Auto Merge" conflict resolution strategy.
3. edit schema.graphql to append `@index`
4. amplify push
5. configure selective sync (below is the code I'm using to sync selectively)

**sample vue component**
```js
<template>
  <p>hi</p>
</template>

<script lang="ts">
import {Amplify, DataStore, Hub, syncExpression} from 'aws-amplify'
import {Todo} from "~/src/models";
import awsmobile from "~/src/aws-exports";

export default {
  async mounted() {
[...]

PR #811, opened 2022-09-19
Set default lastSync correctly

phani-srikar:
#### Description of changes

The default value for `lastSync` should be set to `null` if it doesn't exist in the input arguments. However, currently it is set to "null" string because of converting it to JSON twice, which is being wrongly interpreted in the Sync Query request. This change fixes it.

#### Issue #, if available

https://github.com/aws-amplify/amplify-category-api/issues/806

Using the customer provided schema in the issue and a sample app.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-graphql-index-transformer/src/__tests__/__snapshots__/amplify-graphql-index-transformer.test.ts.snap
+++ b/packages/amplify-graphql-index-transformer/src/__tests__/__snapshots__/amplify-graphql-index-transformer.test.ts.snap
@@ -3681,7 +3681,7 @@ $util.qr($PkMap.put('createdAt' , 'ByCreatedAt'))
   \\"version\\": \\"2018-05-29\\",
   \\"operation\\": \\"Sync\\",
   \\"limit\\": $limit,
-  \\"lastSync\\": $util.toJson($util.defaultIfNull($ctx.args.lastSync, null)),
+  \\"lastSync\\": $util.defaultIfNull($ctx.args.lastSync, null),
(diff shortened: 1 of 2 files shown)


alharris-at:
Is there an e2e test that would have caught this/should be added to protect against regressions?

phani-srikar:
We have an [e2e test](https://github.com/aws-amplify/amplify-category-api/blob/adc5a4b328f0932fbeb2306b8b3e87b27935ccb6/packages/amplify-e2e-tests/src/__tests__/graphql-v2/sync_query_datastore.test.ts#L54) that tests this scenario by setting the `lastSync` to `null` as seen from [here](https://github.com/aws-amplify/amplify-category-api/blob/adc5a4b328f0932fbeb2306b8b3e87b27935ccb6/packages/amplify-e2e-tests/src/__tests__/graphql-v2/sync_query_datastore.test.ts#L181). This test passed previously. However, here we are constructing the GQL query ourselves versus DataStore constructing that Sync query. So, in short we need a test with DS enabled in sample app to test it fully end-to-end.

alharris-at (approved):

marcvberg (approved):
LGTM

GuruprasadHebbar:
When this fix available. same issue we are encountered. 

waiting for the fix

GuruprasadHebbar:
its really urgent, its impacting our delivery. and it is very important.  can you please provide alternate approach or provide the fix. 

We have delivery tomorrow.

PR #817, opened 2022-09-20
fix(relation-transformer): relation field nullability in object and input type

AaronZyLee:
#### Description of changes
Refactor the impact of non-nullable field with relation directive on the generated connection fields in object definition and input types.

- Connection fields(foreign keys) will always be nullable in the `UpdateInput` type, which is consistent to other non-nullable fields.

#### Issue #, if available
Fix https://github.com/aws-amplify/amplify-codegen/issues/461

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-graphql-relational-transformer/src/__tests__/amplify-graphql-belongs-to-transformer.test.ts
+++ b/packages/amplify-graphql-relational-transformer/src/__tests__/amplify-graphql-belongs-to-transformer.test.ts
@@ -617,7 +617,7 @@ describe('@belongsTo connection field nullability tests', () => {
   });
 
   describe('@belongsTo with @hasMany', () => {
-    test('Should generate nullable connection fields in type definition and create/update input when belongsTo field is nullable', () => {
+    test('Should generate nullable connection fields in type definition and create/update input when hasMany field is nullable', () => {
       const inputSchema = `
         type Todo @model {
           todoid: ID! @primaryKey(sortKeyFields:["name"])
@@ -630,7 +630,7 @@ describe('@belongsTo connection field nullability tests', () => {
           taskid: ID! @primaryKey(sortKeyFields:["name"])
           name: String!
           description: String
-          todo: Todo @belongsTo
+          todo: Todo! @belongsTo
         }
       `;
       const transformer = new GraphQLTransform({
@@ -681,20 +681,20 @@ describe('@belongsTo connection field nullability tests', () => {
       expect(updateInputConnectedField2.type.name.value).toBe('String');
     });
   
-    test('Should generate non-nullable connection fields in type definition and create input while keeping nullable in update input when belongsTo field is non-nullable', () => {
+    test('Should generate non-nullable connection fields in type definition and create input while keeping nullable in update input when hasMany field is non-nullable', () => {
       const inputSchema = `
         type Todo @model {
           todoid: ID! @primaryKey(sortKeyFields:["name"])
           name: String!
           title: String!
           priority: Int
-          tasks: [Task] @hasMany
+          tasks: [Task]! @hasMany
         }
         type Task @model {
           taskid: ID! @primaryKey(sortKeyFields:["name"])
           name: String!
           description: String
-          todo: Todo! @belongsTo
+          todo: Todo @belongsTo
         }
       `;
       const transformer = new GraphQLTransform({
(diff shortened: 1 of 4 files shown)


packages/amplify-graphql-relational-transformer/src/schema.ts
```diff
@@ -566,21 +578,22 @@ const updateInputWithConnectionFields = (
   primaryKeyConnectionFieldType: string,
   field: FieldDefinitionNode,
   sortKeyFields: FieldDefinitionNode[],
+  isConnectionFieldsNonNull: boolean = false,
```
alharris-at, line 581: This may be a case where we want to be explicit, rather than default to false.

packages/amplify-graphql-relational-transformer/src/schema.ts
```diff
@@ -262,6 +268,13 @@ export const ensureHasManyConnectionField = (
   const relatedTypeObject = ctx.output.getType(relatedType.name.value) as ObjectTypeDefinitionNode;
   const connectionAttributeName = getConnectionAttributeName(ctx.featureFlags, object.name.value, field.name.value, connectionFieldName);
 
+  // The nullabilty of connection fields for hasMany depends on the other side belongsTo field
+  // Whereas in update input, they are always optional
+  const isConnectionFieldsNonNull = relatedTypeObject.fields?.some(field =>
+    getBaseType(field.type) === object.name.value
+    && field.directives?.some(d => d.name.value === 'belongsTo')
+    && isNonNullType(field.type)) ?? false;
```
phani-srikar, line 276: If the `@belongsTo` does not exist on the connected model, shouldn't we default to the nullability of the field, i.e `isNonNullType(field.type)` instead of `false`?

packages/amplify-graphql-relational-transformer/src/__tests__/amplify-graphql-belongs-to-transformer.test.ts
```diff
@@ -472,3 +473,277 @@ test('Should not resolve to secondary index of connected model if the index is d
...
+      expect(updateInputConnectedField1.type.name.value).toBe('ID');
+      const updateInputConnectedField2 = updateInput.fields.find((f: any) => f.name.value === connectionFieldName2);
+      expect(updateInputConnectedField2).toBeDefined();
+      expect(updateInputConnectedField2.type.kind).toBe(Kind.NAMED_TYPE);
+      expect(updateInputConnectedField2.type.name.value).toBe('String');
+    });
+  });
+
+  describe('@belongsTo with @hasMany', () => {
+    test('Should generate nullable connection fields in type definition and create/update input when belongsTo field is nullable', () => {
```
phani-srikar, line 620: Is this test is repeated twice since I see the same schema and description [here](https://github.com/aws-amplify/amplify-category-api/pull/817/files#diff-acd0902518c587805b3f7217654188e3cc323f5dc9500f146db4f3673109b1f6R490)

packages/amplify-graphql-relational-transformer/src/__tests__/amplify-graphql-has-many-transformer.test.ts
```diff
@@ -850,3 +851,78 @@ describe('Pre Processing Has Many Tests', () => {
...
+    expect(relatedField2).toBeDefined();
+    expect(relatedField2.type.kind).toBe(Kind.NAMED_TYPE);
+    expect(relatedField2.type.name.value).toBe('String');
+    //Create Input
+    const createInput = schema.definitions.find((def: any) => def.name && def.name.value === 'CreateTaskInput') as any;
+    expect(createInput).toBeDefined();
+    expect(createInput.fields.length).toEqual(5);
+    const createInputConnectedField1 = createInput.fields.find((f: any) => f.name.value === connectionFieldName1);
+    expect(createInputConnectedField1).toBeDefined();
+    expect(createInputConnectedField1.type.kind).toBe(Kind.NAMED_TYPE);
```
phani-srikar, line 909: Double check this test if we should return optional types for connected fields in case of a unidirectional `@hasMany` attached to a required field.

packages/amplify-graphql-relational-transformer/src/__tests__/amplify-graphql-belongs-to-transformer.test.ts
AaronZyLee: The first one is `hasOne` and this one is `hasMany`

packages/amplify-graphql-relational-transformer/src/schema.ts
AaronZyLee: I think this might need discussion on whether set the default to false and this is out of the scope of resolving the mentioned codegen issue. I will keep it as the original one as `isNonNullType(field.type)`

packages/amplify-graphql-relational-transformer/src/schema.ts
```diff
@@ -262,6 +268,9 @@ export const ensureHasManyConnectionField = (
   const relatedTypeObject = ctx.output.getType(relatedType.name.value) as ObjectTypeDefinitionNode;
   const connectionAttributeName = getConnectionAttributeName(ctx.featureFlags, object.name.value, field.name.value, connectionFieldName);
 
+  // The nullabilty of connection fields for hasMany depends on the hasMany field
```
AaronZyLee, line 271: I keep this as the original implementation. This is out of the scope of fixing the issue.

phani-srikar (approved):

alharris-at (approved):

PR #830, opened 2022-09-28
test: add e2e tests for sync expression behavior

phani-srikar:
#### Description of changes

Add e2e tests that cover [this advanced use-case](https://docs.amplify.aws/lib/datastore/sync/q/platform/js/#advanced-use-case---query-instead-of-scan) where a Sync query using GSI will query the base table using that GSI instead of a scan operation.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


alharris-at (approved):

phani-srikar:
From my discussion with Ivan, there is no easy way to identify that unless we have logging enabled for the resolvers. However, the tests added here cover the part of the resolver that is known to generate a Query expression like using filters on GSI.

PR #860, opened 2022-10-06
feat: migrate index and model transformers to CDK v2

sobolk:
#### Description of changes

This PR makes the following packages compile with CDK v2.
- `@aws-amplify/graphql-index-transformer`
- `@aws-amplify/graphql-model-transformer`
- `@aws-amplify/graphql-transformer-core`
- `@aws-amplify/graphql-transformer-interfaces`

#### Description of how you validated changes

This PR compiles and passes `yarn test`, however we won't be able to run e2e tests until we migrate whole category.
See https://app.circleci.com/pipelines/github/aws-amplify/amplify-category-api/1412/workflows/79a728ce-2ecd-4822-ae15-35e5ee32cfb0 

```
src/__tests__/BelongsToTransformerV2.e2e.test.ts(49,69): error TS2345: Argument of type '() => import("/Users/sobkamil/git/amplify-category-api/packages/amplify-graphql-auth-transformer/node_modules/@aws-amplify/graphql-transformer-core/lib/transformation/transform").GraphQLTransform' is not assignable to parameter of type '() => import("/Users/sobkamil/git/amplify-category-api/packages/amplify-graphql-transformer-core/lib/transformation/transform").GraphQLTransform'.
  Type 'import("/Users/sobkamil/git/amplify-category-api/packages/amplify-graphql-auth-transformer/node_modules/@aws-amplify/graphql-transformer-core/lib/transformation/transform").GraphQLTransform' is not assignable to type 'import("/Users/sobkamil/git/amplify-category-api/packages/amplify-graphql-transformer-core/lib/transformation/transform").GraphQLTransform'.
    Types have separate declarations of a private property 'options'.
src/__tests__/MapsToTransformer.e2e.test.ts(57,64): error TS2345: Argument of type '() => import("/Users/sobkamil/git/amplify-category-api/packages/amplify-graphql-auth-transformer/node_modules/@aws-amplify/graphql-transformer-core/lib/transformation/transform").GraphQLTransform' is not assignable to parameter of type '() => import("/Users/sobkamil/git/amplify-category-api/packages/amplify-graphql-transformer-core/lib/transformation/transform").GraphQLTransform'.
[...]

PR #893, opened 2022-10-17
fix: resolution of nested stack template url

sobolk:
#### Description of changes

Do not cache nested stack template urls. 
The nested stack processing goes through two stage resolution.
1. it resolves root stack (template urls are "<unresolved"> by this time).
2. for each nested stack it invokes `_prepareTemplateAsset` which populates templateUrl.
3. runs root stack resolution again. it's expected that template urls resolve here (if not cached).

See for reference:
https://github.com/aws/aws-cdk/blob/5a0595edb1c5345552acbacefe2d09ec0bea31c1/packages/%40aws-cdk/core/lib/private/prepare-app.ts#L30-L56

Sample error caused by this.
image

#### Checklist

- [ x] PR description included
- [ x] `yarn test` passes

### DIFF
--- a/packages/amplify-graphql-transformer-core/src/cdk-compat/nested-stack.ts
+++ b/packages/amplify-graphql-transformer-core/src/cdk-compat/nested-stack.ts
@@ -47,7 +47,7 @@ export class TransformerNestedStack extends TransformerRootStack {
     this.parameters = props.parameters || {};
 
     this.resource = new CfnStack(parentScope, `${id}.NestedStackResource`, {
-      templateUrl: Lazy.string({
+      templateUrl: Lazy.uncachedString({
         produce: () => this._templateUrl || '<unresolved>',
       }),
       parameters: Lazy.any({


marcvberg:
Can you give some clarity on the issue this is meant to resolve? This is causing many E2E failures which appear to be because the URL is not yet resolved

sobolk:
> Can you give some clarity on the issue this is meant to resolve? This is causing many E2E failures which appear to be because the URL is not yet resolved

@marcvberg this is part of CDKv2 migration. See target branch of this PR. The context behind this is that:
- we've been migrating the code to make it compile with new CDK APIs. But we were unable to run e2e tests at that point, see https://github.com/aws-amplify/amplify-category-api/pull/860#discussion_r989303479 .
- now that we can run e2e tests we're fixing places that contribute to the problems.
- This particular mistake made plenty of tests fail eagerly at cfn deployment and was masking other problems. I.e. couple of tests got green and others advanced to stages where other problems surfaced. (which will be fixed in other PR, I don't want to mix them up).

alharris-at (approved):

PR #1076, opened 2022-12-08
feat(api): convert rds schema graphql

sundersc:
#### Description of changes

Generate GraphQL schema from schema internal representation data structure.

Temporarily we are reading the DB details from `rds.env` file from the API resource folder. Eventually, we will modify the workflow to prompt from the customer.

Sample rds.env file below:

```json
{
  "host": "host.server.com",
  "database": "db_name",
  "port": 3306,
  "username": "username",
  "password": "password"
}
```

Running `amplify import api` will import the database schema and convert it to GraphQL schema and store it in `schema.rds.graphql` file.

- manual test
- added unit tests

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-category-api/src/commands/api/import.ts
+++ b/packages/amplify-category-api/src/commands/api/import.ts
@@ -2,7 +2,7 @@ import { $TSAny, $TSContext, AmplifyCategories, pathManager, stateManager } from
 import { printer } from 'amplify-prompts';
 import * as path from 'path';
 import fs from 'fs-extra';
-import { MySQLDataSourceAdapter, generteGraphQLSchema, Schema, Engine } from '@aws-amplify/graphql-schema-generator';
+import { MySQLDataSourceAdapter, generateGraphQLSchema, Schema, Engine } from '@aws-amplify/graphql-schema-generator';
 const subcommand = 'import';
 
 export const name = subcommand;
@@ -25,7 +25,7 @@ export const run = async (context: $TSContext) => {
   const schema = new Schema(new Engine('MySQL'));
   models.forEach(m => schema.addModel(m));
 
-  const schemaString = generteGraphQLSchema(schema);
+  const schemaString = generateGraphQLSchema(schema);
   writeSchemaFile(apiResourceDir, schemaString);
 
   printer.info('Successfully imported the database schema.');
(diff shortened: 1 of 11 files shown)


packages/amplify-category-api/src/commands/api/import.ts
```diff
@@ -0,0 +1,66 @@
...
+  printer.info('Successfully imported the database schema.');
+};
+
+const getResourceDir = () => {
+  const apiNames = Object.entries(stateManager.getMeta()?.api || {})
+    .filter(([_, apiResource]) => (apiResource as $TSAny).service === 'AppSync')
+    .map(([name]) => name);
+  if (apiNames.length === 0) {
+    printer.info(
+      'No GraphQL API configured in the project.',
```
phani-srikar, line 40: Because this is part of `import` flow, so there might not be any existing APIs. So, we should assert length as 0?

packages/amplify-category-api/src/commands/api/import.ts
```diff
@@ -0,0 +1,66 @@
...
+  const apiName = apiNames[0];
+  const apiResourceDir = path.join(pathManager.getBackendDirPath(), AmplifyCategories.API, apiName);
+  return apiResourceDir;
+};
+
+// TODO: Move the file handling to amplify-cli-core
+
+const readConfigFile = async (apiResourceDir: string) => {
+  const configFilePath = path.join(apiResourceDir, 'rds.env');
+  return fs.readJSON(configFilePath);
```
phani-srikar, line 60: when is this file created? Asking because this might  throw if it doesn't exist.

packages/amplify-graphql-schema-generator/src/schema-generator/generate-schema.ts
```diff
@@ -0,0 +1,207 @@
...
+      },
+      arguments: [
+        {
+          kind: "Argument",
+          name: {
+            kind: "Name",
+            value: "value",
+          },
+          value: {
+            kind: "StringValue",
```
phani-srikar, line 56: Just a comment for future, is it possible to avoid constructing these GraphQL objects via big JSON inputs and do it via some API?

phani-srikar (changes requested):

packages/amplify-category-api/src/commands/api/import.ts
sundersc: We need to have a GraphQL API for the import datasource to work. Since we haven't started those tasks yet, this is a temporary solution to test the import workflow with rds.env file.

packages/amplify-category-api/src/commands/api/import.ts
sundersc: This will not exist in the final product. This is a temporary solution since we haven't completed the CLI workflow changes yet.

phani-srikar (approved):

PR #1083, opened 2022-12-10
fix: resolvers generated from index respect StackMappings

alharris-at:
#### Description of changes
Use stack resource manage to get stack to attach index-based resolvers to. Without this change, the StackMapping isn't being respected for these items. Add an e2e test to verify the behavior.

#### Issue #, if available
Internal ticket.

Tested manually, and via e2e test which verifies that resolvers are in the correct stack, and that they can still be queried against.

[E2E Test Flow](https://app.circleci.com/pipelines/github/aws-amplify/amplify-category-api/2830/workflows/edad96fd-a3e1-439a-a445-e80c4331817e)

#### Checklist
- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-graphql-index-transformer/src/resolvers.ts
+++ b/packages/amplify-graphql-index-transformer/src/resolvers.ts
@@ -458,10 +458,11 @@ function makeQueryResolver(config: IndexDirectiveConfiguration, ctx: Transformer
     throw new Error(`Could not find datasource with name ${dataSourceName} in context.`);
   }
 
+  const resolverResourceId = ResolverResourceIDs.ResolverResourceID(queryTypeName, queryField);
   const resolver = ctx.resolvers.generateQueryResolver(
     queryTypeName,
     queryField,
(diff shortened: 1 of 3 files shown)


marcvberg (approved):
LGTM

sundersc (approved):

packages/amplify-graphql-index-transformer/src/resolvers.ts
```diff
@@ -537,7 +538,8 @@ function makeQueryResolver(config: IndexDirectiveConfiguration, ctx: Transformer
       `${queryTypeName}.${queryField}.{slotName}.{slotIndex}.res.vtl`,
     ),
   );
-  resolver.mapToStack(table.stack);
+
+  resolver.mapToStack(ctx.stackManager.getStackFor(resolverResourceId, table.stack.node._actualNode.id));
```
sobolk, line 542: `node._actualNode` is not available in v2 of cdk :-(

PR #1120, opened 2022-12-21
test(api): e2e core library for rds tests

sundersc:
#### Description of changes

E2E Core library for RDS tests. This includes the below helpers.

- Create a RDS Instance
- Add external IP of current machine running the test to security group
- Remove the IP from the security group
- Delete a RDS Instance
- A sample test to show how to use the above methods

- E2E Test

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-e2e-core/src/utils/rds.ts
+++ b/packages/amplify-e2e-core/src/utils/rds.ts
@@ -1,0 +1,173 @@
+import { 
+  RDSClient,
+  CreateDBInstanceCommand,
+  waitUntilDBInstanceAvailable,
+  DeleteDBInstanceCommand,
+  waitUntilDBInstanceDeleted 
+} from "@aws-sdk/client-rds";
+import { EC2Client, AuthorizeSecurityGroupIngressCommand } from '@aws-sdk/client-ec2';
+import { knex } from 'knex';
+
+const DEFAULT_DB_INSTANCE_TYPE = "db.m5.large";
+const DEFAULT_DB_STORAGE = 8;
+const DEFAULT_SECURITY_GROUP = "default";
+
+export const createRDSInstance = async (config: {
+  identifier: string,
+  engine: 'mysql',
+  dbname: string,
+  username: string,
+  password: string,
+  region: string,
+  instanceClass?: string,
+  storage?: number,
+}) => {
+  const client = new RDSClient({ region: config.region });
+  const params = {
+    /** input parameters */
+    "DBInstanceClass": config.instanceClass ?? DEFAULT_DB_INSTANCE_TYPE,
+    "DBInstanceIdentifier": config.identifier,
+    "AllocatedStorage": config.storage ?? DEFAULT_DB_STORAGE,
+    "Engine": config.engine,
+    "DBName": config.dbname,
+    "MasterUsername": config.username,
+    "MasterUserPassword": config.password,
+  };
+  const command = new CreateDBInstanceCommand(params);
+  
+  try {
+    await client.send(command);
+    const availableResponse = await waitUntilDBInstanceAvailable(
+      {
+        maxWaitTime: 3600,
+        maxDelay: 120,
+        minDelay: 60,
+        client,
+      },
+      {
+        DBInstanceIdentifier: config.identifier,
+      },
+    );
+
+    const dbInstance = (availableResponse as any).DBInstances[0];
+    if (!dbInstance) {
+      throw new Error("RDS Instance details are missing.");
+    }
+
+    return {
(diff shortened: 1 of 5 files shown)


packages/amplify-e2e-core/src/utils/rds.ts
```diff
@@ -0,0 +1,210 @@
...
+      throw new Error("RDS Instance details are missing.");
+    }
+
+    return {
+      endpoint: dbInstance.Endpoint.Address,
+      port: dbInstance.Endpoint.Port,
+      dbName: dbInstance.DBName,
+    };
+  } catch (error) {
+    console.log(error);
```
alharris-at, line 67: Probably want `console.error` here.

packages/amplify-e2e-core/src/utils/rds.ts
```diff
@@ -0,0 +1,210 @@
...
+  waitUntilDBInstanceDeleted 
+} from "@aws-sdk/client-rds";
+import { EC2Client, AuthorizeSecurityGroupIngressCommand, RevokeSecurityGroupIngressCommand } from '@aws-sdk/client-ec2';
+import { knex } from 'knex';
+
+const DEFAULT_DB_INSTANCE_TYPE = "db.m5.large";
+const DEFAULT_DB_STORAGE = 8;
+const DEFAULT_SECURITY_GROUP = "default";
+
+export const createRDSInstance = async (config: {
```
alharris-at, line 15: Could you add docs to these new utils. No time like the present to improve the state of our docs 👯

alharris-at: Could you also add response types to the declaration?

packages/amplify-e2e-core/src/utils/rds.ts
```diff
@@ -0,0 +1,210 @@
...
+  const client = new RDSClient({ region });
+  const params = {
+    "DBInstanceIdentifier": identifier,
+    "SkipFinalSnapshot": true,
+  };
+  const command = new DeleteDBInstanceCommand(params);
+  try {
+    await client.send(command);
+    
+    // TODO: Revisit the below logic for waitUntilDBInstanceDeleted.
```
alharris-at, line 82: When are we revisiting this? What's blocking us getting this in place today?

alharris-at (commented):
Nice! Just a few comments about making sure our utilities are nicely documented, etc.

packages/amplify-e2e-core/src/utils/rds.ts
sundersc: Added docs for the new methods.

packages/amplify-e2e-core/src/utils/rds.ts
sundersc: I see some issues with the way SDK method 'waitUntilDBInstanceDeleted' works. It polls for the instance status at regular intervals and returns once the status is 'DELETED'. But the problem is once we trigger the instance deletion, it takes few minutes for the instance to be actually deleted. SDK returns 'DELETED' status only for few seconds after instance deletion. After that, it returns 'NO SUCH INSTANCE' error. I can reduce the interval to a lower value and get it to work most of the times but still there are changes of flaky test.

With current implementation, we trigger the deletion and exit the test. If test doesn't fail at this step, the instance will be deleted few minutes (usually 2-3 minutes) after the test run.

alharris-at (approved):

packages/amplify-e2e-tests/src/__tests__/rds-v2.test.ts
```diff
@@ -0,0 +1,63 @@
...
+    // Get the public IP of the machine running the test
+    const url = "http://api.ipify.org/";
+    const response = await axios(url);
+    publicIpCidr = `${response.data.trim()}/32`;
+  });
+
+  it("create database, setup initial tables and delete database", async () => {
+    // This test performs the below
+    // 1. Create a RDS Instance
+    // 2. Add the external IP address of the current machine to security group inbound rule to allow public access
```
phani-srikar, line 19: If for some reason these instances are not cleaned up, allowing public access might raise security alarms?

phani-srikar (approved):

packages/amplify-e2e-tests/src/__tests__/rds-v2.test.ts
sundersc: We delete the security group inbound rule at the end of the test. Even for some reason if it fails to delete the rule, we allow access to one specific IP so the risk is very low. Also, we run the test in a separate account and there is no sensitive data on this account.

PR #1126, opened 2022-12-21
throw a cpk warning when the feature flag is disabled and datastore is used in an Amplify Project

manaswi223:
#### Checklist

- [ ] PR description included
- [x] `yarn test` passes


dpilch:
Looks good, can you add a test to cover the warning message case?

packages/amplify-graphql-transformer-core/src/transformation/transform.ts
```diff
@@ -90,7 +92,7 @@ export interface GraphQLTransformOptions {
   readonly overrideConfig?: OverrideConfig;
 }
 export type StackMapping = { [resourceId: string]: string };
-export class GraphQLTransform {
+export class GraphQLTransform {
```
david-mcafee, line 95: I'm not a fan of committing linting changes alongside actual code changes. I would remove the linting updates from this PR.

packages/amplify-graphql-transformer-core/src/transformation/transform.ts
```diff
@@ -822,4 +825,12 @@ export class GraphQLTransform {
       index++;
     }
   }
+
+// Checks if DataStore is enabled and CPK Feature Flag is true. Throws a warning otherwise.
+public validateCPKFeatureFlag = (context: TransformerContext) => {
+  const isDataStoreEnabled = context.isProjectUsingDataStore();
+  if (isDataStoreEnabled && !isCPKFeatureEnabled(context)) {
+    console?.warn(`WARNING: Your schema has a custom primary key but the Feature Flag "${cpkFeatureFlagName}" is disabled. Check the value in your "amplify/cli.json" file, change it to "true" and re-run.`);
```
david-mcafee, line 833: DataStore being enabled without the CPK feature flag doesn't mean the schema is using a custom primary key. We should only throw the error if the schema actually contains a custom primary key.

david-mcafee (changes requested):

packages/amplify-graphql-transformer-core/src/transformation/transform.ts
manaswi223: sure, can do that!

david-mcafee: Thanks @manaswi223!

packages/amplify-graphql-transformer-core/src/transformation/transform.ts
manaswi223: oh wow, good catch David! Will change accordingly 🤦🏼‍♀️

manaswi223: updated with validating hasprimarykey() or not :)

packages/amplify-graphql-index-transformer/src/graphql-primary-key-transformer.ts
```diff
@@ -116,10 +119,8 @@ function validate(config: PrimaryKeyDirectiveConfiguration, ctx: TransformerCont
   const { object, field, sortKeyFields } = config;
 
   validateNotSelfReferencing(config);
-
-  const modelDirective = object.directives!.find(directive => {
-    return directive.name.value === 'model';
-  });
+  validateCPKFeatureFlag(ctx,config);
```
phani-srikar, line 122: If there are 2 models with primary keys, can you confirm that we don't throw the warning twice?

packages/amplify-graphql-transformer-core/src/transformation/transform.ts
```diff
@@ -1,825 +0,0 @@
-/* eslint-disable no-new */
```
phani-srikar, line 1: I don't think you meant to delete these files. Please re-add them.

phani-srikar (changes requested):

packages/amplify-graphql-index-transformer/src/graphql-primary-key-transformer.ts
manaswi223: Thank you Phani - it was warning 2 times when my schema has 2 @primarykey.
So, changed my code and now the warning isn't producing twice

david-mcafee:
@manaswi223 - is [https://github.com/aws-amplify/amplify-category-api/pull/1133](https://github.com/aws-amplify/amplify-category-api/pull/1133) the latest PR for this change? If so, can we close this PR to eliminate confusion? Thanks!

PR #1133, opened 2022-12-28
throw a cpk warning on the CLI when a schema has CPK, the project has datastore enabled and the corresponding feature flag is disabled

manaswi223:
#### Description of changes

This PR helps the CX by throwing a CPK warning ⚠️ on the CLI when:
1. schema has a CPK - @primarykey and
2. conflict detection is enabled (datastore is used in the project) and
3. the feature flag for CPK - `respectPrimaryKeyAttributesOnConnectionField` is false in "amplify/cli.json" file

When CPK modelgen is being disabled in a schema with CPK's, this PR would aid the CX to enable the feature flag and solve issues with the CPK models.

sample schema used:

```
type Todo @model {
  todoId: ID! @primaryKey
  content: String
}

type TodoCopy @model {
  todoCopyId: ID! @primaryKey
  contentCopy: String
}
```

screenshot of the warning when the schema is compiled by running the command -
`amplify-dev api gql-compile`

Screen Shot 2022-12-28 at 12 23 57 PM

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-graphql-transformer-core/src/transformation/transform.ts
+++ b/packages/amplify-graphql-transformer-core/src/transformation/transform.ts
@@ -832,7 +832,7 @@ public validateCPKFeatureFlag = (context: TransformerContext ) => {
   if (isDataStoreEnabled && !isCPKFeatureEnabled(context) && context.isProjectUsingCPK()) {
     printer.info(
       `
-⚠️  WARNING: Your schema has a custom primary key but the Feature Flag "${cpkFeatureFlagName}" is disabled. Check the value in your "amplify/cli.json" file, change it to "true" and re-run
+⚠️  WARNING: Your schema has a custom primary key but the Feature Flag "${cpkFeatureFlagName}" is disabled. Check the value in your "amplify/cli.json" file, change it to `true`, and execute `amplify codegen models` with the CLI
   `,
       'yellow',
     );
--- a/packages/amplify-graphql-transformer-core/src/transformation/transform.ts
+++ b/packages/amplify-graphql-transformer-core/src/transformation/transform.ts
@@ -31,6 +31,7 @@ import _ from 'lodash';
 import os from 'os';
 import * as path from 'path';
 import * as vm from 'vm2';
+import { DocumentNode } from 'graphql/language';
 import { ResolverConfig, TransformConfig } from '../config/transformer-config';
 import { InvalidTransformerError, SchemaValidationError, UnknownDirectiveError } from '../errors';
 import { GraphQLApi } from '../graphql-api';
@@ -52,9 +53,10 @@ import {
   matchFieldDirective,
   matchInputFieldDirective,
   sortTransformerPlugins,
+  isCPKFeatureEnabled,
+  cpkFeatureFlagName,
 } from './utils';
 import { validateAuthModes, validateModelSchema } from './validation';
-import { DocumentNode } from 'graphql/language';
 import { TransformerPreProcessContext } from '../transformer-context/pre-process-context';
 import { AmplifyApiGraphQlResourceStackTemplate } from '../types/amplify-api-resource-stack-types';
 
@@ -211,6 +213,7 @@ export class GraphQLTransform {
       throw new SchemaValidationError(errors);
     }
 
+    this.validateCPKFeatureFlag(context);
     for (const transformer of this.transformers) {
       if (isFunction(transformer.before)) {
         transformer.before(context);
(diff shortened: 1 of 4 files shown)


packages/amplify-graphql-transformer-core/src/transformation/transform.ts
```diff
@@ -822,4 +825,17 @@ export class GraphQLTransform {
...
     }
   }
+
+// Checks if DataStore is enabled and CPK Feature Flag is true. Throws a warning otherwise.
+public validateCPKFeatureFlag = (context: TransformerContext ) => {
+  const isDataStoreEnabled = context.isProjectUsingDataStore();
+  if (isDataStoreEnabled && !isCPKFeatureEnabled(context) && context.isProjectUsingCPK()) {
+    printer.info(
+      `
+⚠️  WARNING: Your schema has a custom primary key but the Feature Flag "${cpkFeatureFlagName}" is disabled. Check the value in your "amplify/cli.json" file, change it to "true" and re-run
```
david-mcafee, line 835: I would clarify what we mean by `re-run` here. For example:

```suggestion
⚠️  WARNING: Your schema has a custom primary key but the Feature Flag "${cpkFeatureFlagName}" is disabled. Check the value in your "amplify/cli.json" file, change it to `true`, and execute `amplify codegen models` with the CLI
```

@renebrandel any thoughts on the wording, here? Thanks!

david-mcafee (changes requested):
Left a minor suggestion

packages/amplify-graphql-transformer-core/src/transformer-context/index.ts
```diff
@@ -98,4 +101,12 @@ export class TransformerContext implements TransformerContextProvider {
   public isProjectUsingDataStore(): boolean {
     return !!this.resolverConfig?.project || !!this.resolverConfig?.models;
   }
+
+  public isProjectUsingCPK = (): boolean => {
```
david-mcafee, line 105: Can we add unit tests for this check? Thanks!!

david-mcafee (changes requested):

packages/amplify-graphql-transformer-core/src/transformation/transform.ts
manaswi223: Another CX:
For some customers using the old Amplify projects -->
manually switching the FF to true can also be a breaking change and might lose some data. Should we also warn them regarding this? 

thank you @AaronZyLee for catching this perception

david-mcafee: @AaronZyLee - can you elaborate more on this?

david-mcafee:
@manaswi223 looks like this PR can be closed now, since [the other PR](https://github.com/aws-amplify/amplify-category-api/pull/1126) was merged?

manaswi223:
no @david-mcafee David, that PR is closed and this is the actual PR & awaiting review from @renebrandel.

packages/amplify-graphql-transformer-core/src/transformation/transform.ts
david-mcafee: @manaswi223 - thoughts on the suggestion I made here (i.e. asking the customer to re-run `amplify codegen models`?

david-mcafee:
> no @david-mcafee David, that PR is closed and this is the actual PR & awaiting review from @renebrandel.

Thanks @manaswi223! I left a suggestion on updating the messaging above :)

manaswi223:
thanks David. Committed that suggestion :)

packages/amplify-graphql-transformer-core/src/transformer-context/index.ts
```diff
@@ -98,4 +101,12 @@ export class TransformerContext implements TransformerContextProvider {
...
+    const obj = this.inputDocument.definitions.find(def => def.kind === 'ObjectTypeDefinition') as ObjectTypeDefinitionNode;
+    const primaryKeyField = obj.fields?.find(field => field.directives?.find(directive => directive.name.value === 'primaryKey'));
+    if (!primaryKeyField || primaryKeyField.name.value === DEFAULT_ID_FIELD_NAME) {
```
AaronZyLee, line 108: I think the following schema should also be considered as custom primary key (id as primary key and with sort key fields defined)
```GraphQL
type Todo @model {
  id: ID! @primaryKey(sortKeyFields:["name"])
  name: String!
}
```

AaronZyLee (changes requested):

packages/amplify-graphql-transformer-core/src/transformer-context/index.ts
david-mcafee: Looks like we're only mocking the return value for this in the unit tests - thoughts on adding testing for this function?

packages/amplify-graphql-transformer-core/src/transformer-context/index.ts
david-mcafee: +1

david-mcafee (changes requested):

PR #1152, opened 2023-01-10
fix: disable searchable nodeToNode encryption unless it is already deployed to mitigate impact from enabling or disabling.

alharris-at:
#### Description of changes
Mitigate impact from enabling nodeToNodeEncryption, by allow enabling it by default if it is already set in the `#current-cloud-backend` directory at the time of gql compilation.

Also add support for a `NodeToNodeEncryption` param that can be set in `transform.conf.json` to facilitate overrides and manual testing. This isn't modeled as a PARAM like the rest of the params to ensure we're not looking at TOKENs in CFN make making the decision on whether to enable the flag.

So if `NodeToNodeEncryption` is set in the `transform.conf.json` file, that will be respected primarily.

If that flag is not set, and `NodeToNodeEncryption` is found in the `#current-cloud-backend` then enable on the next deployment.

If no occurrence is found, then return false. Fall back to false in case that some part of this process fails.

#### Issue #, if available
N/A

Added unit tests for core logic, and e2e tests to confirm behavior in the following cases:
1. New App with `NodeToNodeEncryption` set to `true` in the `transform.conf.json`
2. New App with `NodeToNodeEncryption` set to `false` in the `transform.conf.json`
3. Subsequent with no `transform.conf.json` flag set, and previous deployment state has `NodeToNodeEncryption` true stays true, there is no data loss.
4. Subsequent with no `transform.conf.json` flag set, and previous deployment state has `NodeToNodeEncryption` false stays false, there is no data loss.
5. Toggling NodeToNodeEncryption from `true` to `false` will rebuild Searchable instance, there will be data loss.

#### Checklist
- [x] PR description included
- [x] `yarn test` passes


packages/amplify-category-api/src/graphql-transformer/transformer-factory.ts
```diff
@@ -102,7 +102,9 @@ const getTransformerFactoryV2 = (
   ];
 
   if (options?.addSearchableTransformer) {
-    transformerList.push(new SearchableModelTransformerV2());
+    const resourceDirParts = resourceDir.split(path.sep);
+    const apiName = resourceDirParts[resourceDirParts.length - 1];
+    transformerList.push(new SearchableModelTransformerV2(apiName));
```
phani-srikar, line 107: I'll leave it upto you to take a call, but instead of passing the `apiName` all the way down, we could fetch it wherever needed using a utility method that does something like:
```
const meta = stateManager.getMeta();
const resourceName = Object.entries(meta[AmplifyCategories.API]).find(
  ([, value]) => (value as { service: string }).service === AmplifySupportedService.APPSYNC,
)?.[0];
```

packages/amplify-e2e-tests/src/__tests__/graphql-v2/searchable-node-to-node-encryption/searchable-previous-deployment-had-node-to-node.test.ts
```diff
@@ -0,0 +1,179 @@
...
+
+    expect(searchDomainPropsFirstDeploy).toHaveProperty('NodeToNodeEncryptionOptions');
+    expect(searchDomainPropsFirstDeploy.NodeToNodeEncryptionOptions.Enabled).toEqual(true);
+
+    // Subsequent deploy without flag set
+    removeTransformConfigValue(projRoot, projectName, 'NodeToNodeEncryption');
+    await amplifyPush(projRoot);
+
+    appSyncClient = getAppSyncClientFromProj(projRoot);
+    await runAndValidateQuery('test1', 'test1', 10, 2); // Expect two records
```
phani-srikar, line 64: In this case, shouldn't the search domain be re-created? Since, now we're turning the encryption off.  How does getting back 2 records test that, sorry if I missed something here.

phani-srikar (changes requested):

packages/amplify-e2e-tests/src/__tests__/graphql-v2/searchable-node-to-node-encryption/searchable-previous-deployment-had-node-to-node.test.ts
alharris-at: Behavior of the systems is such that we will maintain the current state of nodeToNode (either on or off), so there should be no loss of data. The explicit flag is being removed to test the state that "previous deployment had it, I don't care how, now I'm doing a normal deploy"

packages/amplify-category-api/src/graphql-transformer/transformer-factory.ts
alharris-at: I'll take a look at doing this. I'm seen some customer apps where there are multiple directories under API, but I'm guessing that what you're doing here by filtering to APPSYNC should mitigate that issue.

packages/amplify-e2e-tests/src/__tests__/graphql-v2/searchable-node-to-node-encryption/searchable-subsequent-deployment-override-remove-node-to-node.test.ts
```diff
@@ -0,0 +1,178 @@
...
+  it('if previous deployment had NodeToNodeEncryption, and flag is set to false on subsequent push, deploy, search domain will be recreated', async () => {
+    const v2Schema = 'transformer_migration/searchable-v2.graphql';
+
+    await addApiWithoutSchema(projRoot, { apiName: projectName });
+    updateApiSchema(projRoot, projectName, v2Schema);
+
+    const searchableStackPath = path.join(projRoot, 'amplify', 'backend', 'api', projectName, 'build', 'stacks', 'SearchableStack.json');
+
+    // Initial Deploy with no flag set
+    setTransformConfigValue(projRoot, projectName, 'NodeToNodeEncryption', true);
```
AaronZyLee, line 47: Is this line a conflict with the comment above?

AaronZyLee (commented):
The approach looks good if this code is the minimum api version used by customer. The rollback will not flip the value as false. Still I have some concerns as following.
1. How do we prevent the user to use previous version of CLI (version < 10.6 in particular)? If the older version is used when they encounter other issues (not related to search) and decide to downgrade, they will finally have the false value forcefully as the code in this PR does not exist in the previous version. 
In this case, I think we should provide some warning message to the user. If we detect the search domain is provisioned and the nodeToNode is true, we should warn the user not to downgrade to version less than 10.6 which will force the false value and delete the original instance.
2. I see several searchable-related e2e tests are added. I think this is a two bladed sword as it ensures the safety while increasing the pipeline load since the provisioning and deletion of search domain consume plenty of time and will have trouble when the tests fail and when many concurrent cloud e2e runs. 
I suggest if the nodeToNode value is the only thing that matters here we might not need e2e tests (or we can simplify into fewer cases). We just need to check if that value is resolved to the correct value in different scenarios. If the data loss is a consideration in the e2e test, I think it is enough to test `false` => `true` not causing the data loss. It is not necessary to test `true` => `false` causing replacement of domain as it is already acknowledged in the CFN docs.

packages/amplify-category-api/src/graphql-transformer/transformer-factory.ts
alharris-at: Actually, I'd prefer to leave as-is, it's a little uglier, but since we've just smoke tested the behavior using the current impl, I'd rather leave it as close to the same as possible.

phani-srikar (approved):

AaronZyLee (approved):

packages/amplify-e2e-tests/src/__tests__/graphql-v2/searchable-node-to-node-encryption/searchable-subsequent-deployment-overrides-node-to-node.test.ts
```diff
@@ -0,0 +1,177 @@
...
+    });
+    await addAuthWithDefault(projRoot, {});
+  });
+
+  afterEach(async () => {
+    await deleteProject(projRoot);
+    deleteProjectDir(projRoot);
+  });
+
+  it('if previous deployment had no NodeToNodeEncryption, and flag is enabled on subsequent push, deploy, there should be no data loss', async () => {
```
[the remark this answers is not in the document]
alharris-at: I'll mark this as work in progress, but may not want to block this release on it, since we've manually run migration smoke testing.

PR #1175, opened 2023-01-18
API key extension support

marcvberg:
#### Description of changes
Adds support for extending API keys through a command, with headless support, as well as an output in CFN for the expiration time of the API key.

When the expiration time of an API key in the console is farther away than the configured API key lifespan (which is defaulting to 7 days as of this PR) the expiration will not be updated.

Manual testing, added tests

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes


packages/amplify-category-api/src/provider-utils/awscloudformation/service-walkthroughs/appSync-walkthrough.ts
```diff
@@ -858,6 +879,18 @@ export async function askApiKeyQuestions(authSettings: $TSObject = undefined) {
...
+  const apiKeyExtensionQuestion = {
+    message: 'Days to extend API key expiration:',
+    default: '0',
+  };
+  const expirationDaysNum = Number(await prompter.input(
+    apiKeyExtensionQuestion.message,
+    { initial: apiKeyExtensionQuestion.default as string },
+  ));
+  return expirationDaysNum;
+}
```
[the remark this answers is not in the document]
phani-srikar: Also, we need that `validate` param to ensure that user does not give us decimals and probably also check for some upper-bound to the number of days key is extended.

packages/amplify-category-api/src/provider-utils/awscloudformation/utils/manage-api-key.ts
```diff
@@ -0,0 +1,65 @@
...
+export interface ApiKeyStatus {
+  expiration?: number;
+  key?: string;
+  apiId?: string;
+  description?: string;
+  exists: boolean;
+}
+
+const getAppSyncClient = async (context: $TSContext): Promise<AppSyncClient> => {
+  // const credentials = await context.amplify.executeProviderUtils(context, 'awscloudformation', 'retrieveAwsConfig');
```
phani-srikar, line 17: nit: remove commented out code

packages/amplify-category-api/src/provider-utils/awscloudformation/utils/manage-api-key.ts
```diff
@@ -0,0 +1,65 @@
...
+export const updateApiKeyExpiration = async (context: $TSContext, status: ApiKeyStatus, dayOffset: number): Promise<void> => {
+  const client = await getAppSyncClient(context);
+  const command = new UpdateApiKeyCommand({
+    apiId: status.apiId,
+    description: status.description,
+    expires: status.expiration + (dayOffset * SECONDS_PER_DAY),
+    id: status.key,
+  });
+  const response = await client.send(command);
+  if (response.$metadata.httpStatusCode >= 400) {
```
phani-srikar, line 62: `$metadata` is a property? Interesting.

packages/amplify-graphql-transformer-core/src/transformation/transform.ts
```diff
@@ -416,6 +416,12 @@ export class GraphQLTransform {
         description: 'Your GraphQL API ID.',
         exportName: Fn.join(':', [Aws.STACK_NAME, 'GraphQLApiKey']),
       });
+
+      new CfnOutput(rootStack, 'GraphQLAPIKeyExpirationOutput', {
```
phani-srikar, line 420: nit: I missed the part where this new output is needed/used in the implementation.

phani-srikar (changes requested):

packages/amplify-category-api/src/graphql-transformer/transform-graphql-schema-v2.ts
```diff
@@ -191,6 +189,21 @@ const buildAPIProject = async (
...
 
   const builtProject = await _buildProject(opts);
+  const apiKeyStatus = await getApiKeyStatus(context);
+  const apiKeyEntry: AppSyncAuthConfigurationAPIKeyEntry = [
+    opts.authConfig.defaultAuthentication, ...opts.authConfig.additionalAuthenticationProviders,
+  ].find(
+    (mode: AppSyncAuthConfigurationEntry) => mode.authenticationType === AuthorizationType.API_KEY && mode.apiKeyConfig,
+  ) as AppSyncAuthConfigurationAPIKeyEntry;
+  const config = apiKeyEntry?.apiKeyConfig;
+  const apiKeyId = `${(config as any)?.name || 'Default'}ApiKey`;
```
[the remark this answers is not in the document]
marcvberg: Not necessarily, in most cases the API key config does not define the name, I didn't even see it on the type. It won't compile without the cast because the property doesn't exist, but other uses of API key config are referencing the name property because they aren't type safe, so for proper feature parity I still use it

packages/amplify-category-api/src/category-utils/context-util.ts
```diff
@@ -73,6 +73,22 @@ export class ContextUtil {
     this.resourceDir = resourceDir;
     return resourceDir;
   }
+
+  getGraphQLAPIResourceName = async (
+    context: $TSContext,
+  ): Promise<string> => {
```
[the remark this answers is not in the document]
marcvberg: Resolving since I'll cut this out to use the provider util method

packages/amplify-category-api/src/provider-utils/awscloudformation/service-walkthroughs/appSync-walkthrough.ts
```diff
@@ -858,6 +879,18 @@ export async function askApiKeyQuestions(authSettings: $TSObject = undefined) {
   };
 }
 
+export async function askApiKeyExtension() {
+  const apiKeyExtensionQuestion = {
+    message: 'Days to extend API key expiration:',
+    default: '0',
```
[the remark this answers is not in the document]
marcvberg: I'll bump it to 7 - my initial thought was that this is an extension for manual intervention, so maybe we didn't need to assume anything about it

packages/amplify-category-api/src/provider-utils/awscloudformation/service-walkthroughs/appSync-walkthrough.ts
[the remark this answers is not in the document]
marcvberg: The upper bound gets a little tricky because it's dependent on the current date as well as the expiration, I tested that an invalid number will return an appropriately displayed message from appsync and I'd rather leave it at that than try to calculate out their service limits. I will put the integer validation in though

phani-srikar (approved):

packages/amplify-category-api/src/provider-utils/awscloudformation/utils/manage-api-key.ts
```diff
@@ -0,0 +1,71 @@
...
+
+export interface ApiKeyStatus {
+  expiration?: number;
+  key?: string;
+  apiId?: string;
+  description?: string;
+  exists: boolean;
+}
+
+const getAppSyncClient = async (context: $TSContext): Promise<AppSyncClient> => {
```
AaronZyLee, line 16: Is it possible to get the appsync client from `amplify-provider-awscloudformation` package as I see most of the client calls are defined there?

packages/amplify-category-api/src/provider-utils/awscloudformation/utils/manage-api-key.ts
```diff
@@ -0,0 +1,71 @@
...
+  const client = await getAppSyncClient(context);
+  const command = new UpdateApiKeyCommand({
+    apiId: status.apiId,
+    description: status.description,
+    expires: status.expiration + (dayOffset * SECONDS_PER_DAY),
+    id: status.key,
+  });
+  const response = await client.send(command);
+  if (response.$metadata.httpStatusCode >= 400) {
+    throw new Error('Failed to update API Key extension');
```
AaronZyLee, line 69: If I understand correctly, the API key will fail be extend if the key already expires at the update time? If so, is there a way to convey a message like `The API key fails to extend as it already expired. Try to recreate a new API key instead`?

AaronZyLee (approved):
Overall LGTM. Have some comments below

packages/amplify-category-api/src/provider-utils/awscloudformation/utils/manage-api-key.ts
marcvberg: Expired API keys can still be extended so long as they haven't been deleted, you just need to have an extension time that goes far enough to make it valid again.

If they API key is fully deleted, that will be different and is something we can't easily fix right now because of multiple stages of CFN

packages/amplify-category-api/src/provider-utils/awscloudformation/utils/manage-api-key.ts
marcvberg: It doesn't currently expose anything for this, and we don't want to add too many exposed APIs in that package, this is just me de-duping code in the file

PR #1181, opened 2023-01-19
chore(api): add datasource info to transformer context instance

sundersc:
#### Description of changes

Introduces model to datasource mapping to transformer context instance. This will help future work to differentiate between DDB vs MySQL for amplify GraphQL transformers.

Once this is merged, we can use the below logic to get database type.

```ts
const dbInfo = context.modelToDatasourceMap.get(modelName);
// If dbInfo doesn't have an entry for the model, then treat it as DDB.
const dbType = dbInfo ? dbInfo.dbType : 'DDB';
// dbType will be DDB or MySQL
```

- Manual test
- No E2E added (should be added as part of transformer changes)

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-graphql-transformer-core/src/transformation/transform.ts
+++ b/packages/amplify-graphql-transformer-core/src/transformation/transform.ts
@@ -173,14 +173,14 @@ export class GraphQLTransform {
    * @param schema The model schema.
    * @param references Any cloudformation references.
    */
-  public transform(schema: string, modelToDatasourceMap: Map<string, DatasourceType>): DeploymentResources {
+  public transform(schema: string, modelToDatasourceMap?: Map<string, DatasourceType>): DeploymentResources {
     this.seenTransformations = {};
     const parsedDocument = parse(schema);
     this.app = new App();
     const context = new TransformerContext(
       this.app,
       parsedDocument,
-      modelToDatasourceMap,
+      modelToDatasourceMap ?? new Map<string, DatasourceType>(),
       this.stackMappingOverrides,
       this.authConfig,
       this.options.sandboxModeEnabled,
--- a/packages/amplify-graphql-transformer-core/src/transformation/transform.ts
+++ b/packages/amplify-graphql-transformer-core/src/transformation/transform.ts
@@ -57,6 +57,7 @@ import { validateAuthModes, validateModelSchema } from './validation';
 import { DocumentNode } from 'graphql/language';
 import { TransformerPreProcessContext } from '../transformer-context/pre-process-context';
 import { AmplifyApiGraphQlResourceStackTemplate } from '../types/amplify-api-resource-stack-types';
+import { DatasourceType } from '../config/project-config';
 
 // eslint-disable-next-line @typescript-eslint/ban-types
 function isFunction(obj: any): obj is Function {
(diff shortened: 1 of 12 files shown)


packages/amplify-graphql-transformer-core/src/transformation/transform.ts
```diff
@@ -172,13 +173,14 @@ export class GraphQLTransform {
...
    */
-  public transform(schema: string): DeploymentResources {
+  public transform(schema: string, modelToDatasourceMap?: Map<string, DatasourceType>): DeploymentResources {
     this.seenTransformations = {};
     const parsedDocument = parse(schema);
     this.app = new App();
     const context = new TransformerContext(
       this.app,
       parsedDocument,
+      modelToDatasourceMap ?? new Map<string, DatasourceType>(),
```
phani-srikar, line 183: can we pass the input as-is, instead of this default initialization? I see that the target constructor accepts `undefined` - `modelToDatasourceMap: Map<string, DatasourceType> | undefined`

phani-srikar (changes requested):
You might have to re-base your branch to fix the CI job once this [other PR](https://github.com/aws-amplify/amplify-category-api/pull/1186) is merged

packages/amplify-graphql-transformer-core/src/transformation/transform.ts
sundersc: The advantage is we can always use `context. modelToDatasourceMap` without doing a null check in transformer code. Ideally this should be a mandatory field on `TransformerContext` type. To fix some of the failing unit tests, I added `undefined`.

packages/graphql-transformer-core/src/util/transformConfig.ts
```diff
@@ -255,25 +259,40 @@ export async function loadProject(projectDirectory: string, opts?: ProjectOption
  * Preference is given to the `schema.graphql` if provided.
  * @param projectDirectory The project directory.
  */
-export async function readSchema(projectDirectory: string): Promise<string> {
+export async function readSchema(projectDirectory: string): Promise<{schema: string, modelToDatasourceMap: Map<string, DatasourceType>}> {
+  let modelToDatasourceMap = new Map<string, DatasourceType>();
```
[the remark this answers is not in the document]
sundersc: The helper method `constructDataSourceMap` returns a map, so it is easier to merge the maps (in one line) than looping through this result set and calling `modelToDatasourceMap.set`.

packages/amplify-graphql-transformer-core/src/transformation/transform.ts
phani-srikar: But the transformer context still defines it as `public readonly modelToDatasourceMap: Map<string, DatasourceType> | undefined;`, so if we are accessing it, don't we still need to assert that it's defined?

sundersc: Removed the `undefined` on the type. Updated the unit tests to always pass a value.

sundersc:
> You might have to re-base your branch to fix the CI job once this [other PR](https://github.com/aws-amplify/amplify-category-api/pull/1186) is merged

Rebased on the latest feature branch.

phani-srikar (approved):

packages/amplify-graphql-transformer-core/src/transformation/transform.ts
```diff
@@ -172,13 +173,14 @@ export class GraphQLTransform {
    * @param schema The model schema.
    * @param references Any cloudformation references.
```
AaronZyLee, line 174: Nit: This line in comment can be changed to the correct description of the newly added param

AaronZyLee (approved):

PR #1187, opened 2023-01-23
feat: add graphql-schema-validation package

johnpc:
#### Description of changes

This PR adds a new package, `@aws-amplify/graphql-schema-validation`, that exports a sharable, browser-executable function `validate` that takes a string and returns whether it is valid graphql that conforms to amplify-category-api rules, and throws a human-readable error message otherwise.

The implementation is incomplete - it includes just a small handful of validators for now:

1. string is valid graphql
2. @index uses scalar types
3. @belongsTo directive requires that a @hasOne or @hasMany relationship already exists from parent to the related model.
4.  @manyToMany directive must be used in exactly two locations.

more validators will be added in future PRs as needed.

None

#### Issue #, if available

N/A

Unit tests added 

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-schema-validator/README.md
+++ b/packages/amplify-schema-validator/README.md
@@ -1,0 +1,17 @@
+# Amplify GraphQL Schema Validator
+
+This package contains logic used by AWS Amplify development teams to validate
+that customer's `schema.graphql` for their Amplify app is valid.
+
+## Contributing
+
+To add a validator, simply create a new `*.ts` file under `src/validators`
+exporting a function named `validate` that takes a `DocumentNode` object and
+returns void if successful, and returns an array of actionable `ValidationError` objects if not.
+
+Validators should be pure functions that do not depend on or mutate state.
+
+Once your validator is ready, update `src/index.ts` and import your validator
+there.
+
+Ensure your validator is tested and conforms to our coverage standards.
--- a/packages/amplify-schema-validator/package.json
+++ b/packages/amplify-schema-validator/package.json
@@ -1,0 +1,50 @@
+{
+  "name": "@aws-amplify/graphql-schema-validation",
+  "main": "./dist/index.js",
+  "private": true,
+  "types": "./dist/index.d.ts",
+  "version": "0.0.1",
+  "files": [
+    "dist",
+    "!**/__tests__/**/*"
+  ],
+  "scripts": {
+    "build": "tsc && npm run size",
+    "clean": "rm -rf build node_modules dist coverage",
+    "prettier": "prettier --write '**/*.{js,ts,json,md}'",
+    "test": "jest --coverage",
+    "size": "size-limit",
+    "lint-staged": "lint-staged"
+  },
+  "devDependencies": {
+    "@size-limit/preset-small-lib": "^2.2.4",
+    "@types/jest": "^24.0.18",
+    "fs-extra": "^11.1.0",
+    "jest": "^29.0.0",
+    "jest-environment-jsdom": "^29.3.1",
+    "lint-staged": "^9.2.5",
+    "ts-jest": "^29.0.2",
+    "ts-node": "^8.5.0"
+  },
+  "dependencies": {
+    "graphql": "^16.6.0"
+  },
+  "lint-staged": {
+    "*.{js,ts,json,md,scss,html}": [
+      "prettier --write",
+      "git add"
+    ],
+    "*.{js,ts}": [
+      "eslint --fix"
+    ]
+  },
+  "size-limit": [
+    {
+      "path": "./dist/index.js",
+      "limit": "5 KB",
+      "ignore": [
+        "graphql"
+      ]
+    }
+  ]
+}
--- a/packages/amplify-schema-validator/src/exceptions/invalid-directive-error.ts
+++ b/packages/amplify-schema-validator/src/exceptions/invalid-directive-error.ts
@@ -1,0 +1,10 @@
+/**
+ * Exception class for Invalid Directive errors
+ */
+export class InvalidDirectiveError extends Error {
+  constructor(message: string) {
+    super(message);
+    Object.setPrototypeOf(this, InvalidDirectiveError.prototype);
+    this.name = 'InvalidDirectiveError';
+  }
+}
--- a/packages/amplify-schema-validator/src/helpers/get-object-with-name.ts
+++ b/packages/amplify-schema-validator/src/helpers/get-object-with-name.ts
@@ -1,0 +1,17 @@
+import { DocumentNode, Kind, ObjectTypeDefinitionNode } from 'graphql';
+
+/**
+ * Finds a graphql object definition by name
+ *
+ * @param schema graphql schema
+ * @param name name of object being searched for
+ * @returns ObjectTypeDefinitionNode
+ */
+export const getObjectWithName = (schema: DocumentNode, name: string): ObjectTypeDefinitionNode | undefined => {
+  const objectTypeDefinitions = schema.definitions.filter(
+    (defintion) => defintion.kind === Kind.OBJECT_TYPE_DEFINITION,
+  ) as ObjectTypeDefinitionNode[];
+  return objectTypeDefinitions.find(
+    (objectTypeDefinition) => objectTypeDefinition.name.value === name,
+  );
+};
--- a/packages/amplify-schema-validator/src/helpers/is-scalar-or-enum.ts
+++ b/packages/amplify-schema-validator/src/helpers/is-scalar-or-enum.ts
@@ -1,0 +1,60 @@
+import { EnumTypeDefinitionNode, Kind, TypeNode } from 'graphql';
+
+type ScalarMap = {
+  [k: string]: 'String' | 'Int' | 'Float' | 'Boolean' | 'ID';
+};
+export const STANDARD_SCALARS: ScalarMap = {
+  String: 'String',
+  Int: 'Int',
+  Float: 'Float',
+  Boolean: 'Boolean',
+  ID: 'ID',
+};
+
+const OTHER_SCALARS: ScalarMap = {
+  BigInt: 'Int',
+  Double: 'Float',
+};
+
+export const APPSYNC_DEFINED_SCALARS: ScalarMap = {
+  AWSDate: 'String',
+  AWSTime: 'String',
+  AWSDateTime: 'String',
+  AWSTimestamp: 'Int',
+  AWSEmail: 'String',
+  AWSJSON: 'String',
+  AWSURL: 'String',
+  AWSPhone: 'String',
+  AWSIPAddress: 'String',
+};
+
+export const DEFAULT_SCALARS: ScalarMap = {
+  ...STANDARD_SCALARS,
+  ...OTHER_SCALARS,
+  ...APPSYNC_DEFINED_SCALARS,
+};
+
+/**
+ * Returns whether a given type is a scalar or enum
+ *
+ * @param type the type to check whether it's a scalar
+ * @param enums enums to consider as scalars
+ * @returns boolean
+ */
+export const isScalarOrEnum = (
+  type: TypeNode,
+  enums: EnumTypeDefinitionNode[],
+): boolean => {
+  if (type.kind === Kind.NON_NULL_TYPE) {
+    return isScalarOrEnum(type.type, enums);
+  } if (type.kind === Kind.LIST_TYPE) {
+    /* istanbul ignore next */
+    return isScalarOrEnum(type.type, enums);
+  }
+  for (const e of enums) {
+    if (e.name.value === type.name.value) {
+      return true;
+    }
+  }
+  return Boolean(DEFAULT_SCALARS[type.name.value]);
+};
--- a/packages/amplify-schema-validator/src/index.ts
+++ b/packages/amplify-schema-validator/src/index.ts
@@ -1,0 +1,30 @@
+/* eslint-disable @typescript-eslint/no-var-requires */
+/* eslint-disable global-require */
+
+import { parse } from 'graphql';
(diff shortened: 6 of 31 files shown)


packages/amplify-schema-validator/.gitignore
```diff
@@ -0,0 +1,7 @@
+# dependencies
```
alharris-at, line 1: Probably don't need/want a new gitignore in this package, can leverage the root one.

packages/amplify-schema-validator/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+`npm install`
+
+### Running tests
+
+`npm run test`
+
+### Building the package
+
+`npm run build`
```
alharris-at, line 29: As long as this package lives under the `packages` directory I wouldn't recommend this method since this lives in a monorepo, and we don't want multiple lockfiles. yarn, yarn test, and yarn build from root should cover these.

packages/amplify-schema-validator/import-check.ts
```diff
@@ -0,0 +1,13 @@
+import fs from 'fs';
+import path from 'path';
+
+const srcDir = path.join(__dirname, 'src');
+const indexPath = path.join(srcDir, 'index.ts');
+const validatorList = fs.readdirSync(path.join(srcDir, 'validators'));
+const indexFileContents = fs.readFileSync(indexPath);
```
alharris-at, line 7: Rather than this runtime import system which obfuscates dependencies, I'd prefer to explicitly import the validations we're running and losing this check.

packages/amplify-schema-validator/package.json
```diff
@@ -0,0 +1,50 @@
+{
```
alharris-at, line 1: This package should be specified as `private: true`

packages/amplify-schema-validator/src/exceptions/invalid-directive-error.ts
```diff
@@ -0,0 +1,7 @@
+export class InvalidDirectiveError extends Error {
```
alharris-at, line 1: General comment here, but I'd like to see comments on methods and classes in the `src` directory. (excluding tests)

packages/amplify-schema-validator/src/helpers/is-scalar-or-enum.ts
```diff
@@ -0,0 +1,53 @@
...
+};
+export const STANDARD_SCALARS: ScalarMap = {
+  String: 'String',
+  Int: 'Int',
+  Float: 'Float',
+  Boolean: 'Boolean',
+  ID: 'ID',
+};
+
+const OTHER_SCALARS: ScalarMap = {
```
alharris-at, line 14: Does Amplify support these?

packages/amplify-schema-validator/src/index.ts
```diff
@@ -0,0 +1,18 @@
+/* eslint-disable @typescript-eslint/no-var-requires */
+/* eslint-disable global-require */
+
+import { parse } from 'graphql';
+
+const allValidators = [
+  require('./validators/index-scalar-types').validate,
+  require('./validators/relation-required-with-belongs-to').validate,
+  require('./validators/two-many-to-many-locations').validate,
```
alharris-at, line 9: why use requires here instead of `import { blahValidator } from './validators/index-scalar-types';`

yarn.lock
```diff
@@ -40,6 +40,67 @@
     signedsource "^1.0.0"
     yargs "^15.3.1"
 
+"@aws-amplify/amplify-appsync-simulator@2.9.1":
```
alharris-at, line 43: These deps were pulled in from running `yarn setup-dev`, could you rerun w/ just a `yarn` so we're not adding back in the CLI packages to the lockfile?

packages/amplify-schema-validator/src/index.ts
```diff
@@ -0,0 +1,18 @@
...
+];
+
+export const validateSchema = (schemaString: string): boolean => {
+  const schema = parse(schemaString);
+  allValidators.forEach((validator) => {
+    validator(schema);
+  });
+  return true;
```
alharris-at, line 17: We should either return a boolean, or throw an error but if we're rely on the error to communicate info, this and underlying `validator` should just have a void return.

packages/amplify-schema-validator/src/index.ts
```diff
@@ -0,0 +1,18 @@
...
+const allValidators = [
+  require('./validators/index-scalar-types').validate,
+  require('./validators/relation-required-with-belongs-to').validate,
+  require('./validators/two-many-to-many-locations').validate,
+];
+
+export const validateSchema = (schemaString: string): boolean => {
+  const schema = parse(schemaString);
+  allValidators.forEach((validator) => {
+    validator(schema);
```
alharris-at, line 15: n.p. but probably name this validate instead of validator. or grab the validator and execute validate.

alharris-at: Also, instead of throwing on the first error, does it make sense to structure this to return a array of errors, or validation exceptions?

packages/amplify-schema-validator/README.md
johnpc: I'll remove this section since it's covered by the root level readme / contributing guide

packages/amplify-schema-validator/import-check.ts
johnpc: It's intended as a sort of awkward lint rule - we don't want to contributors to forget to actually import and run their validator. But I'm good with trusting code reviewers to catch that.

packages/amplify-schema-validator/src/helpers/is-scalar-or-enum.ts
johnpc: The code is lifted from https://github.com/aws-amplify/amplify-category-api/blob/main/packages/graphql-transformer-common/src/definition.ts#L35-L38, so I think it's safer to keep this in order to maintain parity.

packages/amplify-schema-validator/src/index.ts
johnpc: I'm not sure why, but in terms of test readability,

```
const result = validate(schema);
expect(result).toBe(true)
```

just looks so much cleaner to me than 

```
expect(() => validate(schema)).not.toThrow();
```

packages/amplify-schema-validator/src/index.ts
johnpc: Initially I was thinking of writing this file at build time to guarantee that all validators are run and limit the number of files to be touched by contributors to just their validator and their test. But since that is impractical, I reluctantly determined it's ok to also have contributors touch this file in a one-liner-y kind of way. The import strategy requires touching the file in two places.

But it's not a big deal, I don't think there will be too many merge conflicts in this file, and if there are, they would be very easily resolved.

packages/amplify-schema-validator/src/index.ts
johnpc: I like that idea! 

Think it makes sense if each validator returns an array of error objects, and the root `validate` method concatenates all the validator responses and passes if the resulting array is empty. Otherwise it throws a new error with all the validation failure reasons in one error.

packages/amplify-schema-validator/src/index.ts
alharris-at: From a testing perspective, I think you're probably right, but I think the decision to collect an array of error messages makes this point somewhat moot :D

packages/amplify-schema-validator/import-check.ts
alharris-at: Yeah, in my mind tests should probably cover 'did my new test catch anything'.

johnpc: Just a warning, if they write their tests to only run their own validator (and not the global export) the tests will not notice this.

packages/amplify-schema-validator/src/index.ts
```diff
@@ -0,0 +1,34 @@
...
+ * @param schemaString the graphql schema
+ * @returns void
+ */
+export const validateSchema = (schemaString: string): void => {
+  const schema = parse(schemaString);
+  let aggregatedErrors: Error[] = [];
+  allValidators.forEach((validate) => {
+    const errors = validate(schema);
+    aggregatedErrors = aggregatedErrors.concat(errors);
+  });
```
alharris-at, line 29: Rather than relying on mutable state here, I'd prefer just `flagMapping` over validators.

e.g. ```const validationErrors = allValidators.flatMap(validate => validate(schema));```

packages/amplify-schema-validator/src/exceptions/invalid-directive-error.ts
sundersc: Can we reuse the exception types from '@aws-amplify/graphql-transformer-core' package?

packages/amplify-schema-validator/src/helpers/get-object-with-name.ts
```diff
@@ -0,0 +1,17 @@
...
+/**
+ * Finds a graphql object definition by name
+ *
+ * @param schema graphql schema
+ * @param name name of object being searched for
+ * @returns ObjectTypeDefinitionNode
+ */
+export const getObjectWithName = (schema: DocumentNode, name: string): ObjectTypeDefinitionNode | undefined => {
+  const objectTypeDefinitions = schema.definitions.filter(
+    (defintion) => defintion.kind === Kind.OBJECT_TYPE_DEFINITION,
```
sundersc, line 12: Should we check if the object type definition contains a model directive? For example:
```graphql
type Todo {
  name: String
}
```
vs
```graphql
type Todo @model {
  name: String
}
```

packages/amplify-schema-validator/src/validators/index-scalar-types.ts
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * Validates that every @index directive is of a scalar type
+ *
+ * @param schema graphql schema
+ * @returns true
+ */
+export const validateIndexScalarTypes = (schema: DocumentNode): Error[] => {
+  const errors: Error[] = [];
+  const objectTypeDefinitions = schema.definitions.filter(
```
sundersc, line 20: Index can be defined only on 'model' types. Can we add that validation here?
Also consider filtering like (type is OBJECT_TYPE_DEFINITION and contains a 'model' directive).

packages/amplify-schema-validator/tsconfig.json
```diff
@@ -0,0 +1,13 @@
...
+    "target": "es5",
+    "strict": true,
+    "sourceMap": true,
+    "esModuleInterop": true,
+    "declaration": true,
+    "outDir": "./dist"
+  },
+  "include": ["src"],
+  "exclude": ["src/__tests__"]
+}
```
sundersc, line 13: Can't open the yarn.lock file so adding the comment here. Did you run `yarn setup-dev`? If it is so, we don't check in that version of yarn.lock file. (I'm think we haven't documented that clearly in contribution page).

I don't think yarn file should contain these many changes considering the number of new dependencies added. Can you do `yarn clean && yarn` and give a try?

sundersc (commented):
Left few comments.

It looks like we are validating only the schema file. Some validations need metadata about the api like configured authorization modes and datastore is enabled or not. For example, customer may provide a schema with OIDC @auth rule and not enabled OIDC authorization. In this case, push would fail.

Though we are not adding validation for these cases now, consider passing the metadata to the `validator` for future enhancements.

packages/amplify-schema-validator/src/exceptions/invalid-directive-error.ts
alharris-at: Unfortunately, no. The intent here is not to pollute the dep chain w/ our FS and CDK dependencies so this can run in the Studio UI

packages/amplify-schema-validator/src/helpers/get-object-with-name.ts
johnpc: This particular function is not a validator - just a helper to fetch objects from a graphql document.

What do you mean mean by the checking for @model? Should that be a different validator, or do you mean something else?

packages/amplify-schema-validator/src/validators/index-scalar-types.ts
johnpc: Ideally these validators would be single-purpose to keep each validation rule as simple and discrete as possible.

Perhaps this should be it's own future validator?

packages/amplify-schema-validator/tsconfig.json
johnpc: I did re-run `yarn` per @alharris-at feedback. I just now ran `yarn clean && yarn` and there is no diff.

johnpc:
> Left few comments.
> 
> It looks like we are validating only the schema file. Some validations need metadata about the api like configured authorization modes and datastore is enabled or not. For example, customer may provide a schema with OIDC @auth rule and not enabled OIDC authorization. In this case, push would fail.
> 
> Though we are not adding validation for these cases now, consider passing the metadata to the `validator` for future enhancements.

I'm interested in chatting more about this, as I'm not sure exactly what the ramifications will be from a browser context where may not readily have this information. We should define the structure of the metadata object that the caller is expected to pass in, and ensure all the data is available to caller before we require it.

I don't think this is blocking, at least for these initial validators, and I'm comfortable adding a new argument later when we figure out the details.

alharris-at (approved):

sundersc (approved):
LGTM.

This looks good for the initial release. Validations based on authorization modes and datastore can be added in the future as required.

sundersc (approved):

alharris-at (approved):

PR #1188, opened 2023-01-24
feat(graphql): factory pattern vtl generator for model directive

sundersc:
PR for Model directive with VTL Generator factory pattern.

- Removes the `@rdsmodel` directive introduced in the early work.
- Removes the base and child class design for Model transformer and restores the existing class to handle both DDB and RDS.
- Adds ModelVTLGenerator classes for DDB and RDS implementation.
- Adds a factory method to construct the VTL generator based on the modelToDatasource mapping from transformer context.
- If mapping is not defined in modelToDatasource, fallback to Greenfield DDB behavior.

This PR is intended to remove the multiple class structure for Model directive. This doesn't cover the following cases which needs to be taken care in the future PRs.
- Correctness of VTLs generated for RDS. Current RDS VTLs are just place holders and not tested E2E.
- Type checking on the ModelVTLGenerator methods can be improved.

**Why are we removing the multiple classes per DB type pattern?**

* Triples the number of classes/packages we have to manage. Also, if we decide to support different DB types in future like MongoDB, QLDB (Amazon Quantum Ledger Database), Graph database etc, the number of packages will keep growing.
* We pass the list of transformer instances to GraphQLTransform step. So, introducing new packages means we need to perform the step twice (once with DynamoDB transformer instances and once with RDS transformer instances).
* Compiling the schema separately for DynamoDB and RDS introduces an additional step of reconciling the output which adds too much complexity to the implementation.
* Amplify allows customers to plugin a custom GraphQL transformer. If we separate the GraphQLTransform step, then we will need to provide a way for customers to pass the transformer list for DynamoDB and RDS separately.
* Makes it difficult to implement proper error messages (DX) in certain cases. For example, if customer is using searchable directive in  RDS schema, we should throw an error as it is not supported. If we are using existing packages, then it is just a matter of checking the DBType property for the model and throw a user friendly error. Otherwise, we will get a very generic error saying unknown directive ‘searchable’ which will confuse customers.
* Makes extensibility difficult for future projects. For example, in future if we want to allow DDB and RDS types to be defined in the same file or allow relationship between DDB and RDS types then it is much easier if we do the compilation in single run.


alharris-at:
General comment, and doubly important because this seems to be a change that introduces some structural and pattern changes to our code.

1) I'd love a more comprehensive description of the change (almost a lightweight version of your design doc explaining what the change is, why it's being made) both  for us, as well as for posterity when we go and look at the change itself.

2) It's a pretty large PR (in terms of lines changed), but I think I grok why it's this large. What would help my understand a bit more is a description of whether this could be released more incrementally, and the process you took to refactor towards a single factory. This might be addressed by point 1 as well.

Still going through the code, but wanted to drop that feedback earlier.

packages/amplify-graphql-model-transformer/src/resolvers/generator/vtl-generator.ts
```diff
@@ -0,0 +1,21 @@
...
+  generateDeleteRequestTemplate(modelName: string, isSyncEnabled: boolean, config?: OperationConfig): string;
+  generateUpdateInitSlotTemplate(modelConfig: ModelDirectiveConfiguration, config?: OperationConfig): string;
+  generateGetRequestTemplate(config?: OperationConfig): string;
+  generateGetResponseTemplate(isSyncEnabled: boolean, config?: OperationConfig): string;
+  generateListRequestTemplate(config?: OperationConfig): string;
+  generateSyncRequestTemplate(config?: OperationConfig): string;
+  generateSubscriptionRequestTemplate(config?: OperationConfig): string;
+  generateSubscriptionResponseTemplate(config?: OperationConfig): string;
+  generateDefaultResponseMappingTemplate(isSyncEnabled: boolean, mutation: boolean): string;
+}
```
alharris-at, line 21: qq: can modelName, modelConfig, and isSyncEnabled be pulled out into a constructor? what is really needed to make a decision on a per-field basis? Also, why is the OperationConfig optional all the way down? When do we not know the operation and operationName

packages/amplify-graphql-model-transformer/src/resolvers/generator/rds-vtl-generator.ts
```diff
@@ -0,0 +1,48 @@
...
+import {
+  generateDefaultLambdaResponseMappingTemplate,
+  generateGetLambdaResponseTemplate,
+  generateLambdaRequestTemplate,
+} from '../rds';
+
+// TODO: This class is created only to show the class structure. This needs a revisit to generate correct resolvers for RDS.
+export class RDSModelVTLGenerator implements ModelVTLGenerator {
+  generateUpdateRequestTemplate(modelName: string, isSyncEnabled: boolean, config?: OperationConfig | undefined): string {
+    return generateLambdaRequestTemplate(modelName, config?.operation!, config?.operationName!);
```
alharris-at, line 12: Yeah, see my comment below, but the config?.operation! seems like an anti-pattern, our types aren't protecting us here.

packages/amplify-graphql-model-transformer/src/resolvers/rds/resolver.ts
```diff
@@ -1,3 +1,5 @@
+// TODO: Split this file into seperated query, mutations and subscriptions files.
```
alharris-at, line 1: Is there a task tracking this todo?

sundersc: Currently the RDSModelVTLGenerator is just a place holder, the correctness of these RDS specific VTLs aren't tested. This should be handled in the PR for Model directive.

packages/amplify-graphql-model-transformer/src/resolvers/generator/vtl-generator.ts
sundersc: 100% agreed. This can be improved. I carried forward the existing argument list and added anything that is additionally required for RDS to `config` argument. But this can be improved something like the below, will take care of it in another PR (I'm still not sure how we can use a single type structure for all these methods as the argument varies for each generate* method).

```ts
type RDSModelConfig {
...
}

type DynamoDBModelConfig {
...
}

type ModelConfig = RDSModelConfig | DynamoDBModelConfig;
```

If single config object is not sufficient, then the best approach is to create a config type based for each operation like the below so that the methods are strongly typed.

generateUpdateRequestTemplate(config: ModelConfig): string;

```ts
type RDSModelCreateRequestConfig {
...
}

type DynamoDBModelCreateRequestConfig {
...
}

type ModelCreateRequestConfig = RDSModelCreateRequestConfig | DynamoDBModelCreateRequestConfig;

generateUpdateRequestTemplate(config: ModelCreateRequestConfig): string;

// Similarly create config types for other operations

```

packages/amplify-graphql-model-transformer/src/resolvers/generator/rds-vtl-generator.ts
sundersc: The arguments typing can be improved as stated in the above comment in a separate PR.

packages/amplify-graphql-model-transformer/src/resolvers/generator/vtl-generator.ts
alharris-at: > I'm still not sure how we can use a single type structure for all these methods as the argument varies for each generate* method

I think that's not actually something we need. I'd prefer to have per-method strongly typed props, rather than a single `ConfigOptions` shape that doesn't guarantee correctness for individual methods.

sundersc: I have added a task to RDS project for this.

sundersc:
> 

I have updated the description with more details about why we are making this change.

The change is minimal when compared to the current main branch. The differences shown on this PR are because of the previous split work (got merged to feature branch) done to separate out DynamoDB and RDS transformers into different packages. I reverted the files to how it is in main today and made the changes on top of it.

phani-srikar (approved):
Looks good, as discussed in the meeting.

alharris-at (approved):

Issue #1192, opened 2023-01-24
Per model configuration not being honoured


vishal-dms:
### How did you install the Amplify CLI?

npm

### Amplify CLI Version

10.5.1 or latest as well

### What operating system are you using?

Mac

### Did you make any manual changes to the cloud resources managed by Amplify? Please describe the changes made.

No

### Describe the bug

[Per model configuration](https://docs.amplify.aws/lib/datastore/conflict/q/platform/js/#per-model-configuration), updated through Amplify CLI is not being applied to the respective models (which still continue to use default configuration)

### Expected behavior

Amplify CLI should be able to deploy **Per model configuration** as documented 

### Reproduction steps

1. `amplify add api` - Push any two models
2. `amplify update api` - For the second model override the conflict strategy to **Optimistic Concurrency**  
```
Do you want to override default per model settings? Yes
? Select the models from below: Note
? Select the resolution strategy for Note model Optimistic Concurrency
```
3. Verify from AppSync console as well the model.json files from within _api/api_name/build/stacks/model_name.json_ to verify that this model's resolvers still have default conflict resolution strategy instead of overridden one

### Project Identifier

033d150b45772ea4e67cf4ee6650fc8c

### Log output

<details>

```
# Put your logs below this line


```

</details>

### Additional information

We currently have to look into [overriding](https://docs.amplify.aws/cli/graphql/override/) the sync config for such models using `amplify override api` and then manually update the _syncConfig_ for the desired model's resolvers

### Before submitting, please confirm:

- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.
- [X] I have removed any sensitive information from my code snippets and submission.

chrisbonifacio:
Hi @Dennis-Dekker :wave:

I will re-open this issue and reach out to the team to see what we might've missed in the previous fix.

Dennis-Dekker:
@chrisbonifacio snklip (fellow discord user) and I have been trying out some more different configurations and found some insights that may help you fix this problem, you can find this on this AWS Amplify discord thread: https://discord.com/channels/705853757799399426/1076342999421816892
Feel free to reach out to us here or on Discord if you need more info!

Dennis-Dekker:
@chrisbonifacio Any updates on this issue?

PR #1201, opened 2023-01-25
fix(transformer): conflict detection respects to per model rule

AaronZyLee:
#### Description of changes

Fix the bug that per model conflict resolution is not respected
The `SyncConfig` is now correctly generated based on per model rule.

#### Issue #, if available
Fix https://github.com/aws-amplify/amplify-category-api/issues/1192

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/packages/amplify-graphql-transformer-core/src/transformer-context/resolver.ts
+++ b/packages/amplify-graphql-transformer-core/src/transformer-context/resolver.ts
@@ -278,7 +278,10 @@ export class TransformerResolver implements TransformerResolverProvider {
           }
 
           if (context.isProjectUsingDataStore()) {
-            const syncConfig = SyncUtils.getSyncConfig(context, this.typeName)!;
+            //Remove the suffix "Table" from the datasource name
+            //The stack name cannot be retrieved as during the runtime it is tokenized and value not being revealed
+            const modelName = this.datasource.name.slice(0, -5);
+            const syncConfig = SyncUtils.getSyncConfig(context, modelName)!;
             const funcConf = dataSourceProviderFn.node.children.find(
               (it: any) => it.cfnResourceType === 'AWS::AppSync::FunctionConfiguration',
             ) as CfnFunctionConfiguration;
--- a/packages/amplify-e2e-core/src/categories/api.ts
+++ b/packages/amplify-e2e-core/src/categories/api.ts
@@ -409,6 +409,42 @@ export function updateApiConflictHandlerType(
   });
 }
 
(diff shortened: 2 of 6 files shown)


packages/amplify-graphql-transformer-core/src/transformer-context/resolver.ts
```diff
@@ -278,7 +278,10 @@ export class TransformerResolver implements TransformerResolverProvider {
           }
 
           if (context.isProjectUsingDataStore()) {
-            const syncConfig = SyncUtils.getSyncConfig(context, this.typeName)!;
+            //Remove the suffix "Table" from the datasource name
```
AaronZyLee, line 281: One thing to illustrate is that the previous `this.typeName` is passed as `Query`, `Mutation`, `Subscription` instead of the model name, in which it will never get the model specific sync config.

Due to this, changing it to the model name is the right solution. However, I cannot find a good way to achieve this because
1. This model name is not passed into the class
2. The stack name is one possible way but it is tokenized and its value is not revealed by CDK synthesis process.

The fix is not final here. Please LMK if there are better approaches to pass in the model name values.

alharris-at: I wonder if we can instead model the resolverConfig?.models as a CloudFormation mapping https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/mappings-section-structure.html and let CFN do the effective lookup.

Then the fact that you're dealing with a Token shouldn't matter I don't think.

AaronZyLee: I paste the source code for creating the model table as datasource below. This string manipulation is subject to change once the following code is changed.
https://github.com/aws-amplify/amplify-category-api/blob/9a9b5e929de8126da2c6c058bbf93b79bf8db81c/packages/amplify-graphql-model-transformer/src/graphql-model-transformer.ts#L1272-L1290

phani-srikar: Like Al mentioned above, try to see if we can use `resolverConfig?.models` and if you think that's not working, I am okay with shipping it to unblock the customers.

phani-srikar (approved):

alharris-at (approved):

Dennis-Dekker:
Hi, great to hear you're tackling this issue! Our product requires this feature, do you have a rough estimate of when the solution will be available? Appreciate your response.

PR #1203, opened 2023-01-26
feat: RDS generate schema handles default values

phani-srikar:
#### Description of changes

There are broadly 2 types of Default values that could be specified in MySQL customer's DDL:
- **Literal constants:** These are straight forward constant defaults for scalar types. The fields with these defaults are generated as optionals and `@default` directive is attached to them.
- **Computed values:** These are constructed using built-in scalar functions and are computed by the database. The current `@default` directive does not support this. The fields with these defaults are however still generated as optionals but without the `@default` directive.

[Refer MySQL docs for more info](https://dev.mysql.com/doc/refman/8.0/en/data-type-defaults.html)

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-graphql-schema-generator/src/__tests__/__snapshots__/mysql-datasource-adapter.test.ts.snap
+++ b/packages/amplify-graphql-schema-generator/src/__tests__/__snapshots__/mysql-datasource-adapter.test.ts.snap
@@ -26,7 +26,7 @@ exports[`testDataSourceAdapter test generate graphql schema from internal repren
 }
 
 type Country @model {
-  id: Int @primaryKey
+  id: Int! @primaryKey
   name: String
 }
 
--- a/packages/amplify-graphql-schema-generator/src/__tests__/__snapshots__/mysql-datasource-adapter.test.ts.snap
+++ b/packages/amplify-graphql-schema-generator/src/__tests__/__snapshots__/mysql-datasource-adapter.test.ts.snap
@@ -1,5 +1,23 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`testDataSourceAdapter generates a default directive and optional types for fields with literal default values 1`] = `
+"type Account @model {
+  id: Int! @primaryKey
+  serialNumber: Int @default(value: \\"-1\\")
+  ownerName: String @default(value: \\"na\\")
+  amount: Float @default(value: \\"101.101\\")
+}
+"
+`;
+
+exports[`testDataSourceAdapter generates optional type but no default directive for fields with computed default values 1`] = `
+"type Account @model {
+  id: Int! @primaryKey
+  computed: Float
+}
+"
+`;
+
 exports[`testDataSourceAdapter test generate graphql schema from internal reprensentation 1`] = `
 "type Capital @model {
   id: Int! @primaryKey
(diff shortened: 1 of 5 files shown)


packages/amplify-graphql-schema-generator/src/__tests__/mysql-datasource-adapter.test.ts
```diff
@@ -95,7 +95,7 @@ describe("testDataSourceAdapter", () => {
 
     model = new Model("Country");
     const countryIdField = new Field("id", { "kind": "NonNull", "type": { "kind": "Scalar", "name": "Int" } });
-    countryIdField.default = { kind: "DB_GENERATED", value: "uuid()" };
+    countryIdField.default = { kind: "DB_GENERATED", value: "(uuid())" };
```
sundersc, line 98: Just now checked on a MySQL instance, COLUMN_DEFAULT shows as `uuid()` without the brackets? Does it differ between MySQL versions?

image

packages/amplify-graphql-schema-generator/src/__tests__/__snapshots__/mysql-datasource-adapter.test.ts.snap
```diff
@@ -8,7 +26,7 @@ exports[`testDataSourceAdapter test generate graphql schema from internal repren
 }
 
 type Country @model {
-  id: Int! @default(value: \\"uuid()\\") @primaryKey
+  id: Int @primaryKey
```
sundersc, line 29: Shouldn't the primary key fields required? Currently MySQL doesn't support `RETURNING` so we can't retrieve the last inserted record. For this reason, amplify should assign a value to id fields if customer didn't pass on the input.

marcvberg: I think that could cause problems for some database designs couldn't it? If the customer has set up an auto-incrementing default value as the primary key, they may actually rely on that auto-incrementing design. Maybe that's just a tradeoff here and we won't be able to reasonably solve it with current platform(s) but it's a thought

packages/amplify-graphql-schema-generator/src/__tests__/mysql-datasource-adapter.test.ts
phani-srikar: I've double checked this one and its not related to MySQL version difference. The MySQL doc meant that in the table DDL we need to have these computed values wrapped in `()` like for example, `DEFAULT (RAND())`. But when we query the table info as we do in above snippet, it trims away the `()` if it's a single function and retains if it's more complex, for example, `(RAND() * RAND())`. So, in short I need to update the method that deduces if it's a computed default to handle such cases.

packages/amplify-graphql-schema-generator/src/__tests__/__snapshots__/mysql-datasource-adapter.test.ts.snap
phani-srikar: Updated this case to retain the primary key fields as required, even if there is an auto increment default attached to it. This will align with the current behavior of CPK where they are required as input from the client. [Amplify doc link for CPK](https://docs.amplify.aws/cli/graphql/data-modeling/#configure-a-primary-key).

sundersc (approved):

marcvberg (approved):

PR #1208, opened 2023-01-28
fix: update belongs to relation required logic to handle nested schemas

johnpc:
#### Description of changes

This PR adds a fix to the @aws-amplify/graphql-schema-validation package to support nested schemas

#### Issue #, if available

N/A

Unit tests added

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-schema-validator/src/helpers/resolve-field-type-name.ts
+++ b/packages/amplify-schema-validator/src/helpers/resolve-field-type-name.ts
@@ -1,0 +1,36 @@
+import { NamedTypeNode, TypeNode } from 'graphql';
+
+/**
+ * Gets the name of of a graphql TypeNode
+ *
+ * @param type graphql TypeNode
+ * @returns string|undefined
+ */
+export const resolveFieldTypeName = (type: TypeNode | undefined): string | undefined => {
+  switch (type?.kind) {
+    case 'NamedType': {
+      return type.name.value;
+    }
+    case 'NonNullType':
+      if (type.type.kind === 'NamedType') {
+        return type.type.name.value;
+      } if (type.type.kind === 'ListType') {
+        if (type.type.type.kind === 'NamedType') {
+          return type.type.type.name.value;
+        } if (type.type.type.kind === 'NonNullType') {
+          return (type.type.type.type as NamedTypeNode).name.value;
+        } return undefined;
+      } return undefined;
+
+    case 'ListType': {
+      if (type.type.kind === 'NamedType') {
+        return type.type.name.value;
+      } if (type.type.kind === 'NonNullType') {
+        return (type.type.type as NamedTypeNode).name.value;
+      } return undefined;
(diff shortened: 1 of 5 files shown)


sundersc (approved):

packages/amplify-schema-validator/src/helpers/resolve-field-type-name.ts
```diff
@@ -0,0 +1,30 @@
+import { NamedTypeNode, TypeNode } from 'graphql';
+
+export const resolveFieldTypeName = (type: TypeNode | undefined): string => {
```
alharris-at, line 3: Could you include comments on the methods here as well?

packages/amplify-schema-validator/src/helpers/resolve-field-type-name.ts
```diff
@@ -0,0 +1,30 @@
...
+    case 'ListType': {
+      if (type.type.kind === 'NamedType') {
+        return type.type.name.value;
+      } if (type.type.kind === 'NonNullType') {
+        return (type.type.type as NamedTypeNode).name.value;
+      } return '';
+    }
+    default: {
+      return '';
+    }
```
alharris-at, line 28: Is a string a reasonable sentinel value here, or should we return an undefined here?

alharris-at (commented):
Question/nit on naming. I'm not sure what you mean nested in this test case, etc. Is this a model that is this like A hasMany B and B hasMany C? or is it like A manyToMany A (which I'd refer to probably as recursive instead)

packages/amplify-schema-validator/src/helpers/resolve-field-type-name.ts
johnpc: I prefer undefined - it's more explicit. updated!

packages/amplify-schema-validator/src/validators/relation-required-with-belongs-to.ts
```diff
@@ -24,10 +24,14 @@ export const validateRequireBelongsToRelation = (schema: DocumentNode): Error[]
...
     belongsToFields?.forEach((belongsToField) => {
-      const typeName = (belongsToField.type as NamedTypeNode)?.name.value;
+      const typeName = resolveFieldTypeName(belongsToField.type);
+      if (!typeName) {
+        /* istanbul ignore next */
+        return;
+      }
       const objectOfType = getObjectWithName(schema, typeName);
       const relationField = objectOfType?.fields?.find(
-        (field) => (field.type as NamedTypeNode)?.name?.value === objectName,
```
johnpc, line 30: the primary change in this PR is replacing this line, which assumes a `NamedTypeNode` to instead call a function to get the type name and works for `ListTypeNode` and `NonNullTypeNode` objects as well.

alharris-at (approved):

PR #1209, opened 2023-01-30
feat: add a utility to convert an input graphQL filter into an RDS query expression

manaswi223:
#### Description of changes
This PR adds functionality primarily on two things:
1. adds a utility helper method on [preview 1 readiness] RDS feature support for Amplify that converts the input graphql filter to the AWS RDS or mysql query expressions that maps the fields in a schema type to a data source - AWS RDS resulting in a mySQL query.

2. converts the resultant RDS filter expression to a knex filter type so it'll be agnostic to different DB engines

For example:
A graphQL query passed as an input argument to the helper method `toRDSQueryExpression` results a mapped mySQL query as shown below:

**Input: ( a GraphQL query)**

```
const inputVariables = {
  id: '1',
  name: 'Amplify',
  age: 20,
  address: {
    street: '123 Main St',
    city: 'Seattle',
    state: 'WA',
    zip: '98101',
  },
  filter : {  
    and: [      
      { eq: { id: '1' } },
      { startsWith: { name: 'Amplify' } }
    ],
  },
};
```
**passing the input filter to the helper method:**

```
const rdsQueryExpression = toRDSQueryExpression(inputVariables.filter);

```

**Output: (an RDS expression)**

```
console.log(rdsQueryExpression);

id = 1 AND name = "Amplify%"
```

This PR supports input GraphQL filters' `QueryGroup` & `Operators` as well as Nested QueryGroups with Nested Operators (complexFilter):

**QueryGroup** = [`and`, `or`, `not`]
**Operators** = [`attributeExists`, `beginsWith`, `between`, `contains`, `eq`, `ge`, `gt`, `le`, `lt`, `ne`, `notContains`]

**Nested QueryGroups with NestedOperators:**
```
const complexFilter = {
    and : [
        { id : { eq : '123' } },
        { or : [
            { name : { eq : 'Amplify' } },
            { and : [
                { org : { eq : 'AWS' } },
                { age : { between : [`18`, `160`] } },
            ]},
        ]},
    ],  
};
```
**passing the inputFilter argument to utility method `toRDSQueryExpression`:**

```
const rdsExpression = toRDSQueryExpression(complexFilter);

```

**Output:**

```
console.log(rdsExpression);

((id = '123') AND ((name = 'Amplify') OR ((org = 'AWS') AND (age BETWEEN '18' AND '160'))))

```

With this PR, the resultant query expression - `rdsExpression` also converts to a knex filter type so it'll be agnostic to different DB engines. 
For example, this can be consumed by the knex libraries to perform and fetch filter queries of rows from the actual database.

```
knex.select().from('employee_details').whereRaw(`column = ? `, [rdsExpression]).then(rows => {
            console.log(rows);
});
```

where `employee_details` is a db_table with rows & columns.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-graphql-model-transformer/src/resolvers/common.ts
```diff
@@ -34,3 +34,54 @@ export const generateAuthExpressionForSandboxMode = (enabled: boolean): string =
...
+        } else if (key === 'lt') {
+          subExpressions.push(` < '${filterValue}'`);
+        } else if (key === 'le') {
+          subExpressions.push(` <= '${filterValue}'`);
+        } else if (key === 'gt') {
+          subExpressions.push(` > '${filterValue}'`);
+        } else if (key === 'ge') {
+          subExpressions.push(` >= '${filterValue}'`);
+        } else if (key === 'between') {
+          subExpressions.push(` between '${filterValue[0]}' and '${filterValue[1]}'`);
```
sundersc, line 63: We should validate filterValue is an array and contains 2 elements.

packages/amplify-graphql-model-transformer/src/resolvers/common.ts
```diff
@@ -34,3 +34,54 @@ export const generateAuthExpressionForSandboxMode = (enabled: boolean): string =
...
+        } else if (key === 'attributeNotExists') {
+          subExpressions.push(` is null`);
+        } else if (key === 'in') {
+          subExpressions.push(` in (${filterValue.join(',')})`);
+        } else if (key === 'notContains') {
+          subExpressions.push(` not like '%${filterValue}%'`);
+        } else if (key === 'notIn') {
+          subExpressions.push(` not in (${filterValue.join(',')})`);
+        } else if (key === 'and' || key === 'or') {
+          const subExpression = toRDSFilterExpression(filterValue);
```
sundersc, line 79: Can we add test cases for filters with 'and' and 'or' conditions?

sundersc (changes requested):
Need tests with how we can chain these conditions with knex query.

packages/amplify-graphql-model-transformer/src/resolvers/common.ts
```diff
@@ -34,3 +34,54 @@ export const generateAuthExpressionForSandboxMode = (enabled: boolean): string =
 export const generateResolverKey = (typeName: string, fieldName: string): string => {
   return `${typeName}.${fieldName}`;
 };
+
+/**
+ * Util function to convert any GraphQL input filter argument to an AWS RDS query expression 
+ */
+export const toRDSFilterExpression = (filter: any): string => {
```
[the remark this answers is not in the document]
manaswi223: < will add this code on RDS Lambda in next revision, waiting to get more context >

manaswi223:
Todo:
1. Discuss `NOT` graphQL filter equivalent in RDS Query expression
2. Implement `size` 
3. Write test cases on knex: Test the resultant expression (output of the utility method) - `toRDSQueryExpression` as an input feed to the knex.js libraries & verify it to be in a Knex.js compatible format - agnostic to any DB engines.

marcvberg:
Before you commit, you'll need to run `yarn api-extract` to update the API.md files

marcvberg:
We can't build the queries by just substituting into strings/string concatenation, we'll have to use parameterization to block SQL injection on filter arguments 

`.fromRaw('(select * from "users" where "age" > ?)', '18')`

You can find docs on it [here](https://knexjs.org/guide/query-builder.html#fromraw), best way to handle it would probably be to return the string query and a list of all the arguments in order. Then we can deconstruct that list to pass as the arguments to the knex query

manaswi223:
So to answer Marc's comment:

I think to prevent sql injection situations, as Marc mentioned, we can parameterize the query like:
while using any knex method, for example: `.whereRaw` here:

```
knex.select().from('employee_details').whereRaw(`column = ? `, [rdsExpression]).then(rows => {
            console.log(rows);
});
```

this can separate the actual sql query from the "data" & hence, prevent any malicious sql injection

manaswi223:
closing this in favor of #1236

PR #1214, opened 2023-01-30
Model resource generator

marcvberg:
#### Description of changes
Adds resource generator abstraction to model transformer
VTL generation is also brought behind the resource generators as the resolvers are part of the resources

This is a pretty significant refactor of the model transformer, but functionally should not impact anything. This is done to make it easier to add RDS support

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-graphql-model-transformer/src/graphql-model-transformer.ts
+++ b/packages/amplify-graphql-model-transformer/src/graphql-model-transformer.ts
@@ -399,10 +399,9 @@ export class ModelTransformer extends TransformerModelBase implements Transforme
     fieldName: string,
     resolverLogicalId: string,
   ): TransformerResolverProvider => {
-    const dbInfo = { dbType: 'DDB', provisioned: true };
-    if (dbInfo?.dbType && this.resourceGeneratorMap.has(dbInfo.dbType)) {
+    if (this.resourceGeneratorMap.has(DDB_DB_TYPE)) {
       // Coercing this into being defined as we're running a check on it first
-      return this.resourceGeneratorMap.get(dbInfo.dbType)!.generateOnCreateResolver(ctx, typeName, fieldName, resolverLogicalId);
+      return this.resourceGeneratorMap.get(DDB_DB_TYPE)!.generateOnCreateResolver(ctx, typeName, fieldName, resolverLogicalId);
     }
     throw new Error('Resource generator not provided for DDB');
   };
@@ -413,10 +412,9 @@ export class ModelTransformer extends TransformerModelBase implements Transforme
     fieldName: string,
     resolverLogicalId: string,
   ): TransformerResolverProvider => {
-    const dbInfo = { dbType: 'DDB', provisioned: true };
-    if (dbInfo?.dbType && this.resourceGeneratorMap.has(dbInfo.dbType)) {
+    if (this.resourceGeneratorMap.has(DDB_DB_TYPE)) {
       // Coercing this into being defined as we're running a check on it first
-      return this.resourceGeneratorMap.get(dbInfo.dbType)!.generateOnUpdateResolver(ctx, typeName, fieldName, resolverLogicalId);
+      return this.resourceGeneratorMap.get(DDB_DB_TYPE)!.generateOnUpdateResolver(ctx, typeName, fieldName, resolverLogicalId);
     }
     throw new Error('Resource generator not provided for DDB');
   };
@@ -427,10 +425,9 @@ export class ModelTransformer extends TransformerModelBase implements Transforme
     fieldName: string,
     resolverLogicalId: string,
   ): TransformerResolverProvider => {
-    const dbInfo = { dbType: 'DDB', provisioned: true };
-    if (dbInfo?.dbType && this.resourceGeneratorMap.has(dbInfo.dbType)) {
+    if (this.resourceGeneratorMap.has(DDB_DB_TYPE)) {
       // Coercing this into being defined as we're running a check on it first
-      return this.resourceGeneratorMap.get(dbInfo.dbType)!.generateOnDeleteResolver(ctx, typeName, fieldName, resolverLogicalId);
+      return this.resourceGeneratorMap.get(DDB_DB_TYPE)!.generateOnDeleteResolver(ctx, typeName, fieldName, resolverLogicalId);
     }
     throw new Error('Resource generator not provided for DDB');
   };
--- a/packages/amplify-graphql-model-transformer/src/graphql-model-transformer.ts
+++ b/packages/amplify-graphql-model-transformer/src/graphql-model-transformer.ts
@@ -5,12 +5,9 @@ import {
   getFieldNameFor,
   InputObjectDefinitionWrapper,
   InvalidDirectiveError,
-  MappingTemplate,
   ObjectDefinitionWrapper,
-  SyncConfig,
   SyncUtils,
   TransformerModelBase,
-  TransformerNestedStack,
   DatasourceType,
 } from '@aws-amplify/graphql-transformer-core';
 import {
@@ -30,17 +27,8 @@ import {
   TransformerValidationStepContextProvider,
 } from '@aws-amplify/graphql-transformer-interfaces';
 import {
-  AttributeType,
-  CfnTable,
   ITable,
-  StreamViewType,
-  Table,
-  TableEncryption,
 } from '@aws-cdk/aws-dynamodb';
-import * as iam from '@aws-cdk/aws-iam';
-import { CfnRole } from '@aws-cdk/aws-iam';
-import * as cdk from '@aws-cdk/core';
-import { CfnDataSource } from '@aws-cdk/aws-appsync';
 import {
   DirectiveNode,
   FieldDefinitionNode,
@@ -58,11 +46,6 @@ import {
   makeNamedType,
   makeNonNullType,
   makeValueNode,
-  ModelResourceIDs,
-  ResolverResourceIDs,
-  ResourceConstants,
-  SyncResourceIDs,
-  toCamelCase,
   toPascalCase,
 } from 'graphql-transformer-common';
 import {
@@ -81,13 +64,15 @@ import {
   propagateApiKeyToNestedTypes,
 } from './graphql-types';
 import {
-  generateAuthExpressionForSandboxMode,
-  generateResolverKey,
   DynamoDBModelVTLGenerator,
   RDSModelVTLGenerator,
 } from './resolvers';
 import { API_KEY_DIRECTIVE } from './definitions';
 import { ModelDirectiveConfiguration, SubscriptionLevel } from './directive';
+import { ModelResourceGenerator } from './resources/model-resource-generator';
+import { DynamoModelResourceGenerator } from './resources/dynamo-model-resource-generator';
+import { RdsModelResourceGenerator } from './resources/rds-model-resource-generator';
+import { ModelTransformerOptions } from './types';
 
 /**
  * Nullable
@@ -127,10 +112,10 @@ export const directiveDefinition = /* GraphQl */ `
   }
 `;
 
-type ModelTransformerOptions = {
-  EnableDeletionProtection?: boolean;
-  SyncConfig?: SyncConfig;
-};
+const MYSQL_DB_TYPE = 'MySQL';
+const DDB_DB_TYPE = 'DDB';
+
+const DDB_DATASOURCE_TYPE = { dbType: DDB_DB_TYPE, provisioned: true };
 
 /**
  * ModelTransformer
@@ -141,6 +126,8 @@ export class ModelTransformer extends TransformerModelBase implements Transforme
   private ddbTableMap: Record<string, ITable> = {};
   private resolverMap: Record<string, TransformerResolverProvider> = {};
   private typesWithModelDirective: Set<string> = new Set();
+  private resourceGeneratorMap: Map<string, ModelResourceGenerator> = new Map<string, ModelResourceGenerator>();
+  private modelToDatasourceMap: Map<string, DatasourceType> = new Map<string, DatasourceType>()
   /**
    * A Map to hold the directive configuration
    */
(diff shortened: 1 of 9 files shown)


packages/amplify-graphql-model-transformer/src/resources/dynamo-model-resource-generator.ts
```diff
@@ -0,0 +1,301 @@
...
+
+    // add the data source
+    context.dataSources.add(def!, dataSource);
+    this.datasourceMap[def!.name.value] = dataSource;
+  }
+
+  /**
+   * createIAMRole
+   */
+  createIAMRole = (context: TransformerContextProvider, def: ObjectTypeDefinitionNode, stack: cdk.Stack, tableName: string): iam.Role => {
```
sundersc, line 242: Should this be 'private'? Also, use the same function style for the class methods.

packages/amplify-graphql-model-transformer/src/resources/model-resource-generator.ts
```diff
@@ -0,0 +1,670 @@
...
+        MappingTemplate.s3MappingTemplateFromString(
+          vtlGenerator.generateDefaultResponseMappingTemplate(responseConfig),
+          `${typeName}.${fieldName}.res.vtl`,
+        ),
+      );
+    }
+    return this.resolverMap[resolverKey];
+  };
+
+  generateCreateResolver = (
```
sundersc, line 289: Can all these methods be `private`?

packages/amplify-graphql-model-transformer/src/resources/rds-model-resource-generator.ts
```diff
@@ -0,0 +1,21 @@
...
+import { ModelResourceGenerator } from './model-resource-generator';
+import {DatasourceType} from '@aws-amplify/graphql-transformer-core';
+import {DynamoDBModelVTLGenerator, ModelVTLGenerator, RDSModelVTLGenerator} from '../resolvers';
+
+/**
+ * An implementation of ModelResourceGenerator responsible for generated CloudFormation resources
+ * for models backed by an RDS data source
+ */
+export class RdsModelResourceGenerator extends ModelResourceGenerator {
+  protected generatorType = 'DynamoModelResourceGenerator';
```
sundersc, line 12: Shouldn't this be 'RDS'?

packages/amplify-graphql-model-transformer/src/resources/dynamo-model-resource-generator.ts
```diff
@@ -0,0 +1,301 @@
...
+import { CfnDataSource } from '@aws-cdk/aws-appsync';
+import { CfnRole } from '@aws-cdk/aws-iam';
+import { ModelResourceGenerator } from './model-resource-generator';
+import { DynamoDBModelVTLGenerator, ModelVTLGenerator } from '../resolvers';
+
+/**
+ * DynamoModelResourceGenerator is an implementation of ModelResourceGenerator,
+ * providing necessary utilities to generate the DynamoDB resources for models
+ */
+export class DynamoModelResourceGenerator extends ModelResourceGenerator {
```
sundersc, line 27: Where are we utilizing these new resource generator classes?

packages/amplify-graphql-model-transformer/src/resources/model-resource-generator.ts
sundersc: I don't think we need to bring the `generate*Resolver` into resource provider. Isn't the createTable call alone differ between DDB vs RDS?

packages/amplify-graphql-model-transformer/src/graphql-model-transformer.ts
```diff
@@ -127,10 +110,10 @@ export const directiveDefinition = /* GraphQl */ `
   }
 `;
 
-type ModelTransformerOptions = {
-  EnableDeletionProtection?: boolean;
-  SyncConfig?: SyncConfig;
-};
+const MYSQL_DB_TYPE = 'MySQL';
```
phani-srikar, line 113: Probably better to move these constants to transformer core to be used across transformers.

packages/amplify-graphql-model-transformer/src/graphql-model-transformer.ts
```diff
@@ -141,44 +124,38 @@ export class ModelTransformer extends TransformerModelBase implements Transforme
   private ddbTableMap: Record<string, ITable> = {};
   private resolverMap: Record<string, TransformerResolverProvider> = {};
   private typesWithModelDirective: Set<string> = new Set();
+  private resourceGeneratorMap: Map<string, ModelResourceGenerator> = new Map<string, ModelResourceGenerator>();
+  private modelToDatasourceMap: Map<string, DatasourceType> = new Map<string, DatasourceType>()
```
phani-srikar, line 128: isn't this already available via the Transformer context?

packages/amplify-graphql-model-transformer/src/graphql-model-transformer.ts
```diff
@@ -141,44 +124,38 @@ export class ModelTransformer extends TransformerModelBase implements Transforme
...
+      this.resourceGeneratorMap.get(MYSQL_DB_TYPE)?.enableUnprovisioned();
+    }
+    if (datasourceMapValues.length === 0) {
+      // Just enable DynamoDB provisioned, legacy use
+      this.resourceGeneratorMap.get(DDB_DB_TYPE)?.enableGenerator();
+      this.resourceGeneratorMap.get(DDB_DB_TYPE)?.enableProvisioned();
+    }
+    // We only store this in the model because some of the required override methods need to pass through to the
+    // Resource generators, but do not have access to the context
+    this.modelToDatasourceMap = ctx.modelToDatasourceMap;
```
phani-srikar, line 158: avoid the copy if possible unless you have a strong reason

packages/amplify-graphql-model-transformer/src/resources/dynamo-model-resource-generator.ts
```diff
@@ -0,0 +1,301 @@
...
+        description: 'Enable server side encryption powered by KMS.',
+        type: 'String',
+        default: 'true',
+        allowedValues: ['true', 'false'],
+      });
+    }
+
+    this.models.forEach((model) => {
+      // This name is used by the mock functionality. Changing this can break mock.
+      const tableBaseName = ctx.resourceHelper.getModelNameMapping(model!.name.value);
```
phani-srikar, line 68: Do we need the `!` here? Is there a way to fallback to something if the model is undefined?

packages/amplify-graphql-model-transformer/src/resources/dynamo-model-resource-generator.ts
```diff
@@ -0,0 +1,301 @@
...
+ */
+export class DynamoModelResourceGenerator extends ModelResourceGenerator {
+  protected generatorType = 'DynamoModelResourceGenerator';
+
+  generateResources(ctx: TransformerContextProvider): void {
+    if (!this.isEnabled()) {
+      return;
+    }
+
+    if (this.isProvisioned()) {
```
phani-srikar, line 35: I think we're making an assumption here that DDB resources are already provisioned. I would say a better check would be to check the dataSource instead.

packages/amplify-graphql-model-transformer/src/resources/dynamo-model-resource-generator.ts
```diff
@@ -0,0 +1,301 @@
...
+
+/**
+ * DynamoModelResourceGenerator is an implementation of ModelResourceGenerator,
+ * providing necessary utilities to generate the DynamoDB resources for models
+ */
+export class DynamoModelResourceGenerator extends ModelResourceGenerator {
+  protected generatorType = 'DynamoModelResourceGenerator';
+
+  generateResources(ctx: TransformerContextProvider): void {
+    if (!this.isEnabled()) {
```
phani-srikar, line 31: The `isEnabled` name is not intuitive here. Why wouldn't this `generateResources` be enabled?

packages/amplify-graphql-model-transformer/src/resources/rds-model-resource-generator.ts
```diff
@@ -0,0 +1,21 @@
...
+
+  generateResources(ctx: TransformerContextProvider): void {
```
phani-srikar, line 14: Aren't we generating anything here?

phani-srikar (changes requested):

packages/amplify-graphql-model-transformer/src/graphql-model-transformer.ts
```diff
@@ -290,11 +275,13 @@ export class ModelTransformer extends TransformerModelBase implements Transforme
       const subscriptionsFields = this.createSubscriptionFields(ctx, def!);
       ctx.output.addSubscriptionFields(subscriptionsFields);
 
-      // Update the field with auto generatable Fields
-      this.addAutoGeneratableFields(ctx, type);
+      if ((ctx.modelToDatasourceMap.get(def.name.value) ?? DDB_DATASOURCE_TYPE).dbType === DDB_DB_TYPE) {
```
phani-srikar, line 278: can we refactor this into a util method?

packages/amplify-graphql-model-transformer/src/graphql-model-transformer.ts
```diff
@@ -141,44 +124,38 @@ export class ModelTransformer extends TransformerModelBase implements Transforme
...
+  private modelToDatasourceMap: Map<string, DatasourceType> = new Map<string, DatasourceType>()
   /**
    * A Map to hold the directive configuration
    */
   private modelDirectiveConfig: Map<string, ModelDirectiveConfiguration> = new Map();
   constructor(options: ModelTransformerOptions = {}) {
     super('amplify-model-transformer', directiveDefinition);
     this.options = this.getOptions(options);
+    const rdsGenerator = new RdsModelResourceGenerator();
+    this.resourceGeneratorMap.set(DDB_DB_TYPE, new DynamoModelResourceGenerator());
```
sundersc, line 137: Missing consistency in the way we create ResourceGenerator instance.

packages/amplify-graphql-model-transformer/src/graphql-model-transformer.ts
```diff
@@ -607,19 +399,12 @@ export class ModelTransformer extends TransformerModelBase implements Transforme
...
-    if (!this.resolverMap[resolverKey]) {
-      this.resolverMap[resolverKey] = ctx.resolvers.generateSubscriptionResolver(
-        typeName,
-        fieldName,
-        resolverLogicalId,
-        MappingTemplate.s3MappingTemplateFromString(vtlGenerator.generateSubscriptionRequestTemplate(), `${typeName}.${fieldName}.req.vtl`),
-        MappingTemplate.s3MappingTemplateFromString(vtlGenerator.generateSubscriptionResponseTemplate(), `${typeName}.${fieldName}.res.vtl`),
-      );
+    const dbInfo = { dbType: 'DDB', provisioned: true };
+    if (dbInfo?.dbType && this.resourceGeneratorMap.has(dbInfo.dbType)) {
```
sundersc, line 403: Do we need the check `dbInfo?.dbType`? Isn't it always true?

packages/amplify-graphql-model-transformer/src/resources/dynamo-model-resource-generator.ts
sundersc: Does `this.models` refer to DynamoDB models only? How do we filter it?

packages/amplify-graphql-model-transformer/src/resources/dynamo-model-resource-generator.ts
```diff
@@ -0,0 +1,301 @@
...
+    // eslint-disable-next-line no-new
+    new cdk.CfnOutput(stack, datasourceOutputId, {
+      value: dataSource.ds.attrName,
+      description: 'Your model DataSource name.',
+      exportName: cdk.Fn.join(':', [context.api.apiId, 'GetAtt', datasourceRoleLogicalID, 'Name']),
+    });
+
+    // add the data source
+    context.dataSources.add(def!, dataSource);
+    this.datasourceMap[def!.name.value] = dataSource;
```
sundersc, line 236: Where do we design the datasource for RDS models?

packages/amplify-graphql-model-transformer/src/resources/dynamo-model-resource-generator.ts
```diff
@@ -0,0 +1,301 @@
...
+import { CfnRole } from '@aws-cdk/aws-iam';
+import { ModelResourceGenerator } from './model-resource-generator';
+import { DynamoDBModelVTLGenerator, ModelVTLGenerator } from '../resolvers';
+
+/**
+ * DynamoModelResourceGenerator is an implementation of ModelResourceGenerator,
+ * providing necessary utilities to generate the DynamoDB resources for models
+ */
+export class DynamoModelResourceGenerator extends ModelResourceGenerator {
+  protected generatorType = 'DynamoModelResourceGenerator';
```
sundersc, line 28: Should this be read-only?

packages/amplify-graphql-model-transformer/src/resources/model-resource-generator.ts
```diff
@@ -0,0 +1,670 @@
...
+ * ModelResourceGenerator implementations are intended to create resources for the model transformer plugin
+ * according to the data source that backs the model
+ */
+export abstract class ModelResourceGenerator {
+  protected datasourceMap: Record<string, DataSourceProvider> = {};
+  private resolverMap: Record<string, TransformerResolverProvider> = {};
+  protected generatorType = 'ModelResourceGenerator';
+  private enabled = false;
+  private provisioned = false;
+  private unprovisioned = false;
```
sundersc, line 31: What's the purpose of these flags?

packages/amplify-graphql-model-transformer/src/resources/model-resource-generator.ts
```diff
@@ -0,0 +1,670 @@
...
+import { ModelTransformerOptions } from '../types';
+
+/**
+ * Abstract class definition for ModelResourceGenerator
+ * ModelResourceGenerator implementations are intended to create resources for the model transformer plugin
+ * according to the data source that backs the model
+ */
+export abstract class ModelResourceGenerator {
+  protected datasourceMap: Record<string, DataSourceProvider> = {};
+  private resolverMap: Record<string, TransformerResolverProvider> = {};
```
sundersc, line 27: Moving these to DDB and RDS specific instances make it harder to find collisions?

packages/amplify-graphql-model-transformer/src/resources/model-resource-generator.ts
marcvberg: The problem is that once you start creating the table, the process of creating the table accesses many of the attributes of the transformer plugin. And, creating the table populates an attribute that is used by the resolvers. You could force it so that they are still in their original classes but you get a messy coupling between the two

packages/amplify-graphql-model-transformer/src/resources/dynamo-model-resource-generator.ts
marcvberg: it's public for the sake of passing through calls from the model transformer plugin if any are still made, those methods still have to be there to not break the API. Although we may be able to remove them since RDS should be a major version bump

packages/amplify-graphql-model-transformer/src/graphql-model-transformer.ts
marcvberg: It's completely necessary unless we remove the override methods from the `ModelTransformerBase` interface which we might be able to do per an above comment

packages/amplify-graphql-model-transformer/src/resources/dynamo-model-resource-generator.ts
marcvberg: it's filtered during the process of adding to the resource generators, the `!` is a holdover from copying the existing code in the model transformer, I can probably delete it

packages/amplify-graphql-model-transformer/src/resources/dynamo-model-resource-generator.ts
marcvberg: All of the generators are instantiated/populated in the transformer plugin, it enables the ones it needs based on what models are found, I put a function header on the `isEnabled` method

packages/amplify-graphql-model-transformer/src/resources/rds-model-resource-generator.ts
marcvberg: Yeah but the code wasn't in there to refactor because it got deleted during the merge back, this is just a preliminary PR to get the generators in and I'm still working on the rest

packages/amplify-graphql-model-transformer/src/resources/model-resource-generator.ts
marcvberg: Which collisions are we looking for? The schema is parsed as a whole which prevents model collisions, and the resolver manager used in the transformer context is already blocking resolver collisions

packages/amplify-graphql-model-transformer/src/resources/model-resource-generator.ts
marcvberg: I put these in mostly because of looking at what was in the `before` method for the model transformer, we create parameters for managing the CFN values of the tables in DDB, but if someone only had RDS models, we wouldn't need to create those. Or if someone only had brownfield DDB, they wouldn't need those

I wanted a way to prevent creating resources that don't have any purpose

packages/amplify-graphql-model-transformer/src/resources/model-resource-generator.ts
sundersc: Possibly we can create these instances outside resource generator and inject the same instance to both the generators.

marcvberg:
Closing this PR since there's a new one that encapsulates these changes

PR #1219, opened 2023-02-02
feat(graphql): add index directive support for rds datasource

sundersc:
Adds @index directive support for RDS datasource. Import workflow detects the indexes and adds @index on the field. 
- Customer can edit a generated index and add a queryField attribute which will create a GraphQL query with the given name.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-graphql-index-transformer/src/resolvers/resolvers.ts
+++ b/packages/amplify-graphql-index-transformer/src/resolvers/resolvers.ts
@@ -1,5 +1,5 @@
 import { generateApplyDefaultsToInputTemplate } from '@aws-amplify/graphql-model-transformer';
-import { MappingTemplate, GraphQLTransform, SyncUtils, StackManager } from '@aws-amplify/graphql-transformer-core';
+import { MappingTemplate, GraphQLTransform, SyncUtils, StackManager, DatasourceType } from '@aws-amplify/graphql-transformer-core';
 import { DataSourceProvider, StackManagerProvider, TransformerContextProvider, TransformerPluginProvider, TransformerResolverProvider } from '@aws-amplify/graphql-transformer-interfaces';
 import { DynamoDbDataSource } from '@aws-cdk/aws-appsync';
 import { Table } from '@aws-cdk/aws-dynamodb';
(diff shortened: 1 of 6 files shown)


packages/amplify-graphql-index-transformer/src/resolvers/resolvers.ts
```diff
@@ -335,6 +339,11 @@ export function validateSortDirectionInput(config: PrimaryKeyDirectiveConfigurat
  */
 export function appendSecondaryIndex(config: IndexDirectiveConfiguration, ctx: TransformerContextProvider): void {
   const { name, object, primaryKeyField } = config;
+  const dbType = getDBType(ctx, object.name.value);
+  if (dbType === 'MySQL') {
+    return;
```
phani-srikar, line 344: can we move this whole method to DDB VTL generator? Since it isn't doing anything for RDS data sources.

packages/amplify-graphql-index-transformer/src/resolvers/resolvers.ts
```diff
@@ -418,25 +427,28 @@ export function updateResolversForIndex(
   const deleteResolver = getResolverObject(config, ctx, 'delete');
   const syncResolver = getResolverObject(config, ctx, 'sync');
 
+  const dbType = getDBType(ctx, object.name.value);
+  const isDynamoDB = dbType === 'DDB';
+
   // Ensure any composite sort key values and validate update operations to
   // protect the integrity of composite sort keys.
-  if (createResolver) {
+  if (isDynamoDB && createResolver) {
```
phani-srikar, line 435: Is this method common to both DDB and RDS? If not, I would prefer we move it to DDB VTL generator as well. That would avoid such conditional checks.

phani-srikar (changes requested):

packages/amplify-graphql-index-transformer/src/resolvers/resolvers.ts
sundersc: This is not logically part of VTL generation. This must be a resource provider as it adds CFN resources but I don't see the value of doing to major refactor just to avoid a IF block.

packages/amplify-graphql-index-transformer/src/resolvers/resolvers.ts
sundersc: Yes, it is common for RDS and DDB.

phani-srikar (approved):

PR #1230, opened 2023-02-06
Enable rds lambda

marcvberg:
#### Description of changes
Enables the RDS lambda, connecting to database successfully and making requests

Adds resource generator abstraction to model transformer
VTL generation is also brought behind the resource generators as the resolvers are part of the resources

This is a pretty significant refactor of the model transformer, but functionally should not impact anything. This is done to make it easier to add RDS support

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes


packages/amplify-graphql-model-transformer/package.json
```diff
@@ -21,7 +21,7 @@
     "access": "public"
   },
   "scripts": {
-    "build": "tsc && cd rds-lambda && tsc && bestzip --force node ../lib/streaming-lambda.zip ./lib*",
+    "build": "tsc && cd rds-lambda && npm install && tsc && cd lib && mkdir -p ./node_modules && cp -R ../node_modules/* ./node_modules/ && bestzip --force node ../../lib/rds-lambda.zip ./*",
```
sundersc, line 24: Should we delete the node_modules folder before running `npm install`?

packages/amplify-graphql-model-transformer/src/types.ts
```diff
@@ -0,0 +1,9 @@
+import { SyncConfig } from '@aws-amplify/graphql-transformer-core';
+
+export type ModelTransformerOptions = {
+  EnableDeletionProtection?: boolean;
+  SyncConfig?: SyncConfig;
+};
+
+export const MYSQL_DB_TYPE = 'MySQL';
+export const DDB_DB_TYPE = 'DDB';
```
phani-srikar, line 9: Would love to see these moved to the GQL core as well.

packages/amplify-graphql-model-transformer/rds-lambda/clients/mysql/MySQLClient.ts
```diff
@@ -0,0 +1,95 @@
...
+      const orderByConditions = sortKeys.map((sortKey) => {
+        return {
+          column: sortKey,
+          order: sortDirection.toString().toLowerCase(),
+        };
+      });
+      query.orderBy(orderByConditions);
+    }
+  }
+}
```
sundersc, line 95: Do these differ between DB engines? If not, can we move these to base class or generic `util` file?

packages/amplify-graphql-model-transformer/rds-lambda/clients/mysql/MySQLPasswordClient.ts
```diff
@@ -0,0 +1,57 @@
...
+    });
+  }
+
+  private getSSMClient(): SSMClient {
+    return new SSMClient({});
+  }
+
+  private async getSSMValue(client: SSMClient, key: string | undefined): Promise<string> {
+    if (!key) {
+      return '';
```
phani-srikar, line 43: Would prefer to throw here rather than defaulting to prevent unintended DB connections with empty username/passwords.

packages/amplify-graphql-model-transformer/rds-lambda/clients/mysql/MySQLPasswordClient.ts
```diff
@@ -0,0 +1,57 @@
...
+      Name: key,
+      WithDecryption: true,
+    });
+    const data = await client.send(parameterCommand);
+    if ((data.$metadata?.httpStatusCode && data?.$metadata?.httpStatusCode >= 400) || !data.Parameter?.Value) {
+      throw new Error('Unable to get secret for database connection');
+    }
+    return data.Parameter?.Value ?? '';
+  }
+}
```
sundersc: Same comment as before. Can we retrieve these values outside this class and inject the values through constructor? Database specific implementation shouldn't know about how it gets the secret values.

packages/amplify-graphql-model-transformer/rds-lambda/clients/mysql/MySQLPasswordClient.ts
```diff
@@ -0,0 +1,57 @@
+import Knex from 'knex';
+import { SSMClient, GetParameterCommand } from '@aws-sdk/client-ssm';
+import { MySQLClient } from './MySQLClient';
+
+/**
+ * Postgres client with password connection implementation
+ */
+export class MySQLPasswordClient extends MySQLClient {
+  private clientPromise: Promise<any>;
```
sundersc, line 9: Why do we need `clientPromise` and `client`?

packages/amplify-graphql-model-transformer/package.json
marcvberg: we could, I wasn't as worried about it since the CI runs from a fresh environment but I can add that in quick

packages/amplify-graphql-model-transformer/src/resources/rds-model-resource-generator.ts
```diff
@@ -0,0 +1,69 @@
...
+        lambdaStack,
+        context.api,
+        role,
+        {
+          username: secretEntry?.username ?? '',
+          password: secretEntry?.password ?? '',
+          host: secretEntry?.host ?? '',
+          port: secretEntry?.port ?? '',
+          database: secretEntry?.database ?? '',
+        },
```
phani-srikar, line 47: can we directly pass the `secretEntry as RDSConnectionSecrets` here instead of creating another object?

packages/amplify-graphql-model-transformer/rds-lambda/clients/postgres/PostgresClient.ts
```diff
@@ -1,93 +1,93 @@
 import { Knex } from 'knex';
```
sundersc, line 1: As we are not going to test postgres with this release, I recommend not to include these files now. Atleast we should add `TODO`s with proper comments for future work.

phani-srikar (approved):
Non-blocking comments

phani-srikar:
Please ensure CI jobs pass. Thanks.

packages/amplify-graphql-model-transformer/rds-lambda/package.json
```diff
@@ -6,15 +6,17 @@
     "typescript": "^4.8.4"
   },
   "dependencies": {
+    "@aws-sdk/client-ssm": "^3.264.0",
     "babel-jest": "^29.1.2",
     "jest": "^29.1.2",
     "knex": "^2.3.0",
+    "mysql": "^2.18.1",
```
sundersc, line 13: We used `mysql2` in our POC, it has few advantages over the other version.
https://www.npmjs.com/package/mysql2

packages/amplify-graphql-model-transformer/rds-lambda/tsconfig.json
```diff
@@ -7,7 +7,8 @@
     "target": "es6",
     "outDir": "lib",
     "moduleResolution": "node",
-    "lib": ["es2015"],
+    "module": "CommonJS",
```
sundersc, line 10: Can we use `module`?

packages/amplify-graphql-model-transformer/src/graphql-model-transformer.ts
```diff
@@ -141,44 +121,38 @@ export class ModelTransformer extends TransformerModelBase implements Transforme
...
-      allowedValues: ['true', 'false'],
-    });
+    const datasourceMapValues: Array<DatasourceType> = Array.from(ctx.modelToDatasourceMap.values());
+    if (datasourceMapValues.some((value) => value.dbType === DDB_DB_TYPE && value.provisionDB)) {
+      this.resourceGeneratorMap.get(DDB_DB_TYPE)?.enableGenerator();
+      this.resourceGeneratorMap.get(DDB_DB_TYPE)?.enableProvisioned();
+    }
+    if (datasourceMapValues.some((value) => value.dbType === MYSQL_DB_TYPE && !value.provisionDB)) {
+      this.resourceGeneratorMap.get(MYSQL_DB_TYPE)?.enableGenerator();
+      this.resourceGeneratorMap.get(MYSQL_DB_TYPE)?.enableUnprovisioned();
```
sundersc, line 148: Why do we need these steps? I assume we have to enable these all times.

packages/amplify-graphql-model-transformer/rds-lambda/clients/mysql/MySQLPasswordClient.ts
marcvberg: Fixed locally and about to commit

packages/amplify-graphql-model-transformer/src/types.ts
marcvberg: Moved locally and about to commit

packages/amplify-graphql-model-transformer/src/resolvers/generators/rds-vtl-generator.ts
```diff
@@ -2,50 +2,59 @@ import {
...
-    return generateLambdaRequestTemplate(config.modelName, config.operation, config.operationName);
+    return generateLambdaCreateRequestTemplate(config.modelName, config.operationName);
   }
   generateCreateInitSlotTemplate(config: ModelCreateInitSlotConfig): string {
-    return generateDefaultLambdaResponseMappingTemplate(false);
+    return generateCreateInitSlotTemplate(config.modelConfig);
   }
   generateDeleteRequestTemplate(config: ModelUpdateRequestConfig): string {
-    return generateLambdaRequestTemplate(config.modelName, config.operation, config.operationName);
+    return generateLambdaDeleteRequestTemplate(config.modelName, config.operationName, config.modelIndexFields ?? ['id']);
```
sundersc, line 36: What's the purpose of `modelIndexFields`? Is it meant to retrieve primary key(s)? If so, primaryKey transformer already does this.

packages/amplify-graphql-model-transformer/src/resolvers/rds/mutation.ts
```diff
@@ -0,0 +1,152 @@
...
+    compoundExpression([
+      set(ref('lambdaInput'), obj({})),
+      set(ref('lambdaInput.table'), str(tableName)),
+      set(ref('lambdaInput.args'), obj({})),
+      set(ref('lambdaInput.operation'), str('CREATE')),
+      set(ref('lambdaInput.operationName'), str(operationName)),
+      comment('Set the default values to put request'),
+      set(ref('lambdaInput.args.input'), methodCall(ref('util.defaultIfNull'), ref('ctx.stash.defaultValues'), obj({}))),
+      comment('copy the values from input'),
+      qref(methodCall(ref('lambdaInput.args.input.putAll'), methodCall(ref('util.defaultIfNull'), ref('context.arguments.input'), obj({})))),
```
sundersc, line 64: Did we check how values get passed in case of using variables?

sundersc: Possibly we can pass the full context to the lambda and extract the values from there.

packages/amplify-graphql-model-transformer/rds-lambda/clients/mysql/MySQLClient.ts
```diff
@@ -0,0 +1,95 @@
+import { Knex } from 'knex';
```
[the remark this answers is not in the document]
marcvberg: Mostly an artifact of my original PoC work, but ultimately the reasoning is that we could create a MySQLIAMClient that allows us to connect to DB through IAM

packages/amplify-graphql-model-transformer/rds-lambda/tsconfig.json
marcvberg: because aws lambda has some limitations these were the only configs I was able to use to get it to compile and run correctly, it is of type module in the package.json

packages/amplify-graphql-model-transformer/src/graphql-model-transformer.ts
marcvberg: I did this just to prevent unnecessary resource creation. If for example, there are no RDS models, this allows us to not bother creating the lambda or anything. And the resources we create for provisioned will be different than unprovisioned

packages/amplify-graphql-model-transformer/src/resolvers/generators/rds-vtl-generator.ts
marcvberg: I'll work on this, I wrote this because I didn't have primary key support for RDS in yet

packages/amplify-graphql-model-transformer/src/resolvers/rds/mutation.ts
marcvberg: I may leave this as a followup optimization just because of the size of scope

packages/amplify-graphql-model-transformer/rds-lambda/clients/postgres/PostgresClient.ts
marcvberg: removed

packages/amplify-graphql-model-transformer/rds-lambda/package.json
marcvberg: done

packages/amplify-category-api/src/graphql-transformer/transform-graphql-schema-v2.ts
```diff
@@ -23,17 +19,17 @@ import {
...
-import { showSandboxModePrompts } from './sandbox-mode-helpers';
 import { parseUserDefinedSlots } from './user-defined-slots';
 import {
   mergeUserConfigWithTransformOutput, writeDeploymentToDisk,
 } from './utils';
 import { generateTransformerOptions } from './transformer-options-v2';
 import { TransformerFactoryArgs, TransformerProjectOptions } from './transformer-options-types';
-import { ProjectOptions } from './transform-config';
+import {contextUtil} from '../category-utils/context-util';
+import {getExistingConnectionSecretNames} from '../provider-utils/awscloudformation/utils/rds-secrets/database-secrets';
```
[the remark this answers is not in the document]
marcvberg: Fixed

packages/amplify-graphql-model-transformer/rds-lambda/clients/mysql/MySQLClient.ts
marcvberg: Pulled up

packages/amplify-graphql-model-transformer/rds-lambda/clients/mysql/MySQLPasswordClient.ts
marcvberg: I'll leave this as an optimization for now, since it's a singleton pattern and we don't want the top level function calling SSM each time. We can refactor this for preview 2, I have a running list

sundersc (approved):

PR #1237, opened 2023-02-08
fix(graphql): rds import errors due to primary key and index

sundersc:
- Fixes generate schema issue if a table doesn't contain a primary key.
- Fixes an issue with index directive assigning incorrect datasource for the resolver.
- Fixes use of mysql2 package in RDS lambda.
- Disable populating timestamps for RDS datasource. (We need to populate these fields only if RDS table contains it, it will be addressed in a follow up PR).
- Fixes CREATE, UPDATE, DELETE and GET lambda behavior.

- Manual test

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-graphql-model-transformer/rds-lambda/package.json
```diff
@@ -10,7 +10,7 @@
     "babel-jest": "^29.1.2",
     "jest": "^29.1.2",
     "knex": "^2.3.0",
-    "mysql2": "^3.1.0",
+    "mysql2": "2.3.3",
```
marcvberg, line 13: If the new version doesn't support that's alright, any particular reason we can't use the latest mysql2?

marcvberg (approved):
LGTM, question is not a blocker

packages/amplify-graphql-model-transformer/rds-lambda/package.json
sundersc: There are some contract changes, I ran into some issues with types with 3.x. Also 2.x is stable and well-tested during POC.

packages/amplify-graphql-index-transformer/src/resolvers/resolvers.ts
```diff
@@ -457,25 +457,35 @@ export function updateResolversForIndex(
...
   if (!(name && queryField)) {
     throw new Error('Expected name and queryField to be defined while generating resolver.');
   }
   const modelName = object.name.value;
   const dbInfo = getDBInfo(ctx, modelName);
-  const dataSourceName = `${object.name.value}Table`;
+  let dataSourceName = `${object.name.value}Table`;
+  if (dbType === MYSQL_DB_TYPE) {
+    dataSourceName = RDSLambdaDataSourceLogicalID;
+  }
```
alharris-at, line 480: As a general pattern I don't love setting a default and then mutating it on conditions. I'd prefer this get extracted to a helper so we can avoid the `let`.

packages/amplify-graphql-index-transformer/src/resolvers/resolvers.ts
```diff
@@ -457,25 +457,35 @@ export function updateResolversForIndex(
...
   const dataSource = ctx.api.host.getDataSource(dataSourceName);
   const queryTypeName = ctx.output.getQueryTypeName() as string;
-  const table = getTable(ctx, object);
+
+  let stackId = object.name.value;
+  if (isDynamoDB) {
+    const table = getTable(ctx, object);
+    stackId = table.stack.node.id;
+  }
```
alharris-at, line 488: Similar comment here.

alharris-at (approved):
Style comment, but not blocking.

packages/amplify-graphql-index-transformer/src/resolvers/resolvers.ts
phani-srikar: +1 here

phani-srikar (approved):

PR #1241, opened 2023-02-09
fix(graphql): index rds query

sundersc:
- Fixes the RDS Index query request mapping template.
- Fixes the RDS lambda to process index query.

### DIFF
--- a/packages/amplify-graphql-model-transformer/rds-lambda/clients/mysql/MySQLClient.ts
+++ b/packages/amplify-graphql-model-transformer/rds-lambda/clients/mysql/MySQLClient.ts
@@ -1,7 +1,7 @@
 import { Knex } from 'knex';
 import { DBClient } from '../DBClient';
 import { BaseRequest, Request } from '../../interfaces/BaseRequest';
-import { ListRequest } from '../../interfaces/ListRequest';
+import { IndexRequest, ListRequest } from '../../interfaces/ListRequest';
 
 export abstract class MySQLClient extends DBClient {
   client: Knex;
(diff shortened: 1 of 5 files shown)


marcvberg (approved):
LGTM

packages/amplify-graphql-model-transformer/rds-lambda/interfaces/BaseRequest.ts
```diff
@@ -9,14 +9,14 @@ export interface RequestArgs extends BaseRequestArgs {
 
 export interface BaseRequest {
   table: string;
-  operation: 'GET' | 'LIST' | 'CREATE' | 'UPDATE' | 'DELETE';
+  operation: 'GET' | 'LIST' | 'CREATE' | 'UPDATE' | 'DELETE' | 'INDEX';
```
phani-srikar, line 12: `INDEX` doesn't feel like it's an operation. It's more like part of DDL.

packages/amplify-graphql-model-transformer/rds-lambda/clients/mysql/MySQLClient.ts
```diff
@@ -89,4 +91,22 @@ export abstract class MySQLClient extends DBClient {
 
     return result ? result[0] : {};
   }
+
+  private executeIndex = async (request: IndexRequest): Promise<any> => {
```
phani-srikar, line 95: "Execute Query with Index" is more readable here.

phani-srikar (approved):
Non-blocking comments.

alharris-at (approved):

packages/amplify-graphql-model-transformer/rds-lambda/interfaces/BaseRequest.ts
sundersc: I'm going to cut a tagged release after this. I will clean up these method names in the next PR.

PR #1249, opened 2023-02-13
chore: adding multiple lambda functions with authorizer will have access to Graphql API

manaswi223:
#### Description of changes
This PR adds additional checks by adding unit & integration tests to the custom authorization rule of the Auth Transformer for a scenario where adding multiple Lambda functions to grant access to the GraphQL API should succeed with an `amplify add/update function` along with an `amplify push --force`.

The following amplify workflow helps understand the use case:

1. `amplify init` in a project root directory
2. `amplify add api` and choose Lambda as **default** authorizer

For a given schema:
```
type Todo @model @auth(rules: [{ allow: custom, provider: function }]) {
  id: ID!
  content: String!
}
```
3. `amplify add function` named `envFunction1` and allow this function to access the GraphQL API
4. `amplify add function` named `envFunction2` and allow this function also to access the GraphQL API
4. `amplify push --force`

Now, checking the working of the two Lambda functions (`envFunction1`, `envFunction2`)on the console and successfully returning the appropriate queries/mutations/subscriptions results.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-graphql-auth-transformer/src/__tests__/custom-auth.test.ts
+++ b/packages/amplify-graphql-auth-transformer/src/__tests__/custom-auth.test.ts
@@ -148,6 +148,87 @@ test('allow: custom and provider: iam error out for invalid combination', () => 
   );
 });
 
+// test if custom auth mode allows two different lambda functions to access the graphql api
+test('allow: custom with provider: two lambda functions authorized to access the graphql api to be pushed in the cloud', () => {
+  const authConfig: AppSyncAuthConfiguration = {
+    defaultAuthentication: {
+      authenticationType: 'AWS_LAMBDA',
+      lambdaAuthorizerConfig: {
+        lambdaFunction: 'test',
+        ttlSeconds: 600,
+      },
+    },
+    additionalAuthenticationProviders: [
+      {
+        authenticationType: 'AWS_LAMBDA',
+        lambdaAuthorizerConfig: {
+          lambdaFunction: 'test2',
+          ttlSeconds: 600,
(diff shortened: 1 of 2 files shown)


packages/amplify-graphql-auth-transformer/src/__tests__/custom-auth.test.ts
```diff
@@ -148,6 +148,87 @@ test('allow: custom and provider: iam error out for invalid combination', () =>
...
+  const out = transformer.transform(validSchema);
+  expect(out).toBeDefined();
+  expect(out.rootStack!.Resources![ResourceConstants.RESOURCES.GraphQLAPILogicalID].Properties.AuthenticationType).toEqual('AWS_LAMBDA');
+  expect(
+    out.rootStack!.Resources![ResourceConstants.RESOURCES.GraphQLAPILogicalID].Properties.AdditionalAuthenticationProviders[0]
+      .AuthenticationType,
+  ).toEqual('AWS_LAMBDA');
+});
+
+test('allow: custom without provider: two lambda functions authorized to access the graphql api to be pushed in the cloud', () => {
```
AaronZyLee, line 192: It looks like the provider will be default for lambda function if not provided and it is already included in previous unit test cases. If there are no other purposes, either of the newly added case can be removed

AaronZyLee (changes requested):

sundersc (changes requested):
The purpose of the PR is unclear. Looks existing tests already do the same checks.

PR #1253, opened 2023-02-14
chore: Automate creation of version bump PR in CLI repo

phani-srikar:
#### Description of changes

Automate the creation of version bump PR in Amplify CLI repo for the API plugin and related dependencies.
This runs the `build` and `test` jobs after doing the version bumps and before pushing the changes to new feature branch.
This also triggers the `cloud-e2e` job after creating the Pull Request.

Tested on this branch until before the part where PR is created. Currently there is no active version bump, so in the next release, that flow will be executed. 

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


.circleci/config.base.yml
```diff
@@ -358,6 +358,20 @@ jobs:
...
+      - attach_workspace:
+          at: ./
+      - restore_cache:
+          keys:
+            - amplify-category-api-ssh-deps-{{ .Branch }}
+      - install_github_cli
+      - run:
+          name: Run Post Publish Tasks
+          command: |
+            chmod +x .circleci/post_publish.sh && ./.circleci/post_publish.sh
```
dpilch, line 373: Why not `chmod +x` and commit change to git so that this doesn't need to be done on circleci?

.circleci/post_publish.sh
```diff
@@ -0,0 +1,69 @@
...
+git commit -m "'$COMMIT_MESSAGE'"
+
+# Sanity check the version bump changes before pushing
+yarn build && yarn test
+
+# Push to upstream
+git push --set-upstream origin $FEATURE_BRANCH
+
+# Create a PR to dev branch of CLI
+gh pr create --base $BASE_BRANCH --head $FEATURE_BRANCH --title "'$COMMIT_MESSAGE'" --body "'$PR_BODY'"
```
dpilch, line 66: What if the pr already exists?

.circleci/post_publish.sh
```diff
@@ -0,0 +1,69 @@
...
+  git branch -D $FEATURE_BRANCH # Delete if already exists locally
+fi
+git checkout -b $FEATURE_BRANCH
+
+# Pull latest changes from remote if any
+git pull
+
+# Bump the versions locally
+sudo npm install -g npm-check-updates --force
+yarn update-data-packages
```
dpilch, line 49: Won't this also update the codegen packages? Is that a concern?

.circleci/config.base.yml
phani-srikar: That should work too, let me do that.

.circleci/post_publish.sh
phani-srikar: This will error out like below:
```
$ gh pr create
a pull request for branch "<head>" into branch "<base>" already exists:
https://github.com/OWNER/REPO/pull/NUMBER
```

.circleci/post_publish.sh
phani-srikar: It will bump both API and Codegen plugin related dependencies. It's not a concern as today we typically bump both of them together. There are few exceptions but those are mainly for hotfixes in which case the target branch is not `dev` anymore. So those need manual intervention from on-call.

AaronZyLee (approved):

PR #1258, opened 2023-02-15
chore: main to cdkv2 merge

sobolk:
#### Description of changes

Merges main to cdkv2 branch.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-category-api/package.json
+++ b/packages/amplify-category-api/package.json
@@ -59,7 +59,7 @@
     "graphql-http-transformer": "5.2.48-beta.0",
     "graphql-key-transformer": "3.2.48-beta.0",
     "graphql-predictions-transformer": "3.2.48-beta.0",
-    "graphql-relational-schema-transformer": "2.21.15-beta.0",
+    "graphql-relational-schema-transformer": "2.21.15",
     "graphql-transformer-common": "4.24.5-beta.0",
     "graphql-transformer-core": "8.0.0-beta.4",
     "graphql-versioned-transformer": "5.2.48-beta.0",
--- a/packages/amplify-e2e-core/package.json
+++ b/packages/amplify-e2e-core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "amplify-category-api-e2e-core",
-  "version": "4.0.16-beta.0",
+  "version": "4.0.16",
   "description": "Core testing library",
   "repository": {
     "type": "git",
--- a/packages/amplify-e2e-tests/package.json
+++ b/packages/amplify-e2e-tests/package.json
@@ -26,7 +26,7 @@
   "dependencies": {
     "@aws-amplify/graphql-transformer-core": "1.1.0-beta.4",
     "amplify-app": "4.3.10-beta.0",
-    "amplify-category-api-e2e-core": "4.0.16-beta.0",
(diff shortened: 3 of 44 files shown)


packages/amplify-category-api/package.json
```diff
@@ -59,7 +59,7 @@
     "graphql-http-transformer": "5.2.48-beta.0",
     "graphql-key-transformer": "3.2.48-beta.0",
     "graphql-predictions-transformer": "3.2.48-beta.0",
-    "graphql-relational-schema-transformer": "2.21.15-beta.0",
+    "graphql-relational-schema-transformer": "2.21.15",
```
akshbhu, line 62:
```suggestion
    "graphql-relational-schema-transformer": "2.21.15-beta.0",
```

packages/amplify-e2e-core/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "amplify-category-api-e2e-core",
-  "version": "4.0.16-beta.0",
+  "version": "4.0.16",
```
akshbhu, line 3:
```suggestion
  "version": "4.0.16-beta.0",
```

packages/amplify-e2e-tests/package.json
```diff
@@ -26,7 +26,7 @@
   "dependencies": {
     "@aws-amplify/graphql-transformer-core": "1.1.0-beta.4",
     "amplify-app": "4.3.10-beta.0",
-    "amplify-category-api-e2e-core": "4.0.16-beta.0",
+    "amplify-category-api-e2e-core": "4.0.16",
```
akshbhu, line 29: same here

packages/amplify-category-api/package.json
sobolk: That was deliberate. If there's no major version change I prefer betas only when they get ahead.

Couple of examples.
- main is `2.21.16` and cdkv2 is `2.21.15-beta.0` - we should pick `2.21.16` to move forward, beta release will be `2.21.17-beta.0`
- main is `2.21.15` and cdkv2 is `2.21.15-beta.0` - we should pick `2.21.15` to move forward, beta release will be `2.21.16-beta.0` (like in case of this PR)
- main is `2.21.16` and cdkv2 is `2.21.17-beta.0` - we should pick `2.21.17-beta.0` to move forward, beta release will be `2.21.17-beta.1` (i.e. beta is already ahead so we should keep it so that lerna increments that last number)

akshbhu: I agree in 1st and 3rd . In second case , since main is 2.21.15 , would it be preferred to choose 2.21.15-beta.0  as there is no change in module, this will simply beta  release 2.21.15-beta.1

sobolk: The problem is that when we decide to GA then lerna will calculate that GA version for `2.21.15-beta.1` is `2.21.15` and that will clash with already existing `2.21.15` in npm.
If we resolve to `2.21.15` and ship `2.21.16-beta.0` then GA is automatic to `2.21.16`.
Please keep in mind that there are dependency changes on branch, that's why it keeps releasing those packages.

alharris-at (approved):

akshbhu (approved):

PR #1264, opened 2023-02-15
feat: added schema validators

bzsurbhi:
#### Description of changes
This PR adds 2 schema validators
1) Validates a field is only defined once in a model
2) Validates that any of the fields passed in are in the parent model

None

#### Issue #, if available
N/A

Unit tests added

#### Checklist

- [x ] PR description included
- [x ] `yarn test` passes
- [x ] Tests are [changed or added](https://github.com/aws-amplify/amplify-cli/blob/master/CONTRIBUTING.md#tests)


johnpc (changes requested):
Looks great! One quick change required: `yarn run test` is failing because of insufficient line coverage. Specifically, line 23 of field-is-defined-once.ts and line 32 of fields-passed-are-in-parent-model.ts are uncovered.

packages/amplify-schema-validator/src/validators/field-is-defined-once.ts
```diff
@@ -0,0 +1,40 @@
...
+  export const validateFieldIsDefinedOnce = (schema: DocumentNode): Error[] => {
+    const errors: Error[] = [];
+    const objectTypeDefinitions = schema.definitions.filter(
+      (defintion) => defintion.kind === Kind.OBJECT_TYPE_DEFINITION,
+    ) as ObjectTypeDefinitionNode[];
+    objectTypeDefinitions.forEach((objectTypeDefinition) => {
+      const objectName = objectTypeDefinition.name.value;
+      const fields = objectTypeDefinition.fields;
+      if(!fields) {
+        return;
```
johnpc, line 23: this line is not covered

packages/amplify-schema-validator/src/validators/fields-passed-are-in-parent-model.ts
```diff
@@ -0,0 +1,64 @@
...
+    ) as ObjectTypeDefinitionNode[];
+    objectTypeDefinitions.forEach((objectTypeDefinition) => {
+      const directiveFields = objectTypeDefinition.fields?.filter((objectField) => objectField.directives?.find(
+        (directive) => directive.name.value,
+      ));
+
+      directiveFields?.forEach((directiveField) => {
+        const typeName = resolveFieldTypeName(directiveField.type);
+        if (!typeName) {
+          return;
```
johnpc, line 32: this line is not covered

johnpc: if there is not value in covering this case in unit tests, you can add `/* istanbul ignore next */` to ignore the uncovered line.

packages/amplify-schema-validator/src/validators/field-is-defined-once.ts
bzsurbhi: Fixed

packages/amplify-schema-validator/src/validators/fields-passed-are-in-parent-model.ts
bzsurbhi: Fixed

packages/amplify-schema-validator/src/validators/field-is-defined-once.ts
```diff
@@ -0,0 +1,40 @@
...
+  Kind,
+  ObjectTypeDefinitionNode,
+} from 'graphql';
+import { InvalidDirectiveError } from '../exceptions/invalid-directive-error';
+
+/**
+   * Field is defined once in a model
+   *
+   * @param schema graphql schema
+   * @returns true
```
alharris-at, line 12: Returns true if... - would you mind finishing this doc string up?

packages/amplify-schema-validator/src/validators/fields-passed-are-in-parent-model.ts
```diff
@@ -0,0 +1,64 @@
...
+   * @param schema graphql schema
+   * @returns true
+   */
+export const validateFieldInParentModel = (schema: DocumentNode): Error[] => {
+  const errors: Error[] = [];
+  const objectTypeDefinitions = schema.definitions.filter(
+    (defintion) => defintion.kind === Kind.OBJECT_TYPE_DEFINITION,
+  ) as ObjectTypeDefinitionNode[];
+  objectTypeDefinitions.forEach((objectTypeDefinition) => {
+    const directiveFields = objectTypeDefinition.fields?.filter((objectField) => objectField.directives?.find(
```
sundersc, line 24: Are you you trying to get the fields with directives defined? If so, can this be simplified something like this?
```suggestion
    const directiveFields = objectTypeDefinition.fields?.filter((objectField) => objectField.directives?.length > 0)
```

packages/amplify-schema-validator/src/validators/fields-passed-are-in-parent-model.ts
```diff
@@ -0,0 +1,64 @@
...
+      (directive) => directive.name.value,
+    ));
+
+    directiveFields?.forEach((directiveField) => {
+      const typeName = resolveFieldTypeName(directiveField.type);
+      if (!typeName) {
+        /* istanbul ignore next */
+        return;
+      }
```
sundersc, line 33: This may be unreachable code. Since we are working with a parsed document node, typeName should always have a value.

Please check whether this is required.

packages/amplify-schema-validator/src/validators/fields-passed-are-in-parent-model.ts
```diff
@@ -0,0 +1,64 @@
...
+      const objectOfType = getObjectWithName(schema, typeName);
+      const fields = objectOfType?.fields;
+      if (!fields) {
+        /* istanbul ignore next */
+        return;
+      }
+      const fieldVals: string[] = [];
```
sundersc, line 40: This can be simplified with array.map function.

packages/amplify-schema-validator/src/validators/fields-passed-are-in-parent-model.ts
```diff
@@ -0,0 +1,64 @@
...
+      for (let i = 0; i < fields?.length; i++) {
+        const val = fields[i].name.value;
+        fieldVals.push(val);
+      }
```
sundersc, line 44: This can be simplified with array.map function.

packages/amplify-schema-validator/src/validators/fields-passed-are-in-parent-model.ts
```diff
@@ -0,0 +1,64 @@
...
+
+      const directiveArgs = directiveField.directives?.find((directive) => directive.arguments);
+      const fieldArg = directiveArgs?.arguments?.find((arg) => arg.name.value === 'fields');
+      if (!fieldArg) {
+        return;
+      }
```
sundersc, line 50: Are you looking for specific directives with `fields` attribute? If so, I would recommend to add those directive names explicitly in the filter above.

packages/amplify-schema-validator/src/validators/fields-passed-are-in-parent-model.ts
```diff
@@ -0,0 +1,64 @@
...
+import { InvalidDirectiveError } from '../exceptions/invalid-directive-error';
+import { getObjectWithName } from '../helpers/get-object-with-name';
+import { resolveFieldTypeName } from '../helpers/resolve-field-type-name';
+
+/**
+   * Validates that any of the fields passed in are in the parent model
+   *
+   * @param schema graphql schema
+   * @returns true
+   */
```
alharris-at, line 17: Could you also flesh out the documentation here a bit more too?

In general I wouldn't use the term parent model, Could we refer to it here as the specific relationship type, or if it applies to all relationship types, maybe just call it `validateFieldsMatchInRelatedModel`

packages/amplify-schema-validator/src/validators/fields-passed-are-in-parent-model.ts
sundersc: Also, find returns the first element only. What if there are more than one directives on the field.

alharris-at (commented):
Overall LGTM - a couple q's on docs and naming.

packages/amplify-schema-validator/src/validators/fields-passed-are-in-parent-model.ts
```diff
@@ -0,0 +1,64 @@
...
+      const fieldArgVals = (fieldArg.value as ListValueNode).values;
+      for (let i = 0; i < fieldArgVals.length; i++) {
+        const val = (fieldArgVals[i] as StringValueNode).value;
+        if (!fieldVals.includes(val)) {
+          errors.push(new InvalidDirectiveError(
+            `${val} is not a field in ${objectOfType.name.value}`,
+          ));
+        }
+      }
+    });
```
sundersc, line 61: Please change this to array.foreach to match code style.

packages/amplify-schema-validator/src/validators/fields-passed-are-in-parent-model.ts
sundersc: Also, for `hasMany` we do not want to do this validation. Currently transformers ignore the fields attribute and use the primary key to perform the join.

packages/amplify-schema-validator/src/validators/fields-passed-are-in-parent-model.ts
bzsurbhi: Fixed

packages/amplify-schema-validator/src/validators/field-is-defined-once.ts
bzsurbhi: Fixed

packages/amplify-schema-validator/src/validators/fields-passed-are-in-parent-model.ts
bzsurbhi: Fixed

packages/amplify-schema-validator/src/validators/fields-passed-are-in-parent-model.ts
bzsurbhi: Fixed

packages/amplify-schema-validator/src/validators/fields-passed-are-in-parent-model.ts
bzsurbhi: Refactored

alharris-at (approved):

packages/amplify-schema-validator/src/validators/fields-passed-are-in-parent-model.ts
bzsurbhi: I added the specific directive names in the filter. Did not include hasMany in the directive filter

packages/amplify-schema-validator/src/validators/fields-passed-are-in-parent-model.ts
bzsurbhi: Added specific directive names to filter

sundersc (approved):
LGTM

packages/amplify-schema-validator/src/helpers/is-scalar-or-enum.ts
```diff
@@ -49,7 +49,7 @@ export const isScalarOrEnum = (
     return isScalarOrEnum(type.type, enums);
   } if (type.kind === Kind.LIST_TYPE) {
     /* istanbul ignore next */
-    return isScalarOrEnum(type.type, enums);
+    return false;
```
bzsurbhi, line 52: Index should be non scalar validation is failing for list type. Return false directly for list type fields

sundersc: LGTM.

johnpc (approved):

PR #1273, opened 2023-02-20
add support for eu-south-1 MXP

johnpc:
#### Description of changes

This PR enables support in eu-south-1 region (MXP). Associated with https://github.com/aws-amplify/amplify-cli/pull/12042

N/A

#### Issue #, if available

N/A

I've created a tagged release of both `amplify-category-api` with https://github.com/aws-amplify/amplify-category-api/pull/1273 and `amplify-cli` including https://github.com/aws-amplify/amplify-cli/pull/12042.

The eu-south-1 CLI can be installed via 

```
npm i -g @aws-amplify/cli@10.8.0-mxp.0
```

I then set eu-south-1 in my `~/.aws/config` for my amplify profile.

First, to test amplify-cli, I ran:

```
amplify init
amplify add auth
amplify add storage
amplify push
```

accepting all default values.

Then, to test `amplify-category-api`, I ran `amplify add api && amplify push`.

It all works and the expected resources / CFN stacks appear in my AWS account in eu-south-1.

* e2e suite: https://app.circleci.com/pipelines/github/aws-amplify/amplify-category-api?branch=run-e2e/xss/main

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-e2e-core/src/utils/pinpoint.ts
+++ b/packages/amplify-e2e-core/src/utils/pinpoint.ts
@@ -38,6 +38,7 @@ const serviceRegionMap = {
   'ap-northeast-1': 'ap-northeast-1',
   'eu-central-1': 'eu-central-1',
   'eu-north-1': 'eu-central-1',
+  'eu-south-1': 'eu-central-1',
   'eu-west-1': 'eu-west-1',
   'eu-west-2': 'eu-west-2',
   'eu-west-3': 'eu-west-1',
--- a/packages/amplify-category-api/src/provider-utils/supported-datasources.ts
+++ b/packages/amplify-category-api/src/provider-utils/supported-datasources.ts
@@ -43,6 +43,7 @@ export const supportedDataSources = {
       'ca-central-1',
       'eu-central-1',
       'eu-north-1',
+      'eu-south-1',
       'eu-west-1',
       'eu-west-2',
       'eu-west-3',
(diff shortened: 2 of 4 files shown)


packages/amplify-e2e-core/src/utils/pinpoint.ts
```diff
@@ -38,6 +38,7 @@ const serviceRegionMap = {
   'ap-northeast-1': 'ap-northeast-1',
   'eu-central-1': 'eu-central-1',
   'eu-north-1': 'eu-central-1',
+  'eu-south-1': 'eu-central-1',
```
johnpc, line 41: Pinpoint is not available in eu-south-1, so we fall back to central 

https://docs.aws.amazon.com/general/latest/gr/pinpoint.html

alharris-at (approved):

phani-srikar:
One suggestion I have here is to add this new region to our e2e testing regions, so we run tests consistently across that region. [Link](https://github.com/aws-amplify/amplify-category-api/blob/3e983ff3672188cb33731297540a80d6f4b0d9a6/scripts/split-e2e-tests.ts#L67). We can do that in a different PR as well.

johnpc:
> One suggestion I have here is to add this new region to our e2e testing regions, so we run tests consistently across that region. [Link](https://github.com/aws-amplify/amplify-category-api/blob/3e983ff3672188cb33731297540a80d6f4b0d9a6/scripts/split-e2e-tests.ts#L67). We can do that in a different PR as well.

I considered that, but note that it's an opt-in region. This means that your e2e test accounts will need to enable the region in order for the tests to work. 

Additionally, the e2e tests use amplify cli, so until https://github.com/aws-amplify/amplify-cli/pull/12042 is released, the tests wouldn't work in that region even if MXP were enabled in the test accounts.

alharris-at (approved):

PR #1291, opened 2023-02-22
fix: non-capitalized embeddable type name is also respected by the GraphQL V2 Model Transformer

manaswi223:
#### Description of changes

Today, adding a non-capitalized first letter/camelCased  `Embeddable` type (non-model type) to the GraphQL schema throws error in the V2 transformer while pushing the resources to the cloud. This PR adds support to such camelCased embeddable types to be respected by the V2 transformer and push to the cloud successfully.

For example, add  `contentSupport`  which is an embeddable type in a typical GraphQL schema with a `Todo` model:
(observe that the embeddable type `contentSupport` is camelCased in contrast with the `Todo` model which has a capitalized first letter)

sample schema:
```
type Todo @model {
  id: ID!
  content: [contentSupport]
}
type contentSupport {
  name: String!
  description: String!
}
```

Now, run an amplify push to see a deployment failure like below
```
🛑 The following resources failed to deploy:
Resource Name: GraphQLAPITransformerSchema*** (AWS::AppSync::GraphQLSchema)
Event Type: update
Reason: Schema Creation Status is FAILED with details: Found 2 problem(s) with the schema:
The input value type 'ContentSupport' is not present when resolving type 'CreateTodoInput' [@382:1].
The input value type 'ContentSupport' is not present when resolving type 'UpdateTodoInput' [@390:1].
```
With this PR, the amplify push in the V2 transformer is successful as in the V1 transformer we support today.

Also resolves this issue https://github.com/aws-amplify/amplify-codegen/issues/426 when migrating from V1 to V2 transformer with such non-capitalized/camelCased Embeddable type names.

#### Issue #, if available
https://github.com/aws-amplify/amplify-codegen/issues/426

#### Checklist

- [x] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/packages/amplify-e2e-tests/schemas/custom_query.graphql
+++ b/packages/amplify-e2e-tests/schemas/custom_query.graphql
@@ -11,7 +11,7 @@ type Comment @model {
   todo: Todo @hasOne
 }
 
-type CommentConnection {
+type commentConnection {
   items: [Comment]
   nextToken: String
 }
--- a/packages/amplify-e2e-tests/schemas/custom_query.graphql
+++ b/packages/amplify-e2e-tests/schemas/custom_query.graphql
@@ -17,5 +17,5 @@ type commentConnection {
 }
 
 type Query {
-  commentsForTodo(todoId: ID!, limit: Int, nextToken: String): CommentConnection
+  commentsForTodo(todoId: ID!, limit: Int, nextToken: String): commentConnection
 }
--- a/packages/amplify-e2e-tests/schemas/simple_model.graphql
+++ b/packages/amplify-e2e-tests/schemas/simple_model.graphql
@@ -1,4 +1,9 @@
 type Todo @model {
   id: ID!
   content: String
+  content: [contentSupport]
+}
+type contentSupport {
+  name: String!
+  description: String!
 }
--- a/packages/amplify-e2e-tests/schemas/simple_model.graphql
+++ b/packages/amplify-e2e-tests/schemas/simple_model.graphql
@@ -1,7 +1,7 @@
 type Todo @model {
   id: ID!
   content: String
-  content: [contentSupport]
+  contentSupport: [contentSupport]
 }
 type contentSupport {
   name: String!
--- a/packages/amplify-e2e-tests/src/__tests__/api_3.test.ts
+++ b/packages/amplify-e2e-tests/src/__tests__/api_3.test.ts
@@ -188,5 +188,4 @@ describe('amplify add api (GraphQL)', () => {
       expect(err.message).toBe(`GraphQL API ${GraphQLAPIIdOutput} not found.`);
     }
   });
-
 });
--- a/packages/amplify-e2e-tests/src/__tests__/api_3.test.ts
+++ b/packages/amplify-e2e-tests/src/__tests__/api_3.test.ts
@@ -67,54 +67,12 @@ describe('amplify add api (GraphQL)', () => {
     expect(graphqlApi.apiId).toEqual(GraphQLAPIIdOutput);
   });
 
-  // init a project and add the non-capitalized model & Embeddable types api, change transformer version to base version and push
-  it('init a project and add the non-capitalized model & Embeddable types api, change transformer version to base version and push', async () => {
-    const name = `noncapitalizedmodelv${TRANSFORM_BASE_VERSION}`;
-    await initJSProjectWithProfile(projRoot, { name });
-    await addApiWithoutSchema(projRoot, { transformerVersion: 1 });
-    await updateApiSchema(projRoot, name, 'non-capitalized-model.graphql');
-    const transformConfig = getTransformConfig(projRoot, name);
-    expect(transformConfig).toBeDefined();
-    expect(transformConfig.Version).toBeDefined();
-    expect(transformConfig.Version).toEqual(TRANSFORM_CURRENT_VERSION);
-
-    transformConfig.Version = TRANSFORM_BASE_VERSION;
-    const apiRoot = path.join(projRoot, 'amplify', 'backend', 'api', name);
-    writeTransformerConfiguration(apiRoot, transformConfig);
-
-    await amplifyPush(projRoot);
-
-    const meta = getProjectMeta(projRoot);
-    const { output } = meta.api[name];
-    const { GraphQLAPIIdOutput, GraphQLAPIEndpointOutput, GraphQLAPIKeyOutput } = output;
-    const { graphqlApi } = await getAppSyncApi(GraphQLAPIIdOutput, meta.providers.awscloudformation.Region);
-
-    expect(GraphQLAPIIdOutput).toBeDefined();
-    expect(GraphQLAPIEndpointOutput).toBeDefined();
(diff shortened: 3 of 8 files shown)


packages/amplify-graphql-model-transformer/src/graphql-model-transformer.ts
```diff
@@ -1084,7 +1085,7 @@ export class ModelTransformer extends TransformerModelBase implements Transforme
 
   private isModelField = (name: string): boolean => (!!this.typesWithModelDirective.has(name));
 
-  private getNonModelInputObjectName = (name: string): string => `${name}Input`;
+  private getNonModelInputObjectName = (name: string): string => `${toUpper(name)}Input`;
```
sundersc, line 1088: Looking at the issue, you may need to use 'PascalCase'. Could this break existing v2 apps?

manaswi223: 👋 @sundersc To the first half of the comment: actually `toUpper(word: string)` is a method that capitalizes the first letter in a given argument which is a String. 
whereas `toPascalCase(words: string[])` does the same but appends any given input either before the given argument and/or after, like: 
`const name = toPascalCase(['Model', type.name.value, 'Connection']);` `// output:- Model${#xyz}Connection`

So, in our use case we're just capitalizing the first letter of a given Embeddable type and appending the `Input` that is  generated in the `schema.graphql` of the build folder as a `NonModelInput` input type. I think this would not break any existing apps on V2 but provide CX opportunity to not worry about the capitalization of the embeddable type. 

Both "camelCased" Embeddable type & "camelCased" Model type names are currently respected by V1 transformer whereas only "camelCased" Model type names are currently supported in V2 transformer. So, this PR adds support to the former in V1 case i.e, support to "camelCased" Embeddable type names.

sundersc (changes requested):
Please add an E2E test.

packages/amplify-graphql-model-transformer/src/graphql-model-transformer.ts
phani-srikar: In this case, we can either say `${toPascalCase(name, "Input")}` or `${toUpper(name)Input}` as both give the same output. I've tried it with below schema:
```
type multiWordstring @model {
  name: String!
  description: String!
  multiWordstringNonmodel: multiWordstringNonmodel
}

type multiWordstringNonmodel {
  name: String!
  description: String!
}
```
And the expected type name for non-model input is `MultiWordstringNonmodelInput` which is what this PR is generating. I would add a unit test for this if possible.

phani-srikar: I don't think it would break customers since the current experience is that we fail during push due to missing type.

phani-srikar (changes requested):
Can you please add an e2e test in `amplify-e2e-tests` package to verify that given a schema with such type names, we are now able to deploy successfully.

packages/amplify-e2e-tests/src/__tests__/api_3.test.ts
```diff
@@ -188,4 +230,27 @@ describe('amplify add api (GraphQL)', () => {
       expect(err.message).toBe(`GraphQL API ${GraphQLAPIIdOutput} not found.`);
     }
   });
+
+  it('creates AppSync API in headless mode with non-capitalized model', async () => {
```
sundersc, line 234: Can we add this test to one of the existing schema? Spinning 3 new projects for this test increases the resource usage.

packages/amplify-e2e-tests/schemas/custom_query.graphql
```diff
@@ -11,7 +11,7 @@ type Comment @model {
   todo: Todo @hasOne
 }
 
-type CommentConnection {
```
manaswi223, line 14: ideal test case for the fix

packages/amplify-e2e-tests/schemas/simple_model.graphql
```diff
@@ -1,4 +1,9 @@
 type Todo @model {
   id: ID!
   content: String
+  contentSupport: [contentSupport]
```
manaswi223, line 4: added an Embeddable type `contentSupport` to an existing Todo. Can remove this if suggested to do so

packages/amplify-graphql-model-transformer/src/__tests__/model-transformer.test.ts
```diff
@@ -41,6 +41,48 @@ describe('ModelTransformer: ', () => {
     parse(out.schema);
   });
 
+  // should successfully transform simple Embeddable type (non-model) schema
+  it('should successfully transform simple Embeddable type (non-model) schema', async () => {
+    const validSchema = `
```
manaswi223, line 46: added this test additionally to check a NonModel Model is successfully transformed or not

packages/amplify-graphql-model-transformer/src/__tests__/model-transformer.test.ts
```diff
@@ -41,6 +41,48 @@ describe('ModelTransformer: ', () => {
...
+      `;
+    const transformer = new GraphQLTransform({
+      transformers: [new ModelTransformer()],
+      featureFlags,
+    });
+    const out = transformer.transform(alsoValidSchema);
+    expect(out).toBeDefined();
+    
+    validateModelSchema(parse(out.schema));
+    parse(out.schema);
```
sundersc, line 83: Consider adding a snapshot for the compiled schema.

manaswi223: added

packages/amplify-e2e-tests/schemas/custom_query.graphql
```diff
@@ -11,7 +11,7 @@ type Comment @model {
...
+type commentConnection {
```
phani-srikar, line 14: Do we need to change the reference to this `type` below as well? https://github.com/aws-amplify/amplify-category-api/pull/1291/files#diff-73a9a466ec4f9056241d6a90bae4aec1445c929555121fe15f31581cb518f5c9R20

packages/amplify-graphql-model-transformer/src/__tests__/model-transformer.test.ts
```diff
@@ -41,6 +41,50 @@ describe('ModelTransformer: ', () => {
...
+    expect(out.schema).toMatchSnapshot();
```
phani-srikar, line 85: Another useful check to have which specifically covers your fix is something like:
```
expect(out.schema).toContain("input NonModelTypeInput")
```

phani-srikar (changes requested):

packages/amplify-e2e-tests/schemas/custom_query.graphql
manaswi223: now adding it for code uniformity, but looks like capitalization of first letter doesn't effect after the change :)

sundersc (approved):

phani-srikar (approved):
LGTM

AaronZyLee (approved):

PR #1343, opened 2023-03-14
chore: fix dependabot alerts

phani-srikar:
#### Description of changes

Add appropriate version resolutions and refresh the `yarn.lock` file to fix the below dependabot alerts:
- https://github.com/aws-amplify/amplify-category-api/security/dependabot/40
- https://github.com/aws-amplify/amplify-category-api/security/dependabot/39
- https://github.com/aws-amplify/amplify-category-api/security/dependabot/38
- https://github.com/aws-amplify/amplify-category-api/security/dependabot/37
- https://github.com/aws-amplify/amplify-category-api/security/dependabot/29

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/package.json
+++ b/package.json
@@ -140,6 +140,8 @@
     "parse-url": "^8.1.0",
     "moment-timezone": "^0.5.35",
     "aws-cdk-lib": "~2.53.0",
-    "**/amplify-cli-core": "^4.0.0"
+    "**/amplify-cli-core": "^4.0.0",
+    "ejs": "^3.1.7",
+    "json5": "^2.2.3"
   }
 }
--- a/yarn.lock
+++ b/yarn.lock
@@ -2,7 +2,7 @@
 # yarn lockfile v1
 
 
-"@ampproject/remapping@^2.1.0":
+"@ampproject/remapping@^2.2.0":
   version "2.2.0"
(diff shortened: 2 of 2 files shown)


alharris-at:
Does just refreshing the yarn.lock file fix this issues? I'd prefer that we keep that file up-to-date if possible, rather than increasingly adding resolutions to the package.json file.

phani-srikar:
> Does just refreshing the yarn.lock file fix this issues? I'd prefer that we keep that file up-to-date if possible, rather than increasingly adding resolutions to the package.json file.

It does fix it for minor/patch version bumps but for `ejs` and `json5`, we need resolutions to force the dependency on latest major even though transitive dependencies specify an outdated major version.

alharris-at (approved):

phani-srikar:
Some additional context here:
I have tracked down the parent libs and here's what I found:
1. `json5` 1.x comes from https://www.npmjs.com/package/eslint-plugin-import whose latest version still depends on  1.x and not 2.x, so we cannot get rid of 1.x for now until this external plugin's deps are updated.
2. `ejs` comes from https://github.com/ai/size-limit which we use in amplify-schema-validator package which currently depends on 2.x of that 3rd party package while the latest available is 8.x. Upgrading it to 8.x is breaking the build

AaronZyLee (approved):

PR #1379, opened 2023-03-31
Schema validations

bzsurbhi:
#### Description of changes
This PR adds below mentioned schema validations:

1) XX and YY cannot refer to each other via @hasOne or @hasMany when DataStore is in use. Use @belongsTo instead
2) Schema should not use directives from an older version of the GraphQL Transformer when the version used is 2.
3) Schema should not use directives from a newer version of the GraphQL Transformer when the version used is 1.

Nona

#### Issue #, if available
N/A

Added unit tests

#### Checklist

- [x] PR description include
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-schema-validator/src/index.ts
+++ b/packages/amplify-schema-validator/src/index.ts
@@ -23,6 +23,7 @@ import { validateKeyExistsInRelatedModel } from './validators/key-exists-in-rela
 import { validateBelongsToIsUsedWhenDatastoreInUse } from './validators/use-belongsto-when-datastore-inuse';
 import { validateDirectivesFromOlderTransformerVersionAreNotUsed } from './validators/use-directives-from-older-transformer-version';
 import { validateDirectivesFromNewerTransformerVersionAreNotUsed } from './validators/use-directives-from-newer-transformer-version';
+import { ValidateSchemaProps } from './helpers/schema-validator-props';
 
 const allValidators = [
   validateIndexScalarTypes,
@@ -50,11 +51,6 @@ const allValidatorsWithContext = [
   validateDirectivesFromNewerTransformerVersionAreNotUsed,
 ];
 
-export type ValidateSchemaProps = {
-  graphqlTransformerVersion: number;
-  isDataStoreEnabled: boolean;
-};
-
 /**
  * The primary export of this library
  * runs all validators and throws a ValidationException with all failure reasons
@@ -73,9 +69,10 @@ export const validateSchema = (schemaString: string): void => {
 
 /**
  * The primary export of this library
- * runs all validators and throws a ValidationException with all failure reasons
+ * runs all validators which require context and throws a ValidationException with all failure reasons
  *
  * @param schemaString the graphql schema
+ * @param props feature flags that are required for validation
  * @returns void
  */
 export const validateSchemaWithContext = (schemaString: string, props: ValidateSchemaProps): void => {
--- a/packages/amplify-schema-validator/src/index.ts
+++ b/packages/amplify-schema-validator/src/index.ts
@@ -20,6 +20,9 @@ import { validateIndexIsDefinedOnce } from './validators/index-is-defined-once-i
 import { validateIndexExistsInRelatedModel } from './validators/index-exists-in-related-model';
 import { validateEnumIsDefinedOnce } from './validators/enum-is-defined-once';
 import { validateKeyExistsInRelatedModel } from './validators/key-exists-in-related-model';
+import { validateBelongsToIsUsedWhenDatastoreInUse } from './validators/use-belongsto-when-datastore-inuse';
+import { validateDirectivesFromOlderTransformerVersionAreNotUsed } from './validators/use-directives-from-older-transformer-version';
+import { validateDirectivesFromNewerTransformerVersionAreNotUsed } from './validators/use-directives-from-newer-transformer-version';
 
 const allValidators = [
   validateIndexScalarTypes,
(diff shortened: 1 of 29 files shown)


packages/amplify-schema-validator/src/index.ts
```diff
@@ -48,9 +54,9 @@ const allValidators = [
  * @param schemaString the graphql schema
  * @returns void
  */
-export const validateSchema = (schemaString: string): void => {
+export const validateSchema = (schemaString: string, amplifyFeatureFlags?: string, dataStoreEnabled?: boolean): void => {
```
alharris-at, line 57: Rather than generating an ever-increasing list of options for validate, I'd prefer that we use an input type, which has all the fields. Also if we need something like dataStoreEnabled, we should require it everywhere (or explicitly only pass in the required fields per transformer, but requiring everywhere will likely be simpler given you have e2e control here).

So something like

```typescript
type SharedProps = {
  graphqlTransformerVersion: number;
  isDataStoreEnabled: boolean;
};

export type ValidateSchemaProps = SharedProps & {
  schemaString: string;
};

export type ParsedSchemaValidatorProps = SharedProps & {
  schema: DocumentNode;
};

export const validateSchema = (props: ValidateSchemaProps): void => {
  const schema = parse(props.schemaString);
  const validationErrors = allValidators.flatMap((validate) => validate({ ...props, schema }));
};
```

packages/amplify-schema-validator/src/validators/use-directives-from-newer-transformer-version.ts
```diff
@@ -0,0 +1,61 @@
...
+/**
+ * Validates that @belongsTo, @connection directives from older graphql transformer version is not used in version 1.0
+ *
+ * @param schema graphql schema
+ * @returns true if correct directives are used
+ */
+
+/* eslint-disable @typescript-eslint/no-unused-vars */
+const GRAPHQL_TRANSFORMER_V2_DIRECTIVES = ['hasOne', 'index', 'primaryKey', 'belongsTo', 'manyToMany', 'hasMany', 'default'];
+export const validateDirectivesFromNewerTransformerVersionAreNotUsed = (
```
alharris-at, line 17: I suspect you could refactor the v1 directive validator and v2 directive validator into a helper function that accepts the necessary config, and returns the transformer, so we're not duplicating as much logic across the two validators.

alharris-at (commented):
Dropped specific feedback on input shapes and strongly typing the inputs and params here, but I also want to ensure we're NOT passing in the raw feature flags file to this library. If there is some system upstream which has access to feature flags it should be responsible for parsing those out, and providing more clear parameters to this library, rather than requiring this library to understand CLI-specific concerns (as this should be fully decoupled from the amplify CLI, and only have knowledge of the GraphQL Transformer system).

packages/amplify-schema-validator/src/validators/use-directives-from-newer-transformer-version.ts
```diff
@@ -0,0 +1,61 @@
...
+  schema: DocumentNode, amplifyFeatureFlags?: string, _dataStoreEnabled?: boolean,
+): Error[] => {
+  if (!amplifyFeatureFlags) {
+    return [];
+  }
+  const featureFlags = JSON.parse(amplifyFeatureFlags);
```
alharris-at, line 23: As mentioned in the overall comment, this lib shouldn't be responsible for the parse + check to determine transformer version, but the logic here is also not quite right. The actually behavior is to check `useExperimentalPipelinedTransformer && transformerVersion === 2`.

E.g. https://github.com/aws-amplify/amplify-category-api/blob/main/packages/amplify-category-api/src/graphql-transformer/transformer-version.ts#L15-L37

packages/amplify-schema-validator/src/validators/use-directives-from-newer-transformer-version.ts
```diff
@@ -0,0 +1,61 @@
...
+  const transformerVersion = featureFlags.features.graphqltransformer.transformerversion;
+  if (transformerVersion !== 1) {
+    return [];
+  }
```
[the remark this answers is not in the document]
bzsurbhi: Fixed this by just passing the required parameters from CP

packages/amplify-schema-validator/src/validators/relation-required-with-belongs-to.ts
```diff
@@ -13,7 +13,11 @@ import { resolveFieldTypeName } from '../helpers/resolve-field-type-name';
  * @param schema graphql schema
  * @returns true
  */
-export const validateRequireBelongsToRelation = (schema: DocumentNode): Error[] => {
+
+/* eslint-disable @typescript-eslint/no-unused-vars */
+export const validateRequireBelongsToRelation = (
+  schema: DocumentNode, _amplifyFeatureFlags?: string, _dataStoreEnabled?: boolean,
+): Error[] => {
```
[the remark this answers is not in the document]
bzsurbhi: Done

packages/amplify-schema-validator/src/index.ts
bzsurbhi: Refactored

packages/amplify-schema-validator/src/validators/use-directives-from-newer-transformer-version.ts
bzsurbhi: Refactored by taking out the common code

packages/amplify-schema-validator/src/validators/use-directives-from-newer-transformer-version.ts
bzsurbhi: Checking `useExperimentalPipelinedTransformer && transformerVersion === 2` behavior in CP and passing the resolved graphql version

alharris-at (approved):

Issue #78, opened 2022-03-22
API request filtered by the "_deleted" field from DataStore


sebboe:
### Is this feature request related to a new or existing Amplify category?

api

### Is this related to another service?

DataStore

### Describe the feature you'd like to request

I would like to be able to filter my API requests using the "_deleted" field generated by DataStore to avoid fetching deleted data when fetching from the API (GraphQL). The field is not present in the generated in the API FilterInput model.

Example request (not working):
```
    const flights = this.apiService.ListFlights({_deleted: {ne:true}}).then((evt) => {
      let flights = evt.items;
      return flights
    });
```
Generated FilterInput (_deleted not present):
```
export type ModelFlightFilterInput = {
  id?: ModelIDInput | null;
  and?: Array<ModelFlightFilterInput | null> | null;
  or?: Array<ModelFlightFilterInput | null> | null;
  not?: ModelFlightFilterInput | null;
};
```

### Describe the solution you'd like

Adding the _deleted field in the generated API FilterInput model.

### Describe alternatives you've considered

Filtering after fetching all the data.

### Is this something that you'd be interested in working on?

- [ ] 👋 I may be able to implement this feature request

### Would this feature include a breaking change?

- [ ] ⚠️ This feature might incur a breaking change

bergmorten:
Any update on this feature request. I'm often get in situation where I want to filter out _deleted items?

The same issue is aslo relevant for _createdAt or _updatedAt is not in the filter model...

orome:
Any update on this? Seems like a pretty essential feature.

GabiVeks:
Any update on this feature request. I'm often get in situation where I want to filter out _deleted items?

lien-dkseo:
+1

RikiZimbakov:
+1

efingal:
I am trying to add the _deleted filter in my schema.graphql (there are two files with this name).  Regardless of where I put it, I see it being removed by the initial fetch phase of 'amplify push'.  Is there something I am missing?

danyalxahid:
> I am trying to add the _deleted filter in my schema.graphql (there are two files with this name). Regardless of where I put it, I see it being removed by the initial fetch phase of 'amplify push'. Is there something I am missing?

Having the same problem which adding _Deleted field in the filter input and when i mock graphql it gets removed.

Anyone has a workaround for this? Other than disabling Conflict resolution.

lKrayola:
+1

[...]

PR #1382, opened 2023-04-03
feat(datastore-filters): add filters for DataStore _deleted property

MorCohenAres:
#### Description of changes

Added `_deleted` field to the generated `ModelXXXFilterInput` so people can query the API without the deleted items, for example.

None.

#### Issue #, if available

Implemented #78.

I noticed that people are actively asking in the thread to implement the feature, but wasn't sure why it hasn't been implemented yet.
If there's a special reason or an intentional decision not to implement it (or not do it that way), please let me know.

1. Created an example `Todo` model that has the list/sync query, pushed it and made sure the filter applies in the AppSync console
2. Created the `Todo` model without sync or list, but connected it to another model (`ListOfTodos`) with an `@hasMany` directive. Pushed and made sure the query for `Todo`s through `ListOfTodos` has this field and that it's working.

#### Checklist

- [X] PR description included
- [X] `yarn test` passes

### DIFF
--- a/packages/amplify-graphql-model-transformer/src/__tests__/__snapshots__/model-transformer.test.ts.snap
+++ b/packages/amplify-graphql-model-transformer/src/__tests__/__snapshots__/model-transformer.test.ts.snap
@@ -7,12 +7,18 @@ type Test {
   email: Email
   createdAt: AWSDateTime!
   updatedAt: AWSDateTime!
+  _version: Int!
+  _deleted: Boolean
+  _lastChangedAt: AWSTimestamp!
 }
 
 type Email {
   id: ID!
   createdAt: AWSDateTime!
   updatedAt: AWSDateTime!
+  _version: Int!
+  _deleted: Boolean
+  _lastChangedAt: AWSTimestamp!
 }
 
 input ModelStringInput {
@@ -168,6 +174,7 @@ enum ModelSortDirection {
 type ModelTestConnection {
   items: [Test]!
   nextToken: String
+  startedAt: AWSTimestamp
 }
 
 input ModelTestFilterInput {
@@ -175,31 +182,38 @@ input ModelTestFilterInput {
   and: [ModelTestFilterInput]
   or: [ModelTestFilterInput]
   not: ModelTestFilterInput
+  _deleted: ModelBooleanInput
 }
 
 type Query {
   getTest(id: ID!): Test
   listTests(filter: ModelTestFilterInput, limit: Int, nextToken: String): ModelTestConnection
+  syncTests(filter: ModelTestFilterInput, limit: Int, nextToken: String, lastSync: AWSTimestamp): ModelTestConnection
   getEmail(id: ID!): Email
   listEmails(filter: ModelEmailFilterInput, limit: Int, nextToken: String): ModelEmailConnection
+  syncEmails(filter: ModelEmailFilterInput, limit: Int, nextToken: String, lastSync: AWSTimestamp): ModelEmailConnection
 }
 
 input ModelTestConditionInput {
   and: [ModelTestConditionInput]
   or: [ModelTestConditionInput]
   not: ModelTestConditionInput
+  _deleted: ModelBooleanInput
 }
 
 input CreateTestInput {
   id: ID
+  _version: Int
 }
 
 input UpdateTestInput {
   id: ID!
+  _version: Int
 }
 
 input DeleteTestInput {
   id: ID!
+  _version: Int
 }
 
 type Mutation {
@@ -215,6 +229,7 @@ input ModelSubscriptionTestFilterInput {
   id: ModelSubscriptionIDInput
   and: [ModelSubscriptionTestFilterInput]
   or: [ModelSubscriptionTestFilterInput]
+  _deleted: ModelBooleanInput
 }
 
 type Subscription {
@@ -229,6 +244,7 @@ type Subscription {
 type ModelEmailConnection {
   items: [Email]!
   nextToken: String
+  startedAt: AWSTimestamp
 }
 
 input ModelEmailFilterInput {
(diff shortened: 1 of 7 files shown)


phani-srikar (commented):
Changes look good to me. Only callout is to add some unit and possibly e2e test to modified behavior.

iartemiev (changes requested):
Thanks for opening the PR, @MorCohenAres. 

I see the value in adding the `_deleted` field to the list query filter input specifically. What are the use cases that justify adding `_version` and `_lastChangedAt` as well?

I think that if we make this change, it should only apply to the list query filter input, not the sync query, as DataStore relies on these internal fields and filtering on them can lead to unexpected behavior when using DataStore specifically.

MorCohenAres:
@phani-srikar @iartemiev 
Thank you guys!

1. I added it to the tests :)
2. > I see the value in adding the _deleted field to the list query filter input specifically. What are the use cases that justify adding _version and _lastChangedAt as well?

Giving it a second thought, I think you're right. I narrowed it down to only have the `_deleted` field as a possible filter.

3. > I think that if we make this change, it should only apply to the list query filter input, not the sync query, as DataStore relies on these internal fields and filtering on them can lead to unexpected behavior when using DataStore specifically.

That might be correct. However, the problem with it is that `Sync`, `List` and `Get` queries, all use the same `ModelXXXFilterInput` object.
Making a new separation to declare a new filter object, and testing it well, might be unworthy and significantly larger than the scope that this PR aims to address.

If that's OK with you, my opinion is that this scope should stay narrow. Let me know if you strongly disagree...

Please feel free to review this once again 😄

iartemiev:
@MorCohenAres thank you for making the requested changes.

On the last point - I agree that creating a new filter input object will result in a bigger code change, but I think it's worth it in this case, because it will serve as a guardrail to prevent unintended API use. It's never valid to filter on `_deleted` when using DataStore, so we don't want to implicitly encourage that by exposing that field in the filter input for sync queries. 

Please let us know if you need assistance from the team with this change. We'd be happy to build on top of your PR if that's the case.

MorCohenAres:
@iartemiev Thank you for your comment!
I have made the necessary changes and created a new `ModelSyncXXXFilterInput` type.

Actually it really wasn’t that hard, because of the already-existing separation between `Query` and `Subscription` which we can take influence from.
Also the code in those areas is nice and clean, very comfortable to work with :)

So I updated the tests as well, and you're welcome to review it again.

**NOTE**: 
I'm using the new `ModelSyncXXXFilterInput` only at the sync query level.
In nested fields inside the sync query it's still uses the regular `ModelXXXFilterInput`. This is because the main list of items that is returned from the sync query is the part that requires you not to filter by `_deleted` so you can align your local database with the deleted items. Nested lists of items are not used for this purpose (If you want to sync the nested type you should use the corresponding sync query), so they should use the regular filter.

Please let me know what you think!

phani-srikar (changes requested):
Changes look good. Can you please run `yarn refresh-lockfile` from the root of the repository so that the `yarn.lock` is refreshed? This command basically just deletes the `yarn.lock` file and regenerates it using `yarn` command. Before pushing your change you can verify if the file is up-to date using `yarn verify-yarn-lock` from the repo root.

MorCohenAres:
@dpilch @phani-srikar Fixed `yarn.lock`. Thanks!

dpilch (approved):

phani-srikar (dismissed):

dpilch (approved):

phani-srikar (dismissed):

iartemiev (dismissed):

MorFix:
Thank you for your approvals. Would love to contribute more in the future! ☺️

MorCohenAres:
Thank you for your approvals.
Would love to contribute more in the future! ☺️

phani-srikar:
Running e2e tests on this. [Follow link if you've access](https://app.circleci.com/pipelines/github/aws-amplify/amplify-category-api/5962/workflows/18ab9013-8418-40b5-8fd2-054263fd49d7)

MorCohenAres:
> Running e2e tests on this. [Follow link if you've access](https://app.circleci.com/pipelines/github/aws-amplify/amplify-category-api/5962/workflows/18ab9013-8418-40b5-8fd2-054263fd49d7)

I see that [it failed](https://app.circleci.com/pipelines/github/aws-amplify/amplify-category-api/5962/workflows/1d3af334-fdb8-4078-bea4-5f3d1663f79c/jobs/119117/tests#failed-test-0) bcause of renaming the sync filter input. I have now fixed it.

BUT I'm now realizing that **THIS IS A BREAKING change** for client Amplify SDKs.

For example [here](https://github.com/aws-amplify/amplify-swift/blob/fc9004077a24b7cb52d901c168099d73d1d2d00f/AmplifyPlugins/Core/AWSPluginsCore/Model/Decorator/FilterDecorator.swift#L38) We'll need to change this manually-generated filter name to support!

Same [Here](https://github.com/aws-amplify/amplify-js/blob/268a270a2810a43a8361cbe5c6b02cd2a1f419dd/packages/datastore/src/sync/utils.ts#L360) (JS) and [Here](https://github.com/aws-amplify/amplify-android/blob/3423d4cad9a5f001d24cbec7c364c5be055596a7/aws-datastore/src/main/java/com/amplifyframework/datastore/appsync/AppSyncRequestFactory.java#L145) (Android), and also make sure it works in Flutter.
This should be taken into consideration when people upgrade their CLI with these changes. How should we tackle this?

iartemiev:
@MorCohenAres - you're right, I didn't account for that with the approach I suggested. With that being said, let's revert back to what you had before: a single Input type with `_deleted` added. That way the change will be non-breaking across the platforms . Apologies for the back and forth.

sundersc:
Will adding the filter `_deleted = 0` by default to the `Get` and `List` (when DS enabled) DDB query filter expression address the usecase?

iartemiev:
@MorCohenAres - we discussed this amongst the team and would like to go with the original approach you had - single Input object that's shared across the 3 Query fields. Thanks!

MorFix:
(It's me, Mor. I just abandoned the old account)
@iartemiev OK thank you guys. I have now updated everything according to your latest decision.
Let's try e2e again to see if that works? 😄

dpilch (approved):

iartemiev (approved):

phani-srikar:
Follow [this link](https://app.circleci.com/pipelines/github/aws-amplify/amplify-category-api/6048/workflows/72869782-6b93-457c-84c3-bcc11ba816e4) for latest e2e.

phani-srikar (approved):

phani-srikar:
The `schema-predictions` test failure is a known issue which will be fixed when we re-run the pipeline against latest CLI version.

MorFix:
> The `schema-predictions` test failure is a known issue which will be fixed when we re-run the pipeline against latest CLI version.

Then is it ready to be merged? Are we holding it till the CLI updates in the pipeline?

phani-srikar:
Thanks for the contribution @MorFix. We will keep you updated once this change makes it to Amplify CLI release.

offlineprogrammer:
@MorCohenAres Thank you for your contribution! I would like to follow up with you on this. can you please connect with me on [Twitter](https://twitter.com/_Mo_Malaka_) or [LinkedIn](https://www.linkedin.com/in/momalaka/)

PR #1372, opened 2023-03-27
feat: inject project info into api overrides

sobolk:
#### Description of changes

This PR adds additional parameter to the `override()` that has environment name and project name.
This is a common functionality that has been implemented across other overrides here https://github.com/aws-amplify/amplify-cli/pull/12064 .

Completes coverage for this ask https://github.com/aws-amplify/amplify-cli/issues/9063 .

#### Description of how you validated changes

https://app.circleci.com/pipelines/github/aws-amplify/amplify-category-api?branch=run-e2e%2Fsobkamil%2Finject-project-info

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are [changed or added](https://github.com/aws-amplify/amplify-cli/blob/master/CONTRIBUTING.md#tests)
- [ ] Relevant documentation is changed or added (and PR referenced)
- [ ] New AWS SDK calls or CloudFormation actions have been added to relevant test and service IAM policies
- [ ] Any CDK or CloudFormation parameter changes are called out explicitly

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1417, opened 2023-04-22
feat(api): rds add model, index and default transformers logic

sundersc:
- Merge RDS feature branch to `main`.
- Modify split e2e tests script to run containers test in us-east-1 region.
- Handle s3 bucket ACL issue in predictions test (https://github.com/aws-amplify/amplify-cli/issues/12503).


packages/amplify-category-api/src/index.ts
```diff
@@ -139,6 +147,22 @@ export const initEnv = async (context: $TSContext): Promise<void> => {
...
+    // read and validate the RDS connection parameters
+    const secretsKey = await getSecretsKey();
+
+    const envInfo = {
+      isNewEnv: context.exeInfo?.isNewEnv,
+      sourceEnv: context.exeInfo?.sourceEnvName,
+      yesFlagSet: _.get(context, ['parameters', 'options', 'yes'], false),
+      envName: envName
+    }
+    await configureMultiEnvDBSecrets(context, secretsKey, resourceName, envInfo);
```
phani-srikar, line 163: This wouldn't get called since the RDS schema file doesn't exist?

packages/amplify-graphql-model-transformer/package.json
```diff
@@ -21,7 +21,7 @@
     "access": "public"
   },
   "scripts": {
-    "build": "tsc",
+    "build": "tsc && cd rds-lambda && mkdir -p node_modules && rm -rf node_modules && npm install && tsc && cp -r node_modules lib && cd lib && bestzip --force node ../../lib/rds-lambda.zip ./*",
```
phani-srikar, line 24: do we still need this?

scripts/split-e2e-tests.ts
```diff
@@ -216,8 +220,9 @@ function splitTests(
 
   const newJobs = testSuites.reduce((acc, suite, index) => {
     const newJobName = generateJobName(jobName, suite);
-    const testRegion = FORCE_US_WEST_2.some(job => newJobName.startsWith(job))
-      ? 'us-west-2'
+    const forceRegion = Object.keys(FORCE_REGION).find(key => newJobName.startsWith(key));
+    const testRegion = forceRegion
+      ? FORCE_REGION[forceRegion as FORCE_TESTS]
```
phani-srikar, line 225: these changes are new to me. Any context here?

packages/graphql-transformer-core/src/util/transformConfig.ts
```diff
@@ -254,20 +259,40 @@ export async function loadProject(projectDirectory: string, opts?: ProjectOption
...
+    }
+  } else if (fs.existsSync(schemaDirectoryPath)) {
+    // Schema folder is used only for DynamoDB datasource
+    const datasourceType = constructDataSourceType("DDB");
+    const schemaInDirectory = (await readSchemaDocuments(schemaDirectoryPath)).join('\n');
+    modelToDatasourceMap = new Map([...modelToDatasourceMap.entries(), ...constructDataSourceMap(schemaInDirectory, datasourceType).entries()]);
+    schema += schemaInDirectory;
   } else {
-    throw new Error(`Could not find a schema at ${schemaFilePath}`);
+    throw new Error(`Could not find a schema at ${schemaFilePaths[0]}`);
```
phani-srikar, line 290: If the `schemaFilePaths` is empty, this is bound to throw!

phani-srikar (changes requested):

packages/amplify-graphql-model-transformer/package.json
sundersc: This is going to change in the upcoming PR as we are moving this logic to lambda layer. This is the current state.

scripts/split-e2e-tests.ts
sundersc: This change forces the containers test to run in us-east-1 region.

`containers` feature is not supported in all the regions. We officially launched it for us-east-1 only but there are other regions the feature is available now. Forcing the test to use the supported region fix test flakiness.

packages/graphql-transformer-core/src/util/transformConfig.ts
sundersc: SchemaFilePath array can't be empty, we add the api resource path just few lines before this logic.

packages/amplify-category-api/src/index.ts
sundersc: Don't be call this only on commands 'generate-schema' or 'update-secrets'? Both the commands are disabled now.

phani-srikar: We also call this during the env add to automatically copy over the secrets to née environment.

sundersc: The secrets wouldn't exist on the current environment. Do you still create copy over to the new environment with blank values?

phani-srikar: That's right. This wouldn't be triggered w/o a RDS schema

packages/amplify-category-api/src/provider-utils/awscloudformation/cfn-api-artifact-handler.ts
```diff
@@ -97,20 +97,23 @@ class CfnApiArtifactHandler implements ApiArtifactHandler {
...
+    this.ensureCfnParametersExist(resourceDir, apiParameters);
+    this.context.amplify.updateamplifyMetaAfterResourceAdd(category, serviceConfig.apiName, this.createAmplifyMeta(authConfig, dependsOn));
 
-    await this.context.amplify.executeProviderUtils(this.context, 'awscloudformation', 'compileSchema', {
-      resourceDir,
-      parameters: this.getCfnParameters(serviceConfig.apiName, authConfig, resourceDir),
-      authConfig,
-    });
+    if(serviceConfig?.transformSchema) {
+      // write the template buffer to the project folder
```
alharris-at, line 107: I've squinted at this line for a while, and I can't convince myself that I fully understand the impact (because this is behind the whole service/walkthrough/request generator rigamarole.

If for some reason the schema is not always filled in the same way could this result in us failing to write and compile?

packages/amplify-graphql-index-transformer/src/__tests__/__snapshots__/amplify-graphql-index-transformer.test.ts.snap
```diff
@@ -1184,9 +1181,6 @@ $util.toJson({})
 #end
 #if( !$util.isNullOrEmpty($filter) )
   #set( $filterExpression = $util.parseJson($util.transform.toDynamoDBFilterExpression($filter)) )
-  #if( $util.isNullOrEmpty($filterExpression) )
-    $util.error(\\"Unable to process the filter expression\\", \\"Unrecognized Filter\\")
-  #end
```
alharris-at, line 1189: We don't want to revert this, do we?

alharris-at (commented):
Went through primarily the core code that's been touched (didn't pay much attention to the pure RDS stuff since that's been reviewed in isolation).

Dropping a few questions about some of the overall stuff, but the three primary questions are about 1) conditionally compiling the schema, 2) stripping out the `Amplify` schema input object, and 3) The reverted snapshots with the `isNullOrEmpty( inputFilter )` bit.

packages/amplify-graphql-transformer-core/package.json
```diff
@@ -41,6 +41,7 @@
     "vm2": "^3.9.8"
   },
   "peerDependencies": {
+    "@aws-amplify/amplify-cli-core": "4.0.1",
```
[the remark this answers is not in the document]
sundersc: amplify-cli-core has been used in amplify-graphql-transformer-core without defining the dependency explicitly. In this PR, we are defining it explicitly so that the build tree resolves the dependency correctly.

Here is a previously merged and released PR using amplify-cli-core package's stateManager.
https://github.com/aws-amplify/amplify-category-api/pull/1372

packages/amplify-graphql-transformer-core/src/utils/rds-input-utils.ts
```diff
@@ -0,0 +1,133 @@
...
+    return;
+  }
+
+  const parsedSchema = parse(schemaContent);
+
+  const inputNode = parsedSchema.definitions.find(
+    (definition) =>
+      definition.kind === 'InputObjectTypeDefinition' &&
+      definition.name &&
+      definition.name.value === 'Amplify'
```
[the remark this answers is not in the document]
sundersc: I agree but I would like to reserve this for a follow up PR.

packages/amplify-graphql-transformer-core/src/transformation/transform.ts
```diff
@@ -44,7 +44,7 @@ import { adoptAuthModes, IAM_AUTH_ROLE_PARAMETER, IAM_UNAUTH_ROLE_PARAMETER } fr
 import * as SyncUtils from './sync-utils';
 import { MappingTemplate } from '../cdk-compat';
 
-import { UserDefinedSlot, OverrideConfig } from './types';
+import { UserDefinedSlot, OverrideConfig, DatasourceTransormationConfig  } from './types';
```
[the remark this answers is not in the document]
sundersc: Spelling has been corrected.

packages/amplify-graphql-transformer-core/src/utils/rds-input-utils.ts
```diff
@@ -0,0 +1,133 @@
...
+  , '');
+  return `input Amplify {\n${inputsString}}\n`;
+};
+
+export const readRDSGlobalAmplifyInput = async (pathToSchemaFile: string): Promise<InputObjectTypeDefinitionNode | undefined> => {
+  if (!fs.existsSync(pathToSchemaFile)) {
+    return;
+  }
+  const schemaContent = fs.readFileSync(pathToSchemaFile, 'utf-8');
+  if (_.isEmpty(schemaContent?.replace(/[\r\n]/gm, ''))) {
```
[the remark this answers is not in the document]
sundersc: This still is an atomic operation. I don't think this benefits from moving to a new function.

packages/amplify-graphql-index-transformer/src/__tests__/__snapshots__/amplify-graphql-index-transformer.test.ts.snap
sundersc: Good catch. This was removed in error during rebase. I've added it back.

packages/amplify-graphql-transformer-core/src/transformation/transform.ts
```diff
@@ -481,7 +485,8 @@ export class GraphQLTransform {
         functions[templateName.replace('functions/', '')] = template;
       }
     }
-    const schema = fileAssets.get('schema.graphql') || '';
+    const compiledSchema = fileAssets.get('schema.graphql') || '';
+    const schema = removeAmplifyInputDefinition(compiledSchema);
```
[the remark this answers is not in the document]
phani-srikar: It does strip it away but not from the input schema but from the compiled schema. Post compilation if we have the `Amplify` input object in the schema it will fail during deployment since that type is not recognized by AppSync.

packages/amplify-category-api/src/provider-utils/awscloudformation/cfn-api-artifact-handler.ts
phani-srikar: When we add/import RDS API, this value in the `serviceConfig` which primarily is associated with `schema.graphql` contents is empty/non-existent and without this additional check that flow fails.
For non RDS flows, we need to assess the impact/DX change.

sundersc: This doesn't have any impact on the DDB scenarios. The existing "Blank Schema" option in add api steps adds the below schema. We are checking whether the schema has a value before compiling it.

```
# This "input" configures a global authorization rule to enable public access to
# all models in this schema. Learn more about authorization rules here: https://docs.amplify.aws/cli/graphql/authorization-rules
input AMPLIFY { globalAuthRule: AuthRule = { allow: public } } # FOR TESTING ONLY!
```

phani-srikar (dismissed):

packages/amplify-graphql-schema-generator/src/datasource-adapter/mysql-datasource-adapter.ts
```diff
@@ -0,0 +1,296 @@
...
+          acquireTimeoutMillis: 30000,
+          idleTimeoutMillis: 30000,
+          reapIntervalMillis: 1000,
+          createRetryIntervalMillis: 100
+        },
+        debug: false,
+      });
+    }
+    catch(err) {
+      printer.info(err);
```
dpilch, line 77:
```suggestion
      printer.error(err);
```

dpilch (commented):
Overall LGTM. I don't feel qualified to give an approval due to my own lack of knowledge on the subject. Left a few follow-up nits.

packages/amplify-graphql-transformer-core/src/utils/rds-input-utils.ts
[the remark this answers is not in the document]
alharris-at: I'm less worried for this one line about the DRYness of the code, but specifically that we check `_.isEmpty(schemaContent)` and `_isEmpty(schemaContent?.replace(/[\r\n]/gm, ''))` at different points in this file, that difference in understand about what an empty schema is seems like a smell.

alharris-at (approved):
LGTM - as mentioned before, focused on the snapshot tests for DDB, and verifying we don't see a material shift in generated behavior for existing stuff.

packages/amplify-graphql-transformer-core/src/utils/rds-input-utils.ts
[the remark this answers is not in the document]
sundersc: Looks we are removing the blank lines to check if the schema file is actually empty. I agree that the logic is not consistent across the file, updated to check only the schema content. If required we can create a new utility method to perform additional checks and use it across the transformers.

packages/amplify-graphql-schema-generator/src/datasource-adapter/mysql-datasource-adapter.ts
```diff
@@ -0,0 +1,296 @@
...
+    const columnResult = (await this.dbBuilder.raw(`SELECT * FROM INFORMATION_SCHEMA.COLUMNS WHERE TABLE_SCHEMA = '${this.config.database}'`))[0];
+    this.fields = columnResult.map((item: any) => {
+      return {
+        tableName: item["TABLE_NAME"],
+        columnName: item["COLUMN_NAME"],
+        default: item["COLUMN_DEFAULT"],
+        sequence: item["ORDINAL_POSITION"],
+        datatype: item["DATA_TYPE"],
+        columnType: item["COLUMN_TYPE"],
+        nullable: item["IS_NULLABLE"] === 'YES' ? true : false,
```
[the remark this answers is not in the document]
sundersc: Oh yeah! Updated the logic.

alharris-at (approved):

Issue #1428, opened 2023-04-29
Bug: v2 Transformer Generates optional nested `id` fields for Input types instead of Required


danrivett:
### How did you install the Amplify CLI?

npm -g i @aws-amplify/cli

### Amplify CLI Version

11.1.0

### What operating system are you using?

Mac

### Did you make any manual changes to the cloud resources managed by Amplify? Please describe the changes made.

No

### Describe the bug

Given the following schema:

```graphql
type Customer @model {
  id: ID!
  firstName: String!
  lastName: String!
  paymentMethods: [PaymentMethod!]!
}

type PaymentMethod {
  id: ID!
  cardMask: String!
  expiryDate: String!
}
```

When running `amplify codegen` the following Input types are generated:

```graphql
input CustomerInput {
  id: ID
  firstName: String!
  lastName: String!
  paymentMethods: [PaymentMethod!]!
}

input PaymentMethodInput {
  id: ID # <-- Here is the problem, id should be required as it's a nested type
  cardMask: String!
  expiryDate: String!
}
```

As shown above the `id` field on both `CustomerInput` and `PaymentMethodInput` are defined as required on the type definitions but optional on the corresponding input definitions.

It makes sense for `CustomerInput.id` to be optional, since it's a top level type and and if not specified it will get automatically generated by Amplify DataStore (or request resolver), but it doesn't make sense for `PaymentMethodInput.id` to be optional, since it doesn't get auto-generated and so can lead to an invalid `PaymentMethod` being created if the `id` field is not specified on the input.

Note: We've been running the latest `@aws-amplify/cli` versions for many months, but this only started happening when we recently attempted again at upgrading to v2 of the GraphQL Transformer (this is our 3rd attempt as each time we ran into bugs or issues such as this one).

This is currently causing compile errors in our integration code which uses these generated types.

### Expected behavior

`id` fields on nested `input` types should not have their type overridden and marked optional, but should remain required.

The following input types should be generated:

```graphql
input CustomerInput {
  id: ID # <-- This can be made optional like it is today
  firstName: String!
  lastName: String!
  paymentMethods: [PaymentMethod!]!
}

input PaymentMethodInput {
  id: ID! # <-- This should remain required
  cardMask: String!
  expiryDate: String!
}
```

### Reproduction steps

See the GraphQL schema above for reproduction steps.

[...]

PR #1429, opened 2023-04-29
fix(graphql): excluding ids on nested inputs generated from being made optional

danrivett:
#### Description of changes

Potential fix for #1428.

This PR updates GraphQL Input type generation to only make `id` fields optional for top-level types and not nested-types (see #1428 for more info). It does it by checking if the type is annotated with `@model` as seemed the simplest and least invasive way of implementing that, but I'm not sure if there are edge cases that also need to be considered.

Two things to note:

1. This PR doesn't contain unit tests to verify the fix as I wanted to first show a possible solution before I go to the effort of writing them.
2. I am not clear how to build and run `amplify codegen` with this change against a sample project, to verify it generates the correct `input` types.

None

#### Issue #, if available

#1428

Not validated yet. I'm creating this PR to validate the design approach first.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-graphql-transformer-core/src/wrappers/object-definition-wrapper.ts
+++ b/packages/amplify-graphql-transformer-core/src/wrappers/object-definition-wrapper.ts
@@ -148,7 +148,8 @@ export class InputFieldWrapper extends GenericFieldWrapper {
     };
   };
 
-  static fromField = (name: string, field: FieldDefinitionNode, document: DocumentNode): InputFieldWrapper => {
+  static fromField = (name: string, field: FieldDefinitionNode, parent: ObjectTypeDefinitionNode,
+    document: DocumentNode): InputFieldWrapper => {
     const autoGeneratableFieldsWithType: Record<string, string[]> = {
       id: ['ID'],
       createdAt: ['AWSDateTime', 'String'],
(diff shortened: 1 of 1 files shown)


dpilch:
Hi @danrivett, you can use your changes locally by running `yarn setup-dev`. This will add `amplify-dev` to your path with your changes (`amplify-dev codegen`).

alharris-at (approved):
Approving pending verification from @dpilch that the current case is a broken state, and this doesn't need to be modelled as a MV/breaking change (since we're making input types more restrictive).

dpilch:
It looks like current state is broken.

The current generated mutations include `id` in the selection set for `paymentMethods`.

```typescript
export const createCustomer = /* GraphQL */ `
  mutation CreateCustomer(
    $input: CreateCustomerInput!
    $condition: ModelCustomerConditionInput
  ) {
    createCustomer(input: $input, condition: $condition) {
      id
      firstName
      lastName
      paymentMethods {
        id
        cardMask
        expiryDate
      }
      createdAt
      updatedAt
    }
  }
`;
```

The following query would then result in an error.
```graphql
mutation MyMutation {
  createCustomer(input: {firstName: "firstname", lastName: "lastname", paymentMethods: {cardMask: "mask", expiryDate: "date"}}) {
    id
    lastName
    firstName
    paymentMethods {
      cardMask
      expiryDate
      id
    }
  }
}
```

```json
{
  "data": {
    "createCustomer": null
  },
  "errors": [
    {
      "path": [
        "createCustomer",
        "paymentMethods",
        0,
        "id"
      ],
      "locations": null,
      "message": "Cannot return null for non-nullable type: 'ID' within parent 'PaymentMethod' (/createCustomer/paymentMethods[0]/id)"
    }
  ]
}
```

Updating the schema to require `id` on `PaymentMethodInput` and keeping the same query, we see a validation error on the id field (which is expected).

```json
{
  "data": null,
  "errors": [
    {
      "path": null,
      "locations": [
        {
          "line": 2,
          "column": 18,
          "sourceName": null
        }
      ],
      "message": "Validation error of type WrongType: argument 'input.paymentMethods' with value 'ObjectValue{objectFields=[ObjectField{name='firstName', value=StringValue{value='firstname'}}, ObjectField{name='lastName', value=StringValue{value='lastname'}}, ObjectField{name='paymentMethods', value=ObjectValue{objectFields=[ObjectField{name='cardMask', value=StringValue{value='mask'}}, ObjectField{name='expiryDate', value=StringValue{value='date'}}]}}]}' is missing required fields '[id]' @ 'createCustomer'"
    }
  ]
}
```

So I think it would be reasonable to release this as a bugfix without a major version change.

---

We will need an additional change to fully fix the functionality. The generated TypeScript types also need to be updated.

```typescript
export type PaymentMethodInput = {
  id?: string | null,
  cardMask: string,
  expiryDate: string,
};
```

This allows the following to pass type checking. The request would then meet the validation error at runtime.

```typescript
const input: CreateCustomerInput = {
  firstName: "firstname",
  lastName: "lastname",
  paymentMethods: [{ cardMask: "mask", expiryDate: "date" }],
};
const customer = await API.graphql<GraphQLQuery<CreateCustomerMutation>>({
  query: mutations.createCustomer,
  variables: { input },
});
```

This change would be done in [amplify-codegen](https://github.com/aws-amplify/amplify-codegen). The same change probably needs to be done to other platforms as well, but I have only tested TypeScript.

@danrivett you are welcome to make the change in [amplify-codegen](https://github.com/aws-amplify/amplify-codegen) or leave it up to us, whichever you prefer.

dpilch (approved):

dpilch:
My mistake, no codegen change is needed. I just recently changed to working on CLI packages so I am still learning the ropes!

The PR should be good to merge then.

PR #1439, opened 2023-05-04
refactor: remove amplify-prompts from transformers

dpilch:
#### Description of changes

* Remove `amplify-prompts` peer dep from transformers.
* Add log buffer to transformers that can then be retrieved later.
  * amplify-category-api reads log buffer and then logs the messages using `amplify-prompts`.

I have not removed `amplify-prompts` from `@aws-amplify/graphql-transformer-migrator` yet. I will do that in another PR if necessary.

N/A

#### Issue #, if available

N/A

* Unit testing.

Needs further testing.

#### Checklist

- [x] PR description included
- [ ] `yarn test` passes


packages/graphql-transformer-core/src/collectDirectives.ts
```diff
@@ -66,14 +65,7 @@ export function collectDirectivesByType(sdl: string): Object {
     return {};
   }
   let doc;
-  try {
-    doc = parse(sdl);
-  } catch (e) {
```
dpilch, line 71: This catch block is unreachable in the current state.

alharris-at:
> I have not removed amplify-prompts from @aws-amplify/graphql-transformer-migrator yet. I will do that in another PR if necessary.

I don't think that'll be necessary, since the graphql-transformer-migrator should be principally concerned w/ the Amplify state management, not the transformer internals itself.

packages/amplify-graphql-transformer-core/src/transformation/transform.ts
```diff
@@ -303,6 +308,18 @@ export class GraphQLTransform {
       }
       reverseThroughTransformers -= 1;
     }
+    // is this the right place?
```
alharris-at, line 311: I think so, this is after the visit is complete for the transform.

packages/amplify-graphql-transformer-core/src/transformation/transform.ts
```diff
@@ -303,6 +308,18 @@ export class GraphQLTransform {
...
+    for (const transformer of this.transformers) {
+      if (isFunction(transformer.getLogs)) {
+        const logs = transformer.getLogs();
+        if (logs.length) {
```
alharris-at, line 315: Is this check required? I'd assume getLogs returns Log[], in which case you can just push the potentially empty array

packages/amplify-graphql-transformer-core/src/transformation/transformer-plugin-base.ts
```diff
@@ -51,6 +54,19 @@ export abstract class TransformerPluginBase implements TransformerPluginProvider
     // Transformers can define extra shapes that can be used by the directive
     // and validated. TODO: Validation.
     this.typeDefinitions = extraDefinitions;
+    this.logs = [];
+  }
+
+  protected log(level: TransformerLogLevel, message: string) {
+    this.logs.push({ level, message });
+  }
```
alharris-at, line 62: Perhaps vend a 'warn', 'info', etc. to reflect the log levels? Instead of passing it in at each step.

packages/graphql-transformer-core/src/collectDirectives.ts
alharris-at: Is this because of something specific in the `parse` impl? It's not obvious why this would be unreachable, unless I'm missing something obvious. Also if you're no longer adding the try block, etc. can the let and parse be consolidated into a single line (or better yet, can it be made a const?)

alharris-at (commented):
Overall LGTM - a few q's on the new Logging APIs.

Also, is the `flush` method necessary? Is there a particular reason you want to purge the logs?

packages/amplify-category-api/src/graphql-transformer/transform-graphql-schema-v2.ts
```diff
@@ -250,6 +251,25 @@ const _buildProject = async (context: $TSContext, opts: TransformerProjectOption
...
+      case TransformerLogLevel.INFO:
+        printer.info(log.message);
+        break;
+      case TransformerLogLevel.DEBUG:
+        printer.debug(log.message);
+        break;
+      default:
+        printer.error(log.message);
+    }
+  });
```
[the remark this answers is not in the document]
dpilch: I've added a try finally to fix this. I'm still working on a unit test.

dpilch:
I added `flushLogs` because I wasn't sure if there would be a case where the same transformer instance would be used multiple times (potentially keeping previous warnings).

packages/graphql-transformer-core/src/collectDirectives.ts
dpilch: Sorry, I should have added some for details. It is unreachable in the scenario of running from the CLI. The code would fail somewhere else before reaching this `parse` to throw on.

But now I am doubting myself, so I'll take another look.

packages/amplify-graphql-searchable-transformer/src/nodeToNodeEncryption.ts
```diff
@@ -15,32 +14,32 @@ import { printer } from '@aws-amplify/amplify-prompts';
  * @param apiName the name of the api to attempt and pull the flag from.
  * @returns whether or not NodeToNodeEncryption should be enabled on a searchable instance.
```
AaronZyLee, line 15: The comment here needs change as well as the return type is changed

packages/amplify-graphql-searchable-transformer/src/nodeToNodeEncryption.ts
```diff
@@ -15,32 +14,32 @@ import { printer } from '@aws-amplify/amplify-prompts';
...
-      return nodeToNodeEncryptionParameter;
+      return { enabled: nodeToNodeEncryptionParameter, log };
     }
 
-    return doesExistingBackendHaveNodeToNodeEncryption;
+    return { enabled: doesExistingBackendHaveNodeToNodeEncryption, log };
   } catch (e) {
     // Fail open, and don't set the flag for the purposes of this workaround phase.
-    return false;
+    return { enabled: false, log: undefined };
```
AaronZyLee, line 31: Should the log be set with the error msg in this case?

alharris-at (dismissed):

packages/amplify-graphql-searchable-transformer/src/nodeToNodeEncryption.ts
dpilch: I was trying to keep functionality the same with the refactor. I can update this with a later PR though.

AaronZyLee (approved):

alharris-at (approved):

PR #1441, opened 2023-05-04
chore(api): test with cli rc version

sundersc:
Test API repo pipeline with CLI RC@12.0.0-rc.0d6bca41ff.0.
- Bump up amplify-cli-core resolution to the latest.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -35,7 +35,7 @@ install_cli_from_local_registry: &install_cli_from_local_registry
     startLocalRegistry "$(pwd)/.circleci/verdaccio.yaml"
     setNpmRegistryUrlToLocal
     changeNpmGlobalPath
-    npm install -g @aws-amplify/cli-internal@11.0.3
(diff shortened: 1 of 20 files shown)


.circleci/config.yml
```diff
@@ -35,7 +35,7 @@ install_cli_from_local_registry: &install_cli_from_local_registry
     startLocalRegistry "$(pwd)/.circleci/verdaccio.yaml"
     setNpmRegistryUrlToLocal
     changeNpmGlobalPath
-    npm install -g @aws-amplify/cli-internal@11.0.3
+    npm install -g @aws-amplify/cli-internal@12.0.0-rc.0d6bca41ff.0
```
AaronZyLee, line 38: Do we need to change the version in setup-dev of `package.json` as well?

AaronZyLee (approved):

.circleci/config.yml
sundersc: That's right, updated now. It got reverted when I tested build locally with setup-dev command. Good catch.

phani-srikar (approved):

PR #1456, opened 2023-05-12
refactor: remove/replace some amplify-cli-core deps

dpilch:
#### Description of changes

This PR is based on https://github.com/aws-amplify/amplify-category-api/pull/1439

* Remove `AmplifyError` and replace with standard Error
* Copy `JSONUtilities` from cli
  * The `JSONUtilities` don't have any Amplify specific functionality.
  * I decided to copy here because there would not be an issue if the functionality diverges from the CLI utilities over time 
* Move references to feature flags to `amplify-category-api` and pass down
* Remove use of `AmplifyCategories` and add `APICategory` constant.
* Copy `$TSAny` and `$TSObject`.

To do in future PRs:
* remove `stateManager`
* remove `pathManager`

N/A

#### Issue #, if available

N/A

Unit testing

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-e2e-core/package.json
```diff
@@ -22,12 +22,16 @@
...
+    "@aws-amplify/graphql-transformer-core": "^1.2.1",
+    "@aws-sdk/client-ec2": "^3.226.0",
+    "@aws-sdk/client-rds": "^3.226.0",
     "amplify-headless-interface": "^1.17.3",
     "chalk": "^4.1.1",
     "execa": "^5.1.1",
     "fs-extra": "^8.1.0",
     "graphql-transformer-core": "8.0.3",
     "jest-environment-node": "^26.6.2",
+    "knex": "~2.3.0",
```
phani-srikar, line 34: I think we recently bumped this to `2.4.0` for a security alert. [Refer](https://github.com/aws-amplify/amplify-category-api/pull/1440/files)

packages/amplify-e2e-core/package.json
```diff
@@ -40,8 +44,5 @@
     "@aws-sdk/client-ec2": "^3.226.0",
     "@aws-sdk/client-rds": "^3.226.0",
     "knex": "~2.4.0"
-  },
-  "peerDependencies": {
-    "@aws-amplify/amplify-cli-core": "4.0.4"
```
phani-srikar, line 45: I see that you're still using `stateManager` etc from this dep, don't you want to keep it?

packages/amplify-graphql-transformer-core/src/types/types.ts
```diff
@@ -11,3 +11,13 @@ export type StackMeta = {
...
+
+/**
+ *  Use it for all 'any's where we can't define the type, but doing a strict TypeScript conversion
+ */
+export type $TSAny = any; // eslint-disable-line  @typescript-eslint/no-explicit-any
+
+/**
+ * Use it for all object initializer usages: {}
+ */
+export type $TSObject = Record<string, $TSAny>;
```
phani-srikar, line 23: The other place where I would consider placing these is `amplify-graphql-transformer-interfaces` which we explicitly use to store common interfaces and types. Depending on what you think and others feel, I'll leave it upto you.

packages/amplify-graphql-transformer-migrator/src/utils.ts
```diff
@@ -1,4 +1,5 @@
-import { $TSObject, AmplifySupportedService, stateManager } from '@aws-amplify/amplify-cli-core';
+import { AmplifySupportedService, stateManager } from '@aws-amplify/amplify-cli-core';
```
phani-srikar, line 1: This `AmplifySupportedService` looks like an easy lift here? Maybe we can move it to interfaces or core?

phani-srikar (changes requested):
Some minor recommendations. The number of places where we have these peer dependencies felt less that than what I expected. Are these the only packages with `cli-core` dep?

packages/amplify-e2e-core/package.json
dpilch: Bad rebase, will fix.

packages/amplify-graphql-transformer-migrator/src/utils.ts
dpilch: I'll take a look.

packages/amplify-graphql-transformer-core/src/types/types.ts
dpilch: I think it makes sense to move there.

alharris-at:
Two high-level pieces of feedback before digging in.

> Remove AmplifyError and replace with standard Error

Could you ping Akshay about this? I think this was introduced to allow the CLI to report on specific error types in dashboards accurately (otherwise it just comes back as an 'error' potentially).

> Copy $TSAny and $TSObject

Let's nix `$TSAny` and just use `any`. It was created (I believe) to avoid linting issues, but it's still just `any`. I think the same goes for `$TSObject`, but I'm willing to leave that in instead of creating `{ |key: string|: any }` everywhere, but I'll leave that up to you.

I'm willing to leave the second as a potential followup though.

packages/amplify-graphql-transformer-core/src/transformation/transform.ts
```diff
@@ -392,11 +392,7 @@ export class GraphQLTransform {
     try {
       sandboxNode.run(overrideCode, overrideFilePath).override(appsyncResourceObj, projectInfo);
     } catch (err) {
-      throw new AmplifyError('InvalidOverrideError', {
-        message: 'Executing overrides failed.',
```
alharris-at, line 396: As mentioned at the top-level, let's double-click that this is still being caught up-top and transformed into a well-formed AmplifyError.

packages/graphql-transformer-core/src/util/transformConfig.ts
```diff
@@ -288,10 +287,7 @@ export async function readSchema(projectDirectory: string): Promise<{schema: str
     modelToDatasourceMap = new Map([...modelToDatasourceMap.entries(), ...constructDataSourceMap(schemaInDirectory, datasourceType).entries()]);
     schema += schemaInDirectory;
   } else {
-    throw new AmplifyError('ApiCategorySchemaNotFoundError', {
-      message: `Could not find a schema at ${schemaFilePaths[0]}`,
-      link: 'https://docs.amplify.aws/cli/graphql/overview/#update-schema',
-    });
+    throw new Error(`Could not find a schema at ${schemaFilePaths[0]}. https://docs.amplify.aws/cli/graphql/overview/#update-schema`);
```
alharris-at, line 290: Same comment on errors here.

dpilch:
I found we have an error converter: https://github.com/aws-amplify/amplify-category-api/blob/main/packages/amplify-category-api/src/graphql-transformer/transform-graphql-schema.ts#LL21C11-L21C50

I've replaced these two errors with errors that will be converted to `AmplifyError`.

I'll replace all `$TSAny`/`$TSObject` in a follow up PR.

dpilch:
`$TSAny` change: https://github.com/aws-amplify/amplify-category-api/pull/1481

phani-srikar (approved):

PR #1457, opened 2023-05-12
fix(auth): owner & group field as sort key field of GSI

AaronZyLee:
#### Description of changes

- Pass the key schema infos when generating auth expressions
- Add owner/group auth field with `eq` when it is part of index sort key and missing in the args input
  - When the field is missing in input, it will retrieve the last item of claim list by default
- Add unit/e2e tests

Unit & e2e tests
AppSync Console testing
#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-graphql-auth-transformer/src/resolvers/query.ts
+++ b/packages/amplify-graphql-auth-transformer/src/resolvers/query.ts
@@ -236,8 +236,13 @@ const generateAuthOnModelQueryExpression = (
             methodCall(ref('util.isNull'), ref('ctx.stash.authFilter')),
             not(ref('primaryFieldMap.isEmpty()')),
           ]),
-          sortKeyFields.length > 0
-          ? compoundExpression([
+          /**
+           * The following code block is to handle the cases of dynamic claim field(owner/group)
+           * being part of the index sort key field and is missing from the query input arguments.
+           * If detected, the last item of the claim list will be added by default to the query condition with "eq".
+           * This restricts the records of other users returned from the query in such scenario
+           */
+          compoundExpression([
             set(ref('sortKeyFields'), list(sortKeyFields.map(sk => str(sk)))),
             forEach(ref('entry'), ref('primaryFieldMap.entrySet()'), [
               ifElse(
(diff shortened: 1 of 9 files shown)


phani-srikar (dismissed):

packages/amplify-graphql-auth-transformer/src/resolvers/query.ts
```diff
@@ -234,11 +236,28 @@ const generateAuthOnModelQueryExpression = (
...
+            set(ref('sortKeyFields'), list(sortKeyFields.map(sk => str(sk)))),
             forEach(ref('entry'), ref('primaryFieldMap.entrySet()'), [
-              qref(methodCall(ref('ctx.args.put'), ref('entry.key'), ref('entry.value'))),
+              ifElse(
+                methodCall(ref('sortKeyFields.contains'), ref('entry.key')),
+                compoundExpression([
+                  set(ref('entryVal'), ref('entry.value')),
+                  set(ref('lastIdx'), ref('entryVal.size() - 1')),
+                  set(ref('lastItem'), ref('entryVal.get($lastIdx)')),
+                  qref(methodCall(ref('ctx.args.put'),  ref('entry.key'), obj({ eq: ref('lastItem') }))),
```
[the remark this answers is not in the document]
sundersc: It's worth adding comments here to help us in the future.

packages/amplify-graphql-auth-transformer/src/resolvers/query.ts
```diff
@@ -234,11 +236,28 @@ const generateAuthOnModelQueryExpression = (
             methodCall(ref('util.isNull'), ref('ctx.stash.authFilter')),
             not(ref('primaryFieldMap.isEmpty()')),
           ]),
-          compoundExpression([
+          sortKeyFields.length > 0
```
sundersc, line 239: Do we need this ternary operator here? The `IF` block `methodCall(ref('sortKeyFields.contains'), ref('entry.key')),` should take care of this check if it is initialized with an empty array.

packages/graphql-transformers-e2e-tests/src/__tests__/IndexWithAuthV2WithFF.e2e.test.ts
```diff
@@ -271,7 +271,7 @@ test('listX with primaryKey', async () => {
 
   listResponse = await listFamilyMembersByParent(GRAPHQL_CLIENT_2, { parent: USERNAME1 });
   items = listResponse.data.byParent.items;
-  expect(items).toHaveLength(2);
+  expect(items).toHaveLength(1);
```
AaronZyLee, line 274: This is the expected change due to the restricted records returned. It used to return both `sub::username` and `username` records. However, since the `child` is the sort key field of index `parent` and child is an owner field and missing in the input, it now resolves to the last item in the claim list, which cannot return both at the same time.

sundersc (approved):

phani-srikar (approved):

PR #1486, opened 2023-05-19
refactor: remove pathManager

dpilch:
#### Description of changes

Remove `pathManager` by passing paths down to transformers.

`pathManager` functions used in transformers:
* `getBackendDirPath`
* `findProjectRoot`
* `getCurrentCloudBackendDirPath`

Transformers that used `pathManager`:
* IndexTransformer (`getBackendDirPath`)
* PrimaryKeyTransformer (`getBackendDirPath`)
* SearchableModelTransformer (`findProjectRoot` and `getCurrentCloudBackendDirPath`)

IndexTransformer and PrimaryKeyTransformer use the backend path to get the overrides: https://github.com/aws-amplify/amplify-category-api/blob/main/packages/amplify-graphql-index-transformer/src/resolvers/resolvers.ts#L891

SearchableModelTransformer uses the pathManager in order to determine if node to node encryption should be enabled.
* https://github.com/aws-amplify/amplify-category-api/blob/main/packages/amplify-graphql-searchable-transformer/src/nodeToNodeEncryption.ts#L47
* https://github.com/aws-amplify/amplify-category-api/blob/main/packages/amplify-graphql-searchable-transformer/src/nodeToNodeEncryption.ts#L70

These probably could be pulled up to the transformer factory but I think would require a bit more effort. I think this refactor is sufficient until we work to remove file system access, but would like to know others opinions.

N/A

#### Issue #, if available

N/A

* Unit tests
* E2E tests: https://app.circleci.com/pipelines/github/aws-amplify/amplify-category-api/6626/workflows/c36409c3-da30-44a5-8e77-107bcdb50026

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


alharris-at (dismissed):

alharris-at:
Could you run e2e tests as well on this change?

sundersc (changes requested):
IMO passing the path values through transformer arguments deviate from our current design. Can we pass these values through transformer context?

dpilch:
I've updated to pass the filepaths through the context.

sundersc (dismissed):

packages/amplify-graphql-index-transformer/src/graphql-index-transformer.ts
```diff
@@ -69,7 +69,7 @@ export class IndexTransformer extends TransformerPluginBase {
   public after = (ctx: TransformerContextProvider): void => {
     if (!ctx.isProjectUsingDataStore()) return;
 
-    const overriddenResources = getResourceOverrides([this], ctx?.stackManager);
+    const overriddenResources = getResourceOverrides([this], ctx.filepaths.getBackendDirPath(), ctx?.stackManager);
```
alharris-at, line 72: ctx is optional in the latter, can we rely on it in the former?

packages/amplify-util-mock/src/__e2e_v2__/model-relational-transformer.e2e.test.ts
```diff
@@ -6,6 +6,7 @@ import { FeatureFlagProvider } from '@aws-amplify/graphql-transformer-interfaces
 import { deploy, launchDDBLocal, terminateDDB, logDebug, GraphQLClient } from '../__e2e__/utils';
 import { AuthTransformer } from '@aws-amplify/graphql-auth-transformer';
 import { AmplifyAppSyncSimulator } from '@aws-amplify/amplify-appsync-simulator';
+import { pathManager } from "@aws-amplify/amplify-cli-core";
```
alharris-at, line 9: Why is this import required?

alharris-at (commented):
Couple minor comments/questions about imports and dead variable assignment.

packages/amplify-graphql-index-transformer/src/graphql-index-transformer.ts
dpilch: I'm not really sure why we have optional chaining for `stackManager`. It doesn't look like it does anything in the case anyway because we call `isProjectUsingDataStore` on `ctx` on line 70.

dpilch: Looking at the original commit, I don't see any reasoning for adding the optional chaining here. And it would still have the same issue as today with calling `isProjectUsingDataStore`.

https://github.com/aws-amplify/amplify-category-api/commit/5e59d18e3341b00bfdcc2e5edabf016fed8eb986

alharris-at (dismissed):

alharris-at (approved):

PR #1487, opened 2023-05-19
fix: disabling conflict detection

phani-srikar:
#### Description of changes

**Current Behavior:**
Customers have a deployed API backend with conflict detection turned ON. When they follow the CLI prompts for `update api` and turn it OFF, the subsequent push to cloud fails with the following error:
```
Invalid request provided: AttributeName property of TimeToLiveSpecification is required when TTL status is enabled or while enabling/disabling TTL.
```
This issue was reproduced at-least back until CLI v9.x. 

**Root Cause:**
The actual cause could be a breaking change DDB service introduced which is to **require** attribute name when disabling the TTL spec on a table.
We use a [DDB Table CDK construct](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_dynamodb.Table.html) which does not have the option to set the `enable` flag inside `timeToLiveSpecification` property of the table. CDK construct manages this itself. 

**Solution:**
We override the CfnTable low level construct that is obtained from CDK output to set the `timeToLiveSpecification` property of the table ourself. This property going forward will always exist but if it's `enabled` or not is decided by the conflict resolution configuration for the project.

We have removed the usage of `timeToLiveAttribute` input to the [DDB table construct](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_dynamodb.Table.html) and are now handling it via overriding the properties of added `CfnTable` in the stack. We use this strategy already to override some of the table properties like `sseSpecification` probably due to similar limitations of CDK construct.

Added unit test.
Running full e2e to make sure no regression is introduced.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-graphql-model-transformer/src/__tests__/model-transformer.test.ts
```diff
@@ -1255,6 +1255,26 @@ describe('ModelTransformer: ', () => {
         },
       });
   });
+
+  it('the datastore table TTL spec should not be removed but disabled when datastore is disabled', () => {
```
alharris-at, line 1259: Could you add a positive test case as well? I believe it's covered in snapshots, but would prefer to get towards these explicit tests.

alharris-at:
Could you please fill in the full template, especially `CDK / CloudFormation Parameters Changed`?

The two params here are marked as 'No Interruption', but I still want to ensure we understand the impact here.

A few questions:

1. If I'm already using a field called `_ttl` in my schema, that's a string, or other non-numeric type, is there an issue? (Is that even possible?)
2. I'm assuming you've toggled this forward/backward in testing, but we probably want an e2e test to validate that disabling works (to prevent regression).
3. Any other weird side effects you can think of from having this config unnecessarily on every table?

packages/amplify-graphql-model-transformer/src/__tests__/model-transformer.test.ts
phani-srikar: Added one

phani-srikar:
> 1. If I'm already using a field called `_ttl` in my schema, that's a string, or other non-numeric type, is there an issue? (Is that even possible?)

It is required that the field be a numeric type as per [this doc](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/time-to-live-ttl-before-you-start.html#time-to-live-ttl-before-you-start-formatting). However, if it isn't in valid format or doesn't have a valid value, DDB will ignore those items when scanning for expired items thereby retaining them in DB. 
If there is a field named `_ttl` already, DDB just uses it if conflict detection is enabled. If it is disabled, then this field is no different than any other. We use the underscore prefix to safeguard from clashing with custom fields.


> 2. I'm assuming you've toggled this forward/backward in testing, but we probably want an e2e test to validate that disabling works (to prevent regression).

I agree, but toggling this takes upto an hour as stated in DDB docs "Enabling, disabling, or changing TTL settings on a table can take approximately one hour for the settings to propagate and to allow the execution of any further TTL related actions.". It might be hard to maintain such tests. I did verify toggling this back and forth in a JS app and deploying.

> 3. Any other weird side effects you can think of from having this config unnecessarily on every table?

I haven't discovered any side effects yet. Only thing is that if people are migrating from older version, and they have DS disabled or not added, they'll still see these CFN changes.

packages/amplify-graphql-model-transformer/src/resources/dynamo-model-resource-generator.ts
```diff
@@ -171,6 +171,10 @@ export class DynamoModelResourceGenerator extends ModelResourceGenerator {
     cfnTable.sseSpecification = {
       sseEnabled: cdk.Fn.conditionIf(useSSE.logicalId, true, false),
     };
+    cfnTable.timeToLiveSpecification = {
```
AaronZyLee, line 174: For this specific line

AaronZyLee (approved):
Just one question for existing datastore projects. Will they have the same CFN output before and after this change?

alharris-at (approved):

packages/amplify-graphql-model-transformer/src/resources/dynamo-model-resource-generator.ts
phani-srikar: For customers using DataStore, they won't see any changes, but for customers not using DataStore or those who disable it, there will be this additional block with `enabled: false`.

AaronZyLee (approved):

alharris-at:
Closing this PR now that DDB is mitigating impact in-service.

PR #1509, opened 2023-05-24
chore: bump all specific node references to 18

alharris-at:
#### Description of changes
bump all specific node references to 18

Yes - this changes runtime config for predictions lambda from default 14 to 18 and updates the template for sync conflict resolution in a lambda from 14 to 18.

#### Issue #, if available
N/A

E2E Tests + Unit Tests

#### Checklist
- [x] PR description included
- [x] `yarn test` passes


AaronZyLee (dismissed):

sobolk (approved):

packages/graphql-transformer-common/src/PredictionsResourceIDs.ts
```diff
@@ -5,7 +5,7 @@ export class PredictionsResourceIDs {
   static lambdaName = 'predictionsLambda';
   static lambdaID = 'predictionsLambdaFunction';
   static lambdaHandlerName = 'predictionsLambda.handler';
-  static lambdaRuntime = 'nodejs14.x';
+  static lambdaRuntime = 'nodejs18.x';
```
sobolk, line 8: one thing to double check - does this or conflict handler rely on AWS SDK. If yes then code should be adapted to use V3 instead of V2.

I skimmed both functions and they don't seem to rely on SDK, so we should be fine. But please double check that.

dpilch (approved):

package.json
```diff
@@ -4,7 +4,7 @@
   "description": "amplify-category-api",
   "scripts": {
     "test-changed": "lerna run test --since main",
-    "test": "lerna run test",
+    "test": "lerna run test -- -u",
```
dpilch, line 7: Accidental commit?

alharris-at: Oh yeah, just committed that since yarn test -- -u wasn't working from root. I'll revert (though I expect to see some changes required in E2E tests from the run, so will wait til those have succeeded/failed to get more data)

phani-srikar: Think it's just 'yarn test -u' without the additional dashes

PR #1526, opened 2023-05-31
chore: move fs access for historic search config to api category

alharris-at:
#### Description of changes
move fs access for historic search config to api category.

Since previous state nodeToNodeEncryption is an input concern, as a result we should bound this concern into the api category.

N/A - no functional changes, we should still respect override, then historic context, then default to false for this resource.

#### Issue #, if available
N/A

Unit + e2e tests

#### Checklist
- [x] PR description included
- [x] `yarn test` passes


packages/amplify-graphql-searchable-transformer/src/graphql-searchable-transformer.ts
```diff
@@ -299,14 +302,7 @@ export class SearchableModelTransformer extends TransformerPluginBase {
...
 
-    const nodeToNodeEncryption = this.apiName
-      ? shouldEnableNodeToNodeEncryption(this.apiName, context.filepaths.findProjectRoot(), context.filepaths.getCurrentCloudBackendDirPath())
-      : { enabled: false, log: undefined };
-    if (nodeToNodeEncryption.log) {
-      this.warn(nodeToNodeEncryption.log);
-    }
-
-    const domain = createSearchableDomain(stack, parameterMap, context.api.apiId, nodeToNodeEncryption.enabled);
+    const domain = createSearchableDomain(stack, parameterMap, context.api.apiId, this.options?.enableNodeToNodeEncryption ?? false);
```
sundersc, line 305: Is the default value set to 'false' to handle apps created before nodeToNodeEncryption change?

AaronZyLee (approved):

packages/amplify-graphql-searchable-transformer/src/graphql-searchable-transformer.ts
alharris-at: It's set to false because we want to ensure that customers ONLY enable nodeToNode if it was already enabled previously because disabling it is lossy, we don't want to toggle it on for customers accidentally.

So the behavior (preserved from existing) is:

If override is set (e.g. by manual action), then respect override.
If existingBackendStack contains value, respect existing value.
Else: false.

alharris-at:
Closing this in favor of https://github.com/aws-amplify/amplify-category-api/pull/1527.

PR #1531, opened 2023-06-02
chore: remove rds related commands

phani-srikar:
#### Description of changes

Treat RDS commands as any normal unsupported command.
After this change it'll show as:
image

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-category-api/amplify-plugin.json
+++ b/packages/amplify-category-api/amplify-plugin.json
@@ -12,10 +12,7 @@
     "rebuild",
     "remove",
     "update",
-    "help",
-    "import",
-    "generate-schema",
-    "update-secrets"
+    "help"
   ],
   "commandAliases": {
     "configure": "update"
--- a/packages/amplify-category-api/src/commands/api.ts
+++ b/packages/amplify-category-api/src/commands/api.ts
@@ -55,18 +55,6 @@ export const run = async (context: $TSContext) => {
     {
       name: 'override',
       description: 'Generates overrides file to apply custom modifications to CloudFormation',
-    },
-    {
-      name: 'import',
-      description: 'Imports existing datasource to GraphQL API',
-    },
-    {
-      name: 'generate-schema',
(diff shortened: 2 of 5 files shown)


alharris-at:
Could you also remove the explicit errors that get thrown as part of the commands, since they're now fully removed from execution?

AaronZyLee (approved):

packages/amplify-category-api/amplify-plugin.json
```diff
@@ -12,10 +12,7 @@
     "rebuild",
     "remove",
     "update",
-    "help",
-    "import",
-    "generate-schema",
-    "update-secrets"
```
sundersc, line 18: Can we comment these lines with note so it is easy to enable when required?

packages/amplify-category-api/src/commands/api.ts
```diff
@@ -55,18 +55,6 @@ export const run = async (context: $TSContext) => {
...
-      name: 'import',
-      description: 'Imports existing datasource to GraphQL API',
-    },
-    {
-      name: 'generate-schema',
-      description: 'Generates the GraphQL schema from the Data Source',
-    },
-    {
-      name: 'update-secrets',
-      description: 'Updates the API plugin related secrets',
```
sundersc, line 69: Can we comment these lines with note so it is easy to enable when required?

packages/amplify-category-api/amplify-plugin.json
phani-srikar: Can't do that here since it's a JSON.

packages/amplify-category-api/src/commands/api.ts
phani-srikar: We probably need these reviewed before release since I see some scope for improvement here. Can we re-visit them when these are ready?

alharris-at (approved):

PR #1532, opened 2023-06-02
fix(graphql): renamed subscription should generate auth resolver

sundersc:
#### Description of changes

While renaming subscriptions, transformers do not generate auth resolvers for the operation. This PR fixes the issue.

**For example:** For the below schema, currently no auth resolvers are generated for `onNoteUpdate` subscription.

```graphql
type Note 
  @model(subscriptions: { onUpdate: ["onNoteUpdate"] }) 
  @auth(rules: [{ allow: owner }]) 
{
  noteId: String! @primaryKey
  comments: String
}
```

NA

#### Issue #, if available

Internal ticket

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-graphql-auth-transformer/src/utils/schema.ts
+++ b/packages/amplify-graphql-auth-transformer/src/utils/schema.ts
@@ -345,32 +345,35 @@ export const getSubscriptionFieldNames = (
     typeName: string;
   }> = new Set();
 
-  if (modelDirectiveConfig?.subscriptions?.level === SubscriptionLevel.on) {
-    if (modelDirectiveConfig?.subscriptions?.onCreate && modelDirectiveConfig.mutations?.create) {
-      for (const fieldName of modelDirectiveConfig.subscriptions.onCreate) {
-        fields.add({
-          typeName: 'Subscription',
-          fieldName,
-        });
-      }
-    }
+  const subscriptionLevel = modelDirectiveConfig?.subscriptions?.level ?? SubscriptionLevel.on;
+  if (subscriptionLevel !== SubscriptionLevel.on) {
+    return fields;
+  }
 
-    if (modelDirectiveConfig?.subscriptions?.onUpdate && modelDirectiveConfig.mutations?.update) {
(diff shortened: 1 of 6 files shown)


packages/amplify-graphql-auth-transformer/src/utils/schema.ts
```diff
@@ -345,32 +345,35 @@ export const getSubscriptionFieldNames = (
...
-        });
-      }
+  const subscriptionLevel = modelDirectiveConfig?.subscriptions?.level ?? SubscriptionLevel.on;
+  if (subscriptionLevel !== SubscriptionLevel.on) {
+    return fields;
+  }
+
+  if (modelDirectiveConfig?.subscriptions?.onCreate && modelDirectiveConfig.mutations?.create) {
+    for (const fieldName of modelDirectiveConfig.subscriptions.onCreate) {
+      fields.add({
```
phani-srikar, line 355: I see these fields are just used within the Auth transformer and we are not mutating any final compiled schema. Is that taken care of in the model transformer where we add these custom names in the compiled schema? An e2e test would be nice to have to verify we handle custom sub field names correctly, if we don't have it already.

phani-srikar (changes requested):

packages/amplify-graphql-auth-transformer/src/utils/schema.ts
dpilch: This block doesn't look like it is changing from the current functionality. If so it should be fine to get the fix out without adding the tests.

alharris-at (approved):

packages/amplify-graphql-auth-transformer/src/utils/schema.ts
sundersc: This is no change in behavior here, just added an IF block to early return and avoid nesting IF blocks. The newly added unit tests confirm the resolvers are assigned to the custom operation name based on the file name. The existing tests makes sure that there is no regression.

phani-srikar (approved):

dpilch (approved):

PR #1541, opened 2023-06-06
Add credential refresh logic for long running tests

phani-srikar:
#### Description of changes

- Add refresh credentials for some Searchable tests that could take more than an hour to run. This eliminates the token expiry due to role chaining.
- Use `amplify-integ-test-user` profile to collect flaky test metrics. This step used to fail if job runs for more than an hour.
- Updates to codebuild split-e2e script to use parent account for searchable tests.

Verified that `searchable-migration` test runs beyond an hour and passes.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-e2e-core/package.json
+++ b/packages/amplify-e2e-core/package.json
@@ -39,7 +39,9 @@
     "semver": "^7.3.5",
     "strip-ansi": "^6.0.0",
     "throat": "^5.0.0",
-    "uuid": "^8.3.2"
+    "uuid": "^8.3.2",
+    "ini": "^3.0.1",
+    "@types/ini": "^1.3.31"
   },
   "peerDependencies": {
     "@aws-amplify/amplify-cli-core": "4.0.4"
--- a/packages/amplify-e2e-core/package.json
+++ b/packages/amplify-e2e-core/package.json
@@ -40,7 +40,9 @@
     "strip-ansi": "^6.0.0",
     "throat": "^5.0.0",
     "uuid": "^8.3.2",
-    "ini": "^3.0.1",
+    "ini": "^3.0.1"
+  },
+  "devDependencies": {
     "@types/ini": "^1.3.31"
   },
   "peerDependencies": {
--- a/packages/amplify-e2e-tests/refresh-credentials.sh
+++ b/packages/amplify-e2e-tests/refresh-credentials.sh
@@ -1,0 +1,21 @@
+#!/bin/sh
+
+# set exit on error to true
+set -e
+
+# Unassume the current role
+unset AWS_ACCESS_KEY_ID
+unset AWS_SECRET_ACCESS_KEY
+unset AWS_SESSION_TOKEN
+
+# Reassume the e2e test account role
+creds=$(aws sts assume-role --role-arn $TEST_ACCOUNT_ROLE --role-session-name testSession1248634 --duration-seconds 3600)
+if [ -z $(echo $creds | jq -c -r '.AssumedRoleUser.Arn') ]; then
+    return
+fi
+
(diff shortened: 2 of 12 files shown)


packages/amplify-e2e-tests/refresh-credentials.sh
```diff
@@ -0,0 +1,21 @@
...
+creds=$(aws sts assume-role --role-arn $TEST_ACCOUNT_ROLE --role-session-name testSession1248634 --duration-seconds 3600)
+if [ -z $(echo $creds | jq -c -r '.AssumedRoleUser.Arn') ]; then
+    return
+fi
+
+export AWS_ACCESS_KEY_ID=$(echo $creds | jq -c -r ".Credentials.AccessKeyId")
+export AWS_SECRET_ACCESS_KEY=$(echo $creds | jq -c -r ".Credentials.SecretAccessKey")
+export AWS_SESSION_TOKEN=$(echo $creds | jq -c -r ".Credentials.SessionToken")
+
+echo "{ \"accessKeyId\": \"$AWS_ACCESS_KEY_ID\", \"secretAccessKey\": \"$AWS_SECRET_ACCESS_KEY\", \"sessionToken\": \"$AWS_SESSION_TOKEN\" }"
```
AaronZyLee, line 21: Is the echo here used for the shell return value? Will this echo get exposed in the console?

packages/amplify-e2e-core/package.json
```diff
@@ -39,7 +39,9 @@
     "semver": "^7.3.5",
     "strip-ansi": "^6.0.0",
     "throat": "^5.0.0",
-    "uuid": "^8.3.2"
+    "uuid": "^8.3.2",
+    "ini": "^3.0.1",
+    "@types/ini": "^1.3.31"
```
alharris-at, line 44: should be a dev dependency?

packages/amplify-e2e-tests/src/__tests__/graphql-v2/searchable-datastore.test.ts
```diff
@@ -27,7 +31,8 @@ describe('transformer model searchable migration test', () => {
   });
 
   afterEach(async () => {
-    await deleteProject(projRoot);
+    const newCreds = refreshCredentials();
```
AaronZyLee, line 34: Can we add a codebuild ci check here? The credential refresh is only needed in CodeBuild CI and not needed if running locally or in CircleCI. The current env variables used are `process.env.CI && process.env.CODEBUILD`

AaronZyLee (changes requested):

packages/amplify-e2e-tests/src/__tests__/graphql-v2/searchable-datastore.test.ts
phani-srikar: Added an early return inside the refresh method, that should take care of this.

packages/amplify-e2e-core/package.json
phani-srikar: moved the types to dev-dep. Thanks.

packages/amplify-e2e-tests/refresh-credentials.sh
phani-srikar: Yes. Not unless we run this script outside of a child process. Hence I put this inside the `e2e-tests` package different from the other `shared-scripts.sh` that we run directly from the shell.

AaronZyLee (approved):

alharris-at (approved):

packages/amplify-e2e-tests/refresh-credentials.sh
sobolk, line 21: why not https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide/javascript_sts_code_examples.html ?

PR #1553, opened 2023-06-08
feature(graphql): modify sql lambda to use public layer

sundersc:
#### Description of changes

- Modify SQL lambda to use the beta public layers.
- Adds E2E tests for happy cases on models with default primary key and composite key.

NA

#### Issue #, if available

NA

- Manual test
- Added new E2E tests

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-graphql-model-transformer/rds-lambda/handler.ts
```diff
@@ -1,6 +1,49 @@
...
+  }
+  const parameterCommand = new GetParameterCommand({
+    Name: key,
+    WithDecryption: true,
+  });
+  const data = await secretsClient.send(parameterCommand);
+  if ((data.$metadata?.httpStatusCode && data?.$metadata?.httpStatusCode >= 400) || !data.Parameter?.Value) {
+    throw new Error('Unable to get secret for database connection');
+  }
+  return data.Parameter?.Value ?? '';
```
alharris-at, line 33: Why an empty string by default? Isn't that an error case as well?

alharris-at (approved):

packages/amplify-graphql-model-transformer/rds-lambda/handler.ts
sundersc: I see this logic is carried forward from existing code. We can directly return `data.Parameter?.value`, I will test and add it in a follow up PR.

packages/amplify-graphql-model-transformer/src/resolvers/rds/resolver.ts
```diff
@@ -31,15 +25,94 @@ import {
...
 
 const OPERATION_KEY = '__operation';
 
+const RDSLayerMappingID = 'RDSLayerResourceMapping';
+// TODO: This is temporary state, we need to modify this to a production layer
+export const setRDSLayerMappings = (scope: Construct): CfnMapping => new CfnMapping(
+  scope,
+  RDSLayerMappingID,
+  {
+    mapping: {
```
phani-srikar, line 42: can we generate this mapping via a method call so that it's easier to make changes and maintain? The method can take an array of supported region constant, account id and layerName.

packages/amplify-graphql-model-transformer/rds-lambda/handler.ts
```diff
@@ -1,6 +1,49 @@
...
+}
+
+const getDBConfig = async(): DBConfig  => {
+  if (!secretsClient) {
```
phani-srikar, line 37: This method is doing 2 things: creating SSM client and getting DB config. Moving the SSM client creation part into `getSSMValue` where it is actually required does more justice to this method.

packages/amplify-graphql-model-transformer/rds-lambda/handler.ts
```diff
@@ -1,6 +1,49 @@
...
+    createSSMClient();
+  }
+
+  return {
+    engine: 'mysql',
+    host: await getSSMValue(process.env.host),
+    port: Number.parseInt(await getSSMValue(process.env.port)) || 3306,
+    username: await getSSMValue(process.env.username),
+    password: await getSSMValue(process.env.password),
+    database: await getSSMValue(process.env.database),
```
phani-srikar, line 47: Are these env variables guaranteed to exist?

phani-srikar (changes requested):

packages/amplify-graphql-model-transformer/rds-lambda/handler.ts
sundersc: It must exist. Otherwise it will throw connection refused exception.

packages/amplify-graphql-model-transformer/rds-lambda/handler.ts
sundersc: The client is intentionally kept outside the scope of this method so that it can be reused on subsequent invocations.

packages/amplify-graphql-model-transformer/src/resolvers/rds/resolver.ts
sundersc: The version and account # may be different between regions, I don't see much benefit in moving this to an utility method. Can we revisit when we migrate to production account?

packages/amplify-graphql-model-transformer/rds-lambda/handler.ts
sundersc: Added check to throw an error if any of the config doesn't exist.

packages/amplify-graphql-model-transformer/rds-lambda/handler.ts
sundersc: Removed the default empty string and added a statement to throw exception if any of the config doesn't exist.

packages/amplify-graphql-model-transformer/rds-lambda/handler.ts
phani-srikar: If you meant subsequent lambda invocations, that is something new to me that we can share local variables across invocations.

phani-srikar (approved):

PR #1562, opened 2023-06-09
chore: refactor feature flags out of transformer layer, only use params

alharris-at:
#### Description of changes
Remove FeatureFlagProvider as a concept from the GraphQL V2 transformer dependency chain. Instead constructing the transformation context will produce a strongly typed config object, where previous feature flag values were stored. This is functionally equivalent to using a feature flag provider, but has a couple of nice side effects:

1. Concept count within the Transformer reduces by 1, I no longer need to think about requesting a parameter from the feature flag provider, and passing that provider around. Users of the transformer do not need to construct a feature flag provider to use it.
2. Config becomes strongly typed, because the config object is just a normal type, it's easy to reason about the possible values.
3. Consumers of feature flags cannot set different default behaviors, all behavior is fixed at the invocation of the transform.

The result of this though is that basically all tests which constructed a v2 transform needed to be updated, and any and all code accessing params will need to change as well.

__N.B. This should result in a MV bump for the transformer layers, since it's a fundamental breaking change in the Transformer API.__

N/A

#### Issue #, if available
N/A

Updated Unit + [E2E Tests](https://app.circleci.com/pipelines/github/aws-amplify/amplify-category-api/7199/workflows/c2b282b2-f828-466a-ab6d-3da64cab7f83)

#### Checklist
- [x] PR description included
- [x] `yarn test` passes


dpilch (dismissed):
LGTM

`packages/amplify-category-api/src/__tests__/graphql-transformer/override/amplify-graphql-index-transformer-override.test.ts` and `packages/amplify-category-api/src/__tests__/graphql-transformer/override/amplify-graphql-primary-key-transformer-override.test.ts` will need to be updated from my latest change on main.

AaronZyLee (commented):
I think some changes are also needed in the CLI repo if we are going to decouple the feature flag instance. One place I notice is [the test suites](https://github.com/aws-amplify/amplify-cli/blob/dev/packages/amplify-util-mock/src/__tests__/velocity/relational-auth.test.ts#L72) in `amplify-util-mock` of CLI. Could you double check if there are other similar changes required for the CLI repo as well before this change is merged?

packages/amplify-graphql-relational-transformer/src/graphql-many-to-many-transformer.ts
```diff
@@ -89,13 +94,16 @@ export class ManyToManyTransformer extends TransformerPluginBase {
...
-    } as ManyToManyDirectiveConfiguration, { deepMergeArguments: context.featureFlags.getBoolean('shouldDeepMergeDirectiveConfigDefaults', false) });
+    const args = directiveWrapped.getArguments(
+      {
+        directiveName,
+        object: parent as ObjectTypeDefinitionNode,
+        field: definition,
+        directive,
+        limit: defaultLimit,
+      } as ManyToManyDirectiveConfiguration,
+      generateGetArgumentsInput(context.transformParameters),
```
phani-srikar, line 105: qq: why do we need this extra call to `generateGetArgumentsInput` here?

packages/amplify-util-mock/src/__e2e_v2__/index-transformer.e2e.test.ts
```diff
@@ -63,9 +62,6 @@ describe('@index transformer', () => {
 
     const transformer = new GraphQLTransform({
       transformers: [new ModelTransformer(), new PrimaryKeyTransformer(), new IndexTransformer(), new AuthTransformer()],
-      featureFlags: {
-        getBoolean: name => (name === 'improvePluralization' ? true : false),
-      } as FeatureFlagProvider,
     });
```
phani-srikar, line 65: don't we need to set other params except `improvePluralization` to `false` here?

alharris-at: In general I started by removing any flags that were here, since we're probably not testing that behavior, and only patching as tests failed. This way we're not needlessly preserving old test state.

alharris-at:
@AaronZyLee  - E2Es for CLI here - looking all good https://app.circleci.com/pipelines/github/aws-amplify/amplify-cli/21095/workflows/c49e1ee0-81ae-48a9-8f43-590cef4f87aa

phani-srikar (approved):

AaronZyLee (approved):

PR #1581, opened 2023-06-17
feat(graphql): patching rds lambda layer

sundersc:
#### Description of changes

- Adds patching mechanism for SQL lambda.
- Modifies RDS lambda runtime from Node 16.x to 18.x.
- Fixes a bug related to `input AMPLIFY` object. Now there is no need to have schema.graphql file to enable sandbox mode.

NA

#### Issue #, if available

NA

- Manual test
- E2E tests updated

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-e2e-tests/src/__tests__/rds-import-vpc.test.ts
+++ b/packages/amplify-e2e-tests/src/__tests__/rds-import-vpc.test.ts
@@ -19,6 +19,13 @@ import gql from 'graphql-tag';
 import path from 'path';
 import { print } from 'graphql';
 
+const CDK_FUNCTION_TYPE = 'AWS::Lambda::Function';
+const CDK_SUBSCRIPTION_TYPE = 'AWS::SNS::Subscription';
+const APPSYNC_DATA_SOURCE_TYPE = 'AWS::AppSync::DataSource';
+
+const SNS_TOPIC_REGION = 'us-east-1';
+const SNS_TOPIC_ARN = 'arn:aws:sns:us-east-1:956468067974:AmplifyRDSLayerNotification';
+
 describe("RDS Tests", () => {
   const [db_user, db_password, db_identifier] = generator.generateMultiple(3);
   
@@ -70,7 +77,7 @@ describe("RDS Tests", () => {
     await deleteDBInstance(identifier, region);
   };
 
-  it("import workflow of mysql relational database with public access", async () => {
+  it("import workflow of mysql relational database within vpc with no public access", async () => {
     const apiName = 'rdsapivpc';
     await initJSProjectWithProfile(projRoot, {
       disableAmplifyAppCreation: false,
@@ -120,15 +127,17 @@ describe("RDS Tests", () => {
     updateSchema(projRoot, apiName, print(updatedSchema), 'schema.rds.graphql');
     await apiGqlCompile(projRoot);
 
-    // Validate if the SQL lambda function has VPC configuration
+    // Validate the generated resources in the CloudFormation template
     const apisDirectory = path.join(projRoot, 'amplify', 'backend', 'api');
     const apiDirectory = path.join(apisDirectory, apiName);
     const cfnRDSTemplateFile = path.join(apiDirectory, 'build', 'stacks', `RdsApiStack.json`);
     const cfnTemplate = JSON.parse(readFileSync(cfnRDSTemplateFile, 'utf8'));
     console.log(JSON.stringify(cfnTemplate, null, 4));
     expect(cfnTemplate.Resources).toBeDefined();
-    const resources = Object.values(cfnTemplate.Resources);
-    const rdsLambdaFunction = resources.find((r: any) => r.Type === 'AWS::Lambda::Function') as any;
+    const resources = cfnTemplate.Resources;
+
+    // Validate if the SQL lambda function has VPC configuration
+    const rdsLambdaFunction = getResource(resources, 'RDSLambdaLogicalID', CDK_FUNCTION_TYPE);
     expect(rdsLambdaFunction).toBeDefined();
     expect(rdsLambdaFunction.Properties).toBeDefined();
     expect(rdsLambdaFunction.Properties.VpcConfig).toBeDefined();
@@ -137,6 +146,39 @@ describe("RDS Tests", () => {
     expect(rdsLambdaFunction.Properties.VpcConfig.SecurityGroupIds).toBeDefined();
     expect(rdsLambdaFunction.Properties.VpcConfig.SecurityGroupIds.length).toBeGreaterThan(0);
 
+    // Validate patching lambda and subscription
+    const rdsPatchingLambdaFunction = getResource(resources, 'RDSPatchingLambdaLogicalID', CDK_FUNCTION_TYPE);
+    expect(rdsPatchingLambdaFunction).toBeDefined();
+    expect(rdsPatchingLambdaFunction.Properties).toBeDefined();
+    expect(rdsPatchingLambdaFunction.Properties.Environment).toBeDefined();
+    expect(rdsPatchingLambdaFunction.Properties.Environment.Variables).toBeDefined();
+    expect(rdsPatchingLambdaFunction.Properties.Environment.Variables.LAMBDA_FUNCTION_ARN).toBeDefined();
+    const rdsDataSourceLambda = getResource(resources, 'RDSLambdaDataSource', APPSYNC_DATA_SOURCE_TYPE);
+    expect(rdsPatchingLambdaFunction.Properties.Environment.Variables.LAMBDA_FUNCTION_ARN).toEqual(rdsDataSourceLambda.Properties.LambdaConfig.LambdaFunctionArn);
+
+    // Validate subscription
+    const rdsPatchingSubscription = getResource(resources, 'RDSPatchingLambdaLogicalID', CDK_SUBSCRIPTION_TYPE);
+    expect(rdsPatchingSubscription).toBeDefined();
+    expect(rdsPatchingSubscription.Properties).toBeDefined();
+    expect(rdsPatchingSubscription.Properties.Protocol).toBeDefined();
+    expect(rdsPatchingSubscription.Properties.Protocol).toEqual('lambda');
+    expect(rdsPatchingSubscription.Properties.Endpoint).toBeDefined();
+    expect(rdsPatchingSubscription.Properties.TopicArn).toBeDefined();
+    expect(rdsPatchingSubscription.Properties.TopicArn).toEqual(SNS_TOPIC_ARN);
+    expect(rdsPatchingSubscription.Properties.Region).toEqual(SNS_TOPIC_REGION);
+    expect(rdsPatchingSubscription.Properties.FilterPolicy).toBeDefined();
+    expect(rdsPatchingSubscription.Properties.FilterPolicy.Region).toBeDefined();
+
     await amplifyPush(projRoot);
   });
 }); 
+
+const getResource = (resources: Map<string, any>, resourcePrefix: string, resourceType: string): any => {
+  const keys = Array.from(Object.keys(resources)).filter(key => key.startsWith(resourcePrefix));
+  for (const key of keys) {
+    const resource = resources[key];
+    if (resource.Type === resourceType) {
+      return resource;
+    }
+  }
+};
--- a/packages/amplify-e2e-tests/src/__tests__/rds-import-vpc.test.ts
+++ b/packages/amplify-e2e-tests/src/__tests__/rds-import-vpc.test.ts
@@ -113,7 +113,7 @@ describe("RDS Tests", () => {
     expect(dbObjectType.directives.find(d => d.name.value === 'model')).toBeDefined();
 
     const updatedSchema = gql`
-      input Amplify {
+      input AMPLIFY {
         engine: String = "mysql"
         globalAuthRule: AuthRule = {allow: public}
       }
--- a/packages/amplify-graphql-model-transformer/package.json
+++ b/packages/amplify-graphql-model-transformer/package.json
@@ -21,9 +21,12 @@
     "access": "public"
   },
   "scripts": {
-    "build": "tsc && cd rds-lambda && mkdir -p node_modules && rm -rf node_modules && npm install && tsc && cp -r node_modules lib && cd lib && bestzip --force node ../../lib/rds-lambda.zip ./*",
+    "build-rds-lambda": "cd rds-lambda && mkdir -p node_modules && rm -rf node_modules && npm install && tsc && cp -r node_modules lib && cd lib && bestzip --force node ../../lib/rds-lambda.zip ./* && cd ../..",
+    "build-rds-patching-lambda": "cd rds-patching-lambda && mkdir -p node_modules && rm -rf node_modules && npm install && tsc && cp -r node_modules lib && cd lib && bestzip --force node ../../lib/rds-patching-lambda.zip ./* && cd ../..",
+    "build-notification-lambda": "cd publish-notification-lambda && mkdir -p node_modules && rm -rf node_modules && npm install && tsc && cp -r node_modules lib && cd lib && bestzip --force node ../../lib/rds-notification-lambda.zip ./* && cd ../..",
+    "build": "tsc && yarn build-rds-lambda && yarn build-rds-patching-lambda && yarn build-notification-lambda",
     "watch": "tsc -w",
-    "clean": "rimraf ./lib",
+    "clean": "rimraf ./lib && rimraf ./rds-lambda/lib && rimraf ./rds-patching-lambda/lib && rimraf ./publish-notification-lambda/lib",
     "test": "jest",
     "test-watch": "jest --watch",
     "extract-api": "ts-node ../../scripts/extract-api.ts"
--- a/packages/amplify-graphql-model-transformer/publish-notification-lambda/package.json
+++ b/packages/amplify-graphql-model-transformer/publish-notification-lambda/package.json
@@ -1,0 +1,23 @@
+{
+  "name": "Publish Notification Message",
+  "version": "1.0.0",
+  "description": "",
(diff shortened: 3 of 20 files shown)


packages/amplify-graphql-model-transformer/package.json
```diff
@@ -21,9 +21,12 @@
     "access": "public"
   },
   "scripts": {
-    "build": "tsc && cd rds-lambda && mkdir -p node_modules && rm -rf node_modules && npm install && tsc && cp -r node_modules lib && cd lib && bestzip --force node ../../lib/rds-lambda.zip ./*",
+    "build-rds-lambda": "cd rds-lambda && mkdir -p node_modules && rm -rf node_modules && npm install && tsc && cp -r node_modules lib && cd lib && bestzip --force node ../../lib/rds-lambda.zip ./* && cd ../..",
+    "build-rds-patching-lambda": "cd rds-patching-lambda && mkdir -p node_modules && rm -rf node_modules && npm install && tsc && cp -r node_modules lib && cd lib && bestzip --force node ../../lib/rds-patching-lambda.zip ./* && cd ../..",
+    "build-notification-lambda": "cd publish-notification-lambda && mkdir -p node_modules && rm -rf node_modules && npm install && tsc && cp -r node_modules lib && cd lib && bestzip --force node ../../lib/rds-notification-lambda.zip ./* && cd ../..",
```
alharris-at, line 26: Could we use `rimraf` and `mkdirp` to better support windows platforms? If there's a lib to replace `cp -r` as well, that'd be preferable I think, but I won't block on that for now.

packages/amplify-graphql-model-transformer/publish-notification-lambda/src/index.ts
```diff
@@ -0,0 +1,100 @@
...
+  'ca-central-1',
+  'ap-northeast-3',
+];
+
+type LayerConfig = {
+  layerArn: string | undefined;
+  region: string;
+};
+
+const getLatestVersion = async (layerName: string, region: string): Promise<LayerConfig> => {
```
alharris-at, line 32: Missing comment.

packages/amplify-graphql-model-transformer/publish-notification-lambda/src/index.ts
```diff
@@ -0,0 +1,100 @@
...
+  const client = new LambdaClient({ region });
+  const versions = [];
+  let response: ListLayerVersionsCommandOutput | undefined;
+  do {
+    const command = new ListLayerVersionsCommand({
+      LayerName: layerName,
+      MaxItems: MAX_ITEMS,
+      Marker: response?.NextMarker,
+    });
```
alharris-at, line 41: Is there any way we can perform a sort here so we don't need to page? (Or can page until we find a first element or something)? I'm worried about this growing over time, and theoretically timing out the querying lambda (unlikely, but we should avoid continually expanding the runtime here if possible)

packages/amplify-graphql-model-transformer/publish-notification-lambda/src/index.ts
```diff
@@ -0,0 +1,100 @@
...
+    // eslint-disable-next-line no-await-in-loop
+    response = await client.send(command);
+    if (response.LayerVersions) {
+      versions.push(...response.LayerVersions);
+    }
+  } while (response?.NextMarker);
+  const result = versions.sort((a, b) => b.Version! - a.Version!)[0];
+  return {
+    layerArn: result.LayerVersionArn,
+    region,
```
alharris-at, line 51: result is possibly an empty array, right? should layerArn be `result?.LayerVersionArn`?

packages/amplify-graphql-model-transformer/publish-notification-lambda/src/index.ts
```diff
@@ -0,0 +1,100 @@
...
+  };
+};
+
+const getLatestVersions = async (): Promise<LayerConfig[]> => {
+  const promises = SUPPORTED_REGIONS.map((region) => getLatestVersion(LAYER_NAME!, region));
+  return Promise.all(promises);
+};
```
alharris-at, line 58: Do we need to query all regions? You know the region you're deployed to, right?

packages/amplify-graphql-model-transformer/rds-patching-lambda/package.json
```diff
@@ -0,0 +1,22 @@
+{
+  "name": "rdslayerstresstest",
+  "version": "1.0.0",
+  "description": "",
+  "main": "lib/index.js",
+  "scripts": {
+    "test": "npm run build && node ./lib/index.js",
+    "build": "tsc",
+    "release": "rm -f test-lambda.zip && bestzip --force node ./test-lambda.zip ./*"
```
alharris-at, line 9: Same comment on devdeps

packages/amplify-graphql-model-transformer/rds-patching-lambda/src/index.ts
```diff
@@ -0,0 +1,61 @@
+import { LambdaClient, UpdateFunctionConfigurationCommand, UpdateFunctionConfigurationCommandOutput } from "@aws-sdk/client-lambda";
+
+const snsEventSource = 'aws:sns';
+
+type LayerConfig = {
+  layerArn: string;
+  region: string;
+};
```
alharris-at, line 8: At least this type is shared across your two lambdas, if there's sizable shared resources/shapes/utils I'd consider moving these to a single bundle w/ multiple entry points.

packages/amplify-graphql-model-transformer/rds-patching-lambda/src/index.ts
```diff
@@ -0,0 +1,61 @@
...
+
+const getLayerConfig = (event: any): LayerConfig => {
```
alharris-at, line 10: There should be an `SNSEvent` type we can use here.

packages/amplify-graphql-model-transformer/rds-patching-lambda/src/index.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  const { layerArn, region } = getLayerConfig(event);
+  if (!layerArn || !region) {
+    throw new Error('Layer ARN or region is missing in the notification event');
+  }
+
+  if (region !== process.env.AWS_REGION) {
+    console.log(`Region ${region} in notification is not same as the current region ${process.env.AWS_REGION}. Skipping update.`);
+    return;
+  }
```
alharris-at, line 51: When would this happen?

packages/amplify-graphql-model-transformer/src/resolvers/rds/resolver.ts
```diff
@@ -160,14 +160,43 @@ export const createRdsLambda = (
...
+  return apiGraphql.host.addLambdaFunction(
+    RDSPatchingLambdaLogicalID,
+    `functions/${RDSPatchingLambdaLogicalID}.zip`,
+    'index.handler',
+    path.resolve(__dirname, '..', '..', '..', 'lib', 'rds-patching-lambda.zip'),
+    Runtime.NODEJS_18_X,
+    [],
+    lambdaRole,
+    environment,
+    Duration.seconds(60),
```
alharris-at, line 186: How long is this taking in your testing now?

packages/graphql-transformer-core/src/util/transformConfig.ts
```diff
@@ -294,7 +309,34 @@ export async function readSchema(projectDirectory: string): Promise<{schema: str
     schema,
     modelToDatasourceMap,
   };
-}
+};
+
+const removeAmplifyInput = (schema: string): SchemaReaderConfig => {
```
alharris-at, line 314: Don't we already have this utility somewhere? Did I miss it getting moved from somewhere else.

alharris-at (commented):
Some comments, mostly on the lambdas

packages/amplify-graphql-model-transformer/rds-patching-lambda/src/index.ts
sundersc: This is an additional safety check just in case if someone is invoking the lambda outside the SNS subscription.

packages/amplify-graphql-model-transformer/src/resolvers/rds/resolver.ts
sundersc: About 5-10 seconds. I'm going to add a retry mechanism in a follow-up PR. In that case, we will make first retry after 5 seconds and second retry after 25 seconds. Considering that setting the timeout to 60s.

packages/graphql-transformer-core/src/util/transformConfig.ts
sundersc: I couldn't find it. For the 'AMPLIFY' input object in schema.graphql file, we skip if while processing the AST. (Code reference link below).
https://github.com/aws-amplify/amplify-category-api/blob/9a4f6203095d87ce553e4cc6a4c9210fcc5b2fe0/packages/amplify-graphql-transformer-core/src/transformer-context/output.ts#L77

packages/amplify-graphql-model-transformer/rds-patching-lambda/src/index.ts
sundersc: I intentionally avoided it so that the lambda can be invoked without the full event object. Still thought about creating minimum type definition for an event but the 'MessageAttributes' aren't strongly typed. But we can still do better than 'any', I will take of it in a follow up PR.

packages/amplify-graphql-model-transformer/rds-patching-lambda/src/index.ts
```diff
@@ -0,0 +1,61 @@
...
+    throw new Error('No SNS notification found in the event');
+  }
+
+  const layerArn = Sns?.MessageAttributes?.LayerArn?.Value;
+  const region = Sns?.MessageAttributes?.Region?.Value;
+
+  return {
+    layerArn,
+    region,
+  };
```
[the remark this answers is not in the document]
sundersc: Fixed this.

packages/amplify-graphql-model-transformer/publish-notification-lambda/src/index.ts
sundersc: It returned the latest version as first element every time in my testing. But some blogs warned that this sorting behavior isn't consistent. This lambda lives in our service account so that it can be tweaked later if we run into any issues without impacting customers.

packages/amplify-graphql-model-transformer/publish-notification-lambda/src/index.ts
sundersc: We query only the regions we support and already published layers. So it should have at least one element.

packages/amplify-graphql-model-transformer/publish-notification-lambda/src/index.ts
sundersc: This runs on the service account. We take the latest layer version of all regions and publish the SNS message. However client receive the notification specific to their region only, we have filters set on SNS subscription on the customer account based on their app region.

packages/amplify-graphql-model-transformer/rds-patching-lambda/src/index.ts
sundersc: Though these types are same, right now they live in two different lambdas. We don't share any details between these lambdas. Probably we can export all these types to a common package but let's keep it outside the scope of this PR.

packages/amplify-graphql-model-transformer/publish-notification-lambda/src/index.ts
```diff
@@ -0,0 +1,100 @@
...
+
+type LayerConfig = {
+  layerArn: string | undefined;
+  region: string;
+};
+
+const getLatestVersion = async (layerName: string, region: string): Promise<LayerConfig> => {
+  const client = new LambdaClient({ region });
+  const versions = [];
+  let response: ListLayerVersionsCommandOutput | undefined;
```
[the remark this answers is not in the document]
sundersc: Fixed this.

packages/amplify-graphql-model-transformer/publish-notification-lambda/package.json
```diff
@@ -0,0 +1,23 @@
+{
+  "name": "Publish Notification Message",
+  "version": "1.0.0",
+  "description": "",
+  "main": "lib/index.js",
+  "scripts": {
+    "test": "npm run build && node ./lib/index.js",
+    "build": "tsc",
+    "release": "rm -f notification-lambda.zip && bestzip --force node ./notification-lambda.zip ./*"
```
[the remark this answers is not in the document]
sundersc: I've added them to dependencies by mistake. Moved them to devdependencies.

packages/amplify-graphql-model-transformer/package.json
sundersc: I'm reviewing this seperately, need sometime to test this in windows. I will take care of all these scripts introduced with RDS project in a follow up PR.

packages/amplify-graphql-model-transformer/publish-notification-lambda/src/index.ts
```diff
@@ -0,0 +1,100 @@
...
+  'us-east-2',
+  'ap-northeast-1',
+  'eu-west-3',
+  'ap-southeast-2',
+  'ca-central-1',
+  'ap-northeast-3',
+];
+
+type LayerConfig = {
+  layerArn: string | undefined;
```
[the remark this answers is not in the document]
sundersc: Fixed this.

packages/amplify-graphql-model-transformer/rds-patching-lambda/package.json
sundersc: Updated.

alharris-at (approved):

packages/amplify-e2e-tests/src/__tests__/rds-import-vpc.test.ts
```diff
@@ -19,6 +19,13 @@ import gql from 'graphql-tag';
 import path from 'path';
 import { print } from 'graphql';
 
+const CDK_FUNCTION_TYPE = 'AWS::Lambda::Function';
+const CDK_SUBSCRIPTION_TYPE = 'AWS::SNS::Subscription';
+const APPSYNC_DATA_SOURCE_TYPE = 'AWS::AppSync::DataSource';
+
+const SNS_TOPIC_REGION = 'us-east-1';
+const SNS_TOPIC_ARN = 'arn:aws:sns:us-east-1:956468067974:AmplifyRDSLayerNotification';
```
phani-srikar, line 27: which account does this correspond to? Can we define this as a constant with an intuitive name?

packages/amplify-graphql-model-transformer/publish-notification-lambda/tsconfig.json
```diff
@@ -0,0 +1,109 @@
+{
```
phani-srikar, line 1: I would prefer not having so much commented out code or a quick follow up to enable these.

packages/amplify-graphql-model-transformer/rds-patching-lambda/tsconfig.json
```diff
@@ -0,0 +1,109 @@
+{
+  "compilerOptions": {
+    /* Visit https://aka.ms/tsconfig to read more about this file */
```
phani-srikar, line 3: same here: do we have a way forward to un-comment these?

packages/amplify-graphql-model-transformer/src/resources/rds-model-resource-generator.ts
```diff
@@ -1,12 +1,23 @@
+import { MYSQL_DB_TYPE, RDSConnectionSecrets } from '@aws-amplify/graphql-transformer-core';
```
phani-srikar, line 1: I feel like we're overloading the model transformer with all this functionality. Is there a way to do this outside of this transformer? Since patching is not exactly a concern of a GraphQL transformer.

phani-srikar (changes requested):

packages/amplify-e2e-tests/src/__tests__/rds-import-vpc.test.ts
sundersc: All RDS service resources point to beta account now. Not sure about moving the account number to a different constant will add anything to the clarity. I'm going to have a follow up PR to change all these URLs to point to production account. I will add necessary details in comments for more clarity.

packages/amplify-graphql-model-transformer/publish-notification-lambda/tsconfig.json
sundersc: This is the default tsc generated by 'tsc --init'. I can clean it up as a fast follow.

packages/amplify-graphql-model-transformer/src/resources/rds-model-resource-generator.ts
sundersc: Model directive is mostly responsible for provisioning the datasource. In this case, patching the lambda layer and SQL lambda are core components of datasource for RDS models. The lambda code is not big enough to move into a new package yet. We can start with having them here and move it easily if needed.

packages/amplify-e2e-tests/src/__tests__/rds-import-vpc.test.ts
phani-srikar: What is the difference between using `beta` vs `prod`. Generally we want to test our changes on `beta` before pushing to `prod`, is there a documented way to achieve this?

sundersc: The ARN is different for beta and prod. We can have both in the comments. If required, we can toggle the URL and test it against beta. 

We can also add a flag '--use-rds-beta'. If passed, we can deploy with beta layer and SNS notification. But that can be added iteratively, not necessarily on this PR.

phani-srikar (approved):

PR #1603, opened 2023-06-21
chore: refactor remaining bool config into transormParameters

alharris-at:
#### Description of changes
Refactor the following pieces of config and state into the `transformParameters` in order to collect config flags into a single location.

* `suppressApiKeyGeneration` - this turns the current tri-state for `legacyApiKeyEnabled` into a straightforward non-optional boolean.
* `enableSearchNodeToNodeEncryption`
* `disableResolverDeduping`
* `sandboxModeEnabled`

__N.B. This will require the mock util tests to be updated again in CLI when we release, since they're reaching inside the transformer to run those tests.__

N/A

#### Issue #, if available
N/A

Unit + E2E Tests

#### Checklist
- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-graphql-transformer-interfaces/src/transformer-context/transform-parameters.ts
+++ b/packages/amplify-graphql-transformer-interfaces/src/transformer-context/transform-parameters.ts
@@ -7,10 +7,13 @@
 export type TransformParameters = {
   // General Model Params
   shouldDeepMergeDirectiveConfigDefaults: boolean;
+  disableResolverDeduping: boolean;
+  sandboxModeEnabled: boolean;
 
   // Auth Params
   useSubUsernameForDefaultIdentityClaim: boolean;
   populateOwnerFieldForStaticGroupAuth: boolean;
+  suppressApiKeyGeneration: boolean;
 
   // Index Params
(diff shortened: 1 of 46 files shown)


packages/amplify-graphql-transformer-interfaces/src/transformer-context/transform-parameters.ts
```diff
@@ -7,15 +7,21 @@
...
 
   // Index Params
   secondaryKeyAsGSI: boolean;
   enableAutoIndexQueryNames: boolean;
 
   // Relational Params
   respectPrimaryKeyAttributesOnConnectionField: boolean;
+
+  // Search Params
+  enableNodeToNodeEncryption: boolean;
```
phani-srikar, line 26: This param here doesn't necessarily say that it corresponds to the searchable config. Maybe something like `enableSearchNodeToNodeEncryption` will make it more clear? If the customers are shown appropriate doc string, that should be okay.

phani-srikar (dismissed):

sundersc (dismissed):

sundersc (approved):

phani-srikar (approved):

PR #1604, opened 2023-06-22
chore: updating types to what we expect for alpha scope

alharris-at:
#### Description of changes
Updating types per conversations w/ a few folks, and feedback so far, but hopefully this is close to what we expect the final types to look like (though I expect some minor feedback and updates during alpha/beta testing).

Updates in this PR.
* No longer return a `Record<String, T> `, since the keys of that record are contextual, and not necessarily obvious to users. Instead, return a `T[]` so they can perform simple list comprehensions, filters, etc. to pull out the value they wish based on arbitrary values in the key (there is no data/context loss with this reshape).
* Expose a `referencedFunctions` object, which is intended to provide an affordance to name CDK generated (and named) functions, and eliminate the current amplify idiom of searching for a function by `name` in the account+region, `name+${env}`. This isn't wired through yet, and will throw an exception if used.
* Alias string to `IdentityPoolId` and provide an `IdentityPool` object instead of identityPoolId in the IAM config input, this paves the way for a new union type when [IdentityPool](https://docs.aws.amazon.com/cdk/api/v2/docs/@aws-cdk_aws-cognito-identitypool-alpha.IdentityPool.html)/[IIdentityPool](https://docs.aws.amazon.com/cdk/api/v2/docs/@aws-cdk_aws-cognito-identitypool-alpha.IIdentityPool.html) is out of alpha.
* Rename resource types to align w/ their L1 construct name, including Cfn. This makes it more clear when we add L2 support (where possible) that these are different representations of the same objects (and so we don't need to change context of what a resource key means).
* FunctionSlot now accepts a `function` property, which specifies a subset of an appsync function configuration. This is intended to provide an ingress for JS resolver support, as well as overrides w/ custom datasources. Currently throws if anything other than `requestMappingTemplate` or `responseMappingTemplate` is provided, but it does support both (yay).

__N.B. This is a pre-alpha state, so I dropped the test coverage bar so it doesn't break everytime I tweak the interface while I'm performing rapid iteration. Will stabilize shortly.__

N/A

#### Issue #, if available
N/A

Unit tests pass.

#### Checklist
- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-graphql-api-construct/src/internal/user-defined-slots.ts
+++ b/packages/amplify-graphql-api-construct/src/internal/user-defined-slots.ts
@@ -2,6 +2,62 @@ import { UserDefinedSlot, UserDefinedResolver } from '@aws-amplify/graphql-trans
 import { FunctionSlot } from '../types';
 
 /**
+ * Validate that only supported props are being passed into the funciton slots.
+ * @param functionSlots the slot inputs to validate.
+ */
+export const validateFunctionSlots = (functionSlots: FunctionSlot[]): void => {
+  functionSlots.forEach(({
+    function: {
+      code,
+      runtime,
+      dataSource,
+      name,
+      description,
+      responseMappingTemplate,
+      requestMappingTemplate,
+    },
+  }) => {
+    if (code || runtime || dataSource || name || description) {
+      throw new Error('Unexpected property found on function slot, only requestMappingTemplate and responseMappingTemplate supported');
+    }
+    if (!requestMappingTemplate && !responseMappingTemplate) {
+      throw new Error('Expected at least one of either requestMappingTemplate or responseMappingTemplate');
+    }
+  });
+};
+
+/**
+ * Short term hack, but we'll partition any slots have both a request and response mapping template into two,
+ * to make the existing system work.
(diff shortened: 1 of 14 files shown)


dpilch (approved):

packages/amplify-graphql-api-construct/src/internal/user-defined-slots.ts
```diff
@@ -39,30 +95,37 @@ export const parseUserDefinedSlots = (userDefinedTemplates: FunctionSlot[]): Rec
...
-    // because a slot can have a request and response resolver, we need to group corresponding request and response resolvers
-    if (slotHash in groupedResolversMap && resolverOrder in groupedResolversMap[slotHash]) {
-      setIn(groupedResolversMap, [slotHash, resolverOrder, resolverType], resolver);
-    } else {
-      const slot = {
-        resolverTypeName: slicedSlotName[0],
-        resolverFieldName: slicedSlotName[1],
-        slotName,
-        [resolverType]: resolver,
+  userDefinedTemplates
```
phani-srikar, line 98: nit: this method feels like it might need some refactoring later on.

phani-srikar (approved):

PR #1619, opened 2023-06-26
feat(api): add rds layer mapping json

sundersc:
Add RDS Lambda Layer region mapping JSON.


sundersc:
We decided to host this file in S3 instead.

PR #1624, opened 2023-06-27
fix: modelgen test promise issue

AaronZyLee:
#### Description of changes
The error is not rejected in codebuild but in CircleCI in modelgen test. The PR is to fix this issue.

#### Issue #, if available
The `unknown type` error are both thrown if I check both the artifacts from CCI and CodeBuild. Also the error stack trace is same. However, the codebuild does not reject the promise.  The only difference is that the error of `amplify-meta.json` not existing is displayed in different colors(red in CCI and white in CodeBuild)

- In CCI:
The output is too quick to get a screenshot. Attach the link for reference
https://output.circle-artifacts.com/output/job/1e954839-8ff8-44f8-a7ec-7a5830836d2b/artifacts/0/packages/amplify-e2e-tests/amplify-e2e-reports/index.html
- In CodeBuild:
image

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/packages/amplify-e2e-tests/src/__tests__/datastore-modelgen.test.ts
+++ b/packages/amplify-e2e-tests/src/__tests__/datastore-modelgen.test.ts
@@ -7,6 +7,15 @@ import {
   deleteProjectDir
 } from 'amplify-category-api-e2e-core';
 
+// This is to fix the issue of error not rejected in the codebuild,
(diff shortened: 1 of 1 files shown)


phani-srikar (approved):
Discussed offline, underlying behavior is same which is to throw when an unknown type exists in schema.

alharris-at (approved):

PR #1629, opened 2023-06-28
feat(api): use latest layer for sql lambda

sundersc:
#### Description of changes

- Pull the latest layer version from a HTTPS url.
- Fallback to hardcoded layer version if the url is not reachable.
- Handle SSM timeout for SQL lambda deployed within VPC if endpoints are not defined.

NA

#### Issue #, if available

NA

- E2E tests
- Manual test

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-category-api/src/graphql-transformer/transform-graphql-schema-v2.ts
```diff
@@ -27,6 +28,8 @@ const SCHEMA_FILENAME = 'schema.graphql';
 const SCHEMA_DIR_NAME = 'schema';
 const PROVIDER_NAME = 'awscloudformation';
 
+const LAYER_MAPPING_URL = 'https://amplify-rds-layer-resources.s3.amazonaws.com/rds-layer-mapping.json';
```
sundersc, line 31: This file will be periodically updated with the latest layer information. Clients will use this to determine which layer to use for SQL lambda. If the file is not accessible, it will fallback to the hardcoded version mapping.

phani-srikar: How do we update this S3 object? Is it upon every new layer version we publish from our pipeline or every API plugin release?

packages/amplify-graphql-model-transformer/src/resolvers/rds/resolver.ts
```diff
@@ -47,86 +47,96 @@ const RDSLayerMappingID = 'RDSLayerResourceMapping';
...
+  if (latestLayers && Object.keys(latestLayers).length > 0) {
+    return latestLayers;
+  }
+  const defaultLayerMapping = getDefaultLayerMapping();
+  return defaultLayerMapping;
+};
+
+// For beta use account '956468067974', layer name 'AmplifyRDSLayerBeta' and layer version '12' as of 2023-06-20
+// For prod use account '582037449441', layer name 'AmplifyRDSLayer' and layer version '3' as of 2023-06-20
+const getDefaultLayerMapping = (): RDSLayerMapping => ({
```
phani-srikar, line 68: How do we keep these values in sync with the S3 object? Do we need to?

packages/amplify-category-api/src/graphql-transformer/transform-graphql-schema-v2.ts
sundersc: At this point, it is manual update. Once we add integration tests for the RDS layer, we can confidently automate this.

packages/amplify-graphql-model-transformer/src/resolvers/rds/resolver.ts
sundersc: These values will be used only if the S3 mapping file is unreachable. Ideally we need to check and update if required before publishing a release.

phani-srikar (approved):

packages/amplify-category-api/src/graphql-transformer/transform-graphql-schema-v2.ts
```diff
@@ -213,6 +218,19 @@ const buildAPIProject = async (
   return builtProject;
 };
 
+const getRDSLayerMapping = async (): Promise<RDSLayerMapping> => {
+  try {
+    const response = await fetch(LAYER_MAPPING_URL);
```
alharris-at, line 223: Should we be explicitly importing node-fetch instead of using native `fetch`?

packages/amplify-category-api/src/graphql-transformer/transform-graphql-schema-v2.ts
```diff
@@ -213,6 +218,19 @@ const buildAPIProject = async (
...
+    if (response.status === 200) {
+      const result = await response.json();
+      return result as RDSLayerMapping;
+    }
+  } catch (err) {
+    // Ignore the error and return default layer mapping
+  }
```
alharris-at, line 230: Should we warn on non-200/failing fetch requests? So we at least have some indication in logs that the latest layer isn't being pulled.

packages/amplify-e2e-tests/src/__tests__/rds-v2.test.ts
```diff
@@ -147,4 +148,70 @@ describe("RDS Tests", () => {
     // PrimaryKey directive must be defined on Id field.
     expect(contactsIdFieldType.directives.find(d => d.name.value === 'primaryKey')).toBeDefined();
   });
+
+  // This test must be updated if the rds layer mapping file is updated
+  test("check the rds layer mapping file on the service account is available", async () => {
```
alharris-at, line 153: I'm not sure about this test, this seems like it introduces fragility to the system, in terms of having tests fail randomly (from 1 developer's perspective) because a new layer is pushed.

I'm not firmly against it, but not sure it adds significant value. It does help us know that our layer publish mechanism works, but this seems like the wrong place to be confirming that.

packages/amplify-graphql-model-transformer/rds-lambda/handler.ts
```diff
@@ -18,19 +20,30 @@ const createSSMClient = (): void => {
   secretsClient = new SSMClient({});
 };
 
-const getSSMValue = async(key: string | undefined): Promise<string> => {
+const wait10SecondsAndThrowError = async (): Promise<void> => {
+  await delay(10000);
+  console.log('Unable to retrieve secret for database connection from SSM. If your database is in VPC, verify that you have VPC endpoints for SSM defined and the security group\'s inbound rule for port 443 is defined.');
```
alharris-at, line 25: Instead of console.log, should probably add a log to the transformer logger methods https://github.com/aws-amplify/amplify-category-api/blob/main/packages/amplify-graphql-transformer-core/src/transformation/transformer-plugin-base.ts#L60-L62

sundersc: This logic resides in the SQL lambda deployed on the customer account.

packages/amplify-category-api/src/graphql-transformer/transform-graphql-schema-v2.ts
sundersc: Added a warning message if CLI is unable to pull the latest layer config.

packages/amplify-category-api/src/graphql-transformer/transform-graphql-schema-v2.ts
sundersc: Since we are using node18 now, I don't see a problem with using native fetch.

packages/amplify-e2e-tests/src/__tests__/rds-v2.test.ts
sundersc: Currently publishing the S3 file with the latest layer config is a manual process. Since we do not have the integration tests yet, we don't want to say that the latest layer is safe to use.

The target is eventually to get to a state where all this is automated, we will publish the S3 file through our pipeline and we can add checking the availability of the S3 file to a canary. For now, this test verifies the availability of the file until we automate this.

alharris-at (approved):

PR #1635, opened 2023-06-29
chore: add unit tests, fix bug for date marshalling

alharris-at:
#### Description of changes
Adding unit tests

N/A

#### Issue #, if available
N/A

Unit tests

#### Checklist
- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-graphql-default-value-transformer/src/validators.ts
+++ b/packages/amplify-graphql-default-value-transformer/src/validators.ts
@@ -17,8 +17,8 @@ const validateDate = (date: string): boolean => {
   // Verify the correct number of days for
   // the month contained in the date-string
   const year = Number(date.slice(0, 4));
-  const month = Number(date.slice(5, 2));
-  const day = Number(date.slice(8, 2));
+  const month = Number(date.slice(5, 7));
+  const day = Number(date.slice(8, 10));
 
   switch (month) {
     case 2: // February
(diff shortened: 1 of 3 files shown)


packages/amplify-graphql-default-value-transformer/src/validators.ts
```diff
@@ -17,8 +17,8 @@ const validateDate = (date: string): boolean => {
   // Verify the correct number of days for
   // the month contained in the date-string
   const year = Number(date.slice(0, 4));
-  const month = Number(date.slice(5, 2));
-  const day = Number(date.slice(8, 2));
+  const month = Number(date.slice(5, 7));
```
phani-srikar, line 20: this looks like a bug that we're fixing here?

phani-srikar (commented):
one question

packages/amplify-graphql-default-value-transformer/src/validators.ts
alharris-at: Yeah, I think somebody interpreted the slice api as (start, length), but it's (start, end). The effect here is that day/month were ALWAYS deserialized as 0, and we never did day value validation.

alharris-at: It's tested in the AWSDate tests (e.g. ensuring that feb fails on 29 in some cases, on 30 in nov, etc.

alharris-at: Previously those tests would always pass.

phani-srikar (approved):

PR #1654, opened 2023-07-07
chore(api): upgrade cdk version to 2.80

sundersc:
Upgrade CDK version to 2.80.

### DIFF
--- a/packages/amplify-graphql-function-transformer/package.json
+++ b/packages/amplify-graphql-function-transformer/package.json
@@ -35,7 +35,7 @@
     "graphql-transformer-common": "4.24.7"
   },
   "peerDependencies": {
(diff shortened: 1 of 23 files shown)


dpilch (approved):

packages/amplify-graphql-function-transformer/package.json
```diff
@@ -35,7 +35,7 @@
     "graphql-transformer-common": "4.24.7"
   },
   "peerDependencies": {
-    "aws-cdk-lib": "^2.68.0",
+    "aws-cdk-lib": "^2.80.0",
```
alharris-at, line 38: FWIW I don't think these updates to the peerDeps should be required.

alharris-at (approved):

PR #1667, opened 2023-07-12
feat: support view-test-artifacts rather than a single artifact

alharris-at:
#### Description of changes
Add a utility to make it easier to look at multiple failing test artifacts in parallel.

asset-download

N/A

#### Issue #, if available
N/A

Verified locally.

#### Checklist
- [x] PR description included

### DIFF
--- a/scripts/view-test-artifacts.ts
+++ b/scripts/view-test-artifacts.ts
@@ -1,0 +1,187 @@
+/**
+ * Usage: `yarn view-test-artifacts <buildBatchId>`
+ * 
+ * N.B. It is important to have your local environment configured for the correct codebuild account before running the script.
+ * This script caches resources, but in the case the local resource state is out of sync, you may need to wipe the asset directory
+ * that is printed when the script begins.
+ */
+
+import * as process from 'process';
+import * as path from 'path';
+import * as os from 'os';
+import * as fs from 'fs';
+// eslint-disable-next-line import/no-extraneous-dependencies
+import { CodeBuild, S3 } from 'aws-sdk';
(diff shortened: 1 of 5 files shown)


phani-srikar:
Looks like the `lint` job in cb is failing, can you run `prettier` on the changed file to fix it?

scripts/view-test-artifacts.ts
```diff
@@ -0,0 +1,187 @@
...
+    )
+    .filter(testArtifactIsComplete);
+};
+
+const downloadSingleTestArtifact = async (tempDir: string, artifact: Required<TestArtifact>): Promise<unknown[]> => {
+  const artifactDownloadPath = path.join(tempDir, artifact.jobName);
+  fs.mkdirSync(artifactDownloadPath);
+  const [Bucket, ...rest] = artifact.artifactLocation.split(':::')[1].split('/');
+  const Prefix = rest.join('/');
+  const listObjectsResponse = await s3.listObjects({ Bucket, Prefix }).promise();
```
phani-srikar, line 109: As per the AppSec guidance, we need to enable versioning on the buckets which would need some additional filtering here to only get the latest versions of objects.

alharris-at: Sounds good, I'll wait til that change is out and we've had some runs to verify behavior before we move forward w/ that.

phani-srikar: I've tested this script with versioned buckets and it works. Good to go.

phani-srikar (approved):

PR #1705, opened 2023-07-21
Remove vm2 for Plugin on main

phani-srikar:
#### Description of changes
Removes the `vm2` dependency from the plugin.
The CLI core and cloudformation provider deps from CLI still fetch it, so not fully removed from `yarn.lock` yet.

The following E2E tests cover positive and negative tests for applying overrides:
- GraphQL API -> https://github.com/aws-amplify/amplify-category-api/blob/main/packages/amplify-e2e-tests/src/__tests__/api_7.test.ts
- Rest API -> https://github.com/aws-amplify/amplify-category-api/blob/main/packages/amplify-e2e-tests/src/__tests__/apigw.test.ts 

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-category-api/src/graphql-transformer/override.ts
+++ b/packages/amplify-category-api/src/graphql-transformer/override.ts
@@ -1,9 +1,8 @@
-import * as path from 'path';
-import { CfnResource } from 'aws-cdk-lib';
-import * as fs from 'fs-extra';
-import * as vm from 'vm2';
-import _ from 'lodash';
-import { pathManager, stateManager } from '@aws-amplify/amplify-cli-core';
+import { CfnResource } from "aws-cdk-lib";
+import * as fs from "fs-extra";
+import * as path from "path";
+import _ from "lodash";
+import { pathManager, stateManager } from "@aws-amplify/amplify-cli-core";
 import { StackManager } from '@aws-amplify/graphql-transformer-core';
 import { AmplifyApiGraphQlResourceStackTemplate } from '@aws-amplify/graphql-transformer-interfaces';
 import { getAppSyncAPIName } from '../provider-utils/awscloudformation/utils/amplify-meta-utils';
@@ -63,18 +62,6 @@ export function applyFileBasedOverride(stackManager: StackManager, overrideDirPa
   });
 
   const appsyncResourceObj = convertToAppsyncResourceObj(amplifyApiObj);
-  const overrideCode: string = fs.readFileSync(overrideFilePath, 'utf-8');
-  const sandboxNode = new vm.NodeVM({
-    console: 'inherit',
-    timeout: 5000,
-    sandbox: {},
-    require: {
-      context: 'sandbox',
-      builtin: ['path'],
-      external: true,
-    },
-  });
-  // Remove these when moving override up to amplify-category-api level
   const { envName } = stateManager.getLocalEnvInfo();
   const { projectName } = stateManager.getProjectConfig();
   const projectInfo = {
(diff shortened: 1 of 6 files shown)


packages/amplify-category-api/src/graphql-transformer/override.ts
```diff
@@ -63,26 +62,17 @@ export function applyFileBasedOverride(stackManager: StackManager, overrideDirPa
...
-  // Remove these when moving override up to amplify-category-api level
   const { envName } = stateManager.getLocalEnvInfo();
   const { projectName } = stateManager.getProjectConfig();
   const projectInfo = {
     envName,
     projectName,
   };
   try {
-    sandboxNode.run(overrideCode, overrideFilePath).override(appsyncResourceObj, projectInfo);
+    const overrideImport = require(overrideFilePath);
```
phani-srikar, line 72: Since this is a synchronous method, await import doesn't work. Have to fall back to require.

sobolk: How bad is making call stack async ?

phani-srikar: pretty intense. It ties into the `transform` and thereby `compile`. Will need an API change as well.

alharris-at: That'd be problematic until this refactor https://github.com/aws-amplify/amplify-category-api/pull/1699 is complete (and won't be done for a week probably). We can revisit at that point, but until then, this OverrideCallback is in the CDK construct ctor critical path :(

sobolk (approved):

yarn.lock
```diff
@@ -657,10 +657,10 @@
   resolved "https://registry.npmjs.org/@aws-cdk/asset-node-proxy-agent-v5/-/asset-node-proxy-agent-v5-2.0.165.tgz#c169599d83beceea7e638082ef9833997f04c85d"
   integrity sha512-bsyLQD/vqXQcc9RDmlM1XqiFNO/yewgVFXmkMcQkndJbmE/jgYkzewwYGrBlfL725hGLQipXq19+jwWwdsXQqg==
 
-"@aws-cdk/aws-apigatewayv2-alpha@~2.68.0-alpha.0":
-  version "2.68.0-alpha.0"
-  resolved "https://registry.npmjs.org/@aws-cdk/aws-apigatewayv2-alpha/-/aws-apigatewayv2-alpha-2.68.0-alpha.0.tgz#f9cfe135eedbd60aacc48735b63323ea7961c5db"
-  integrity sha512-UZ1GBt7DfxKLfs49Fg1Y33/RRVoD1NZDd0V+e9Tj/dkSm9d0JEiuMU8TRtandsd38kl7n1fSLEVsXu3/WAYAPg==
+"@aws-cdk/aws-apigatewayv2-alpha@~2.80.0-alpha.0":
+  version "2.80.0-alpha.0"
+  resolved "https://registry.yarnpkg.com/@aws-cdk/aws-apigatewayv2-alpha/-/aws-apigatewayv2-alpha-2.80.0-alpha.0.tgz#37a1c2ef166aea24bcf58b3efa776511c6b8569a"
```
alharris-at, line 662: What's triggering the registry change here?

alharris-at (approved):

dpilch (approved):

packages/amplify-category-api/src/graphql-transformer/override.ts
sobolk: Sounds good.
Lesson learnt - going forward it's good to skew towards async apis earlier.
I.e. if there's any fs access - use async apis not sync as forcing factor.

yarn.lock
phani-srikar: It might be because my `yarn config get registry` is set to the yarn package manager, I'll update that in my dev environment going forward.

PR #1721, opened 2023-07-25
feat: store construct output

dpilch:
#### Description of changes

Store construct output with `storeOutput` function.

Some items are copied here until published by the CLI.

Add outputs:
* awsAppsyncApiEndpoint
* awsAppsyncAuthenticationType
* awsAppsyncRegion
* awsAppsyncApiKey

#### Issue #, if available

N/A

Unit testing

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


alharris-at (dismissed):

alharris-at (approved):

packages/amplify-graphql-api-construct/src/graphql-output.ts
```diff
@@ -0,0 +1,102 @@
...
+export type BackendOutputStorageStrategy<T extends BackendOutputEntry> = {
+  addBackendOutputEntry(keyName: string, backendOutputEntry: T): void;
+
+  /**
+   * Write all pending data to the destination
+   */
+  flush(): void;
+};
+
+export const amplifyStackMetadataKey = 'AWS::Amplify::Output';
```
dpilch, line 45: We should probably use a different metadata key. `AWS::Amplify::Output::Graphql` or something along those lines.

We would also need the CLI to make this configurable in the exported version of `StackMetadataBackendOutputStorageStrategy`.

packages/amplify-graphql-api-construct/API.md
```diff
@@ -198,6 +200,11 @@ export type UserPoolAuthorizationConfig = {
     userPool: IUserPool;
 };
 
+// Warnings were encountered during analysis:
+//
+// src/types.ts:387:3 - (ae-forgotten-export) The symbol "BackendOutputStorageStrategy" needs to be exported by the entry point index.d.ts
+// src/types.ts:387:3 - (ae-forgotten-export) The symbol "GraphqlOutput" needs to be exported by the entry point index.d.ts
```
alharris-at, line 206: Could you get these exports as well, so there are no missing exports?

packages/amplify-graphql-api-construct/src/__tests__/amplify-graphql-api.test.ts
```diff
@@ -33,4 +33,73 @@ describe('AmplifyGraphqlApi', () => {
...
+          apiKeyConfig: { expires: cdk.Duration.days(7) },
+        },
+      });
+
+      const addBackendOutputEntry = jest.fn();
+      const outputStorageStrategyMock = {
+        addBackendOutputEntry,
+        flush: jest.fn(),
+      };
+      api.storeOutput(outputStorageStrategyMock);
```
alharris-at, line 58: You can use `Template.from(stack) to render and avoid testing w/ token Regexes 

https://docs.aws.amazon.com/cdk/v2/guide/testing.html#testing_fine_grained

packages/amplify-graphql-api-construct/src/graphql-output.ts
alharris-at: It'd need to be the same to have `generate` work consistently, right?

alharris-at (commented):
Minor comments, mostly on testing/exports.

packages/amplify-graphql-api-construct/src/__tests__/amplify-graphql-api.test.ts
alharris-at: Sorry, this points to the long lines, should have pointed to the `tokenRegex` usage.

packages/amplify-graphql-api-construct/src/graphql-output.ts
dpilch: Oh yes, you are right. One complication, I think we can only flush once for any given key. Customers would not be able to have multiple constructs with the manual config.

packages/amplify-graphql-api-construct/src/__tests__/amplify-graphql-api.test.ts
dpilch: I don't think this will useful here. We are asserting what is passed to the `outputStorageStrategy`. In this case they are refs. 

I'm going to add a test for the default `outputStorageStrategy`, which won't be a mock. In that case we will call `outputStorageStrategy.flush()` and then we will be able to render the template and see the stored metadata.

alharris-at (dismissed):

packages/amplify-graphql-api-construct/src/graphql-output.ts
alharris-at: Can we read/mutate the key?

alharris-at (approved):

PR #1738, opened 2023-07-28
move transform manager behavior into invokers

alharris-at:
#### Description of changes
Refactor the synthesis and injected providers for the transformer out of `@aws-amplify/graphql-transformer` and into the respective callers (today this is tests, the api category, and the cdk construct).

This sets us up to change those providers and flow in the CDK construct independently, and keep from impacting the API category (as confirmed via unit and e2e tests).

This is a follow where we previously hoisted this behavior out of the transformer core package, and into the transformer entry point package, in order to refactor the transformers and core layer themselves, without affecting callers. This PR basically finishes that job, of hoisting all the way to the callers, so this should have no impact on transformers or core, and only affect the transform entry point, and places where we invoke the transform.

You'll notice there are a few places where there is impact to the core package, but this is just some cleanup I was doing (and moving types which no longer need to be shared globally up to the API category package, etc.)

N/A - just a refactor of existing behavior.

#### Issue #, if available
N/A

Unit + E2E Tests

#### Checklist
- [x] PR description included
- [x] `yarn test` passes


sundersc (approved):
LGTM.

phani-srikar (approved):

PR #1741, opened 2023-07-31
chore: 🤖 add support for ap-northeast-3 KIX

petercwk-aws:
#### Description of changes
This PR enables support in ap-northeast-3 region (KIX)

N/A
#### Issue #, if available
Associated CLI PR: https://github.com/aws-amplify/amplify-cli/pull/13039

I set ap-northeast-3 in my ~/.aws/config for my amplify profile and created an app.

I ran: 
amplify-dev init
amplify-dev add api (default graphQL options)
amplify-dev push

Confirmed that an API was successfully deployed to my account in ap-northeast-3 region

#### Checklist

- [X] PR description included
- [X] `yarn test` passes

### DIFF
--- a/packages/amplify-e2e-core/src/configure/index.ts
+++ b/packages/amplify-e2e-core/src/configure/index.ts
@@ -27,6 +27,7 @@ export const amplifyRegions = [
   'eu-central-1',
   'ap-northeast-1',
   'ap-northeast-2',
+  'ap-northeast-3',
   'ap-southeast-1',
   'ap-southeast-2',
   'ap-south-1',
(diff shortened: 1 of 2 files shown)


dpilch (approved):

packages/amplify-e2e-core/src/configure/index.ts
```diff
@@ -27,6 +27,7 @@ export const amplifyRegions = [
   'eu-central-1',
   'ap-northeast-1',
   'ap-northeast-2',
+  'ap-northeast-3',
```
AaronZyLee, line 30: I think this change should be excluded until the CLI has released the region option in `amplify configure`. Otherwise it could cause the API e2e test failure. (I am testing with latest version `12.2.3` and do not find the new region).

AaronZyLee (changes requested):

packages/amplify-e2e-core/src/configure/index.ts
petercwk-aws: Removed in the last revision.

PR #1751, opened 2023-08-01
Merge Refactor of GraphQL Transformer Concerns into Main Branch

alharris-at:
#### Description of changes
Major refactor of the core transformer system to remove internal synthesis.

Today the graphql transformer core internally creates a cdk app, all custom stacks to support the Amplify CLI ecosystem, and then synthesizes said app into a set of cloudformation files (and other files assets). This is because the Amplify CLI manages S3 file uploads and Cfn stack interactions.

With the release of the Amplify Graphql CDK L3 Construct, we need to break that paradigm, instead executing the `transform` as a mutation on a CDK scope, creating the new resources as necessary, and attaching them to provided scope.

This was primarily done by injecting `providers` and input params where necessary to create slightly different behavior w.r.t. assets and nested stacks (stacks and nested stacks being another class of `asset` for the purposes of this review.

N/A, we're relying on snapshot tests that are unchanging, and e2e tests passing without change to convince ourselves that this refactor is unimpactful for the Graphql API.

Yes, significant CDK changes on the alpha construct, but that's expected, and okay at this lifecycle state.

#### Issue #, if available
N/A

Unit + E2E tests that pass (working on getting e2e suite passing now).

#### Checklist
- [x] PR description included
- [x] `yarn test` passes


alharris-at:
E2Es are all passing except for searchable-migraiton, which is failing due to a timeout (but confirmed locally to be working), and the v6/v10 migration tests, which are also failing on `main`.

packages/amplify-graphql-api-construct/src/internal/construct-exports.ts
```diff
@@ -1,109 +1,163 @@
...
+    .filter(([name]) => name.split('.').length === 6)
+    .map(([name, resolverCode]) => {
+      const [typeName, fieldName, slotName, slotIndex, templateType] = name.split('.');
+      return {
+        typeName,
+        fieldName,
+        slotName,
+        slotIndex: Number.parseInt(slotIndex, 10),
+        function: {
+          // TODO: this should consolidate req/req values back together
```
phani-srikar, line 158: is this any concerning TODO that we left unintentionally?

AaronZyLee (approved):

phani-srikar (approved):

packages/amplify-graphql-api-construct/src/internal/construct-exports.ts
alharris-at: Oh no, I meant to strip the todo, but must have reverted. It's a DX improvement for the CDK construct that's covered by an existing piece of QA.

PR #1752, opened 2023-08-01
chore: dx improvement for codebuild

AaronZyLee:
#### Description of changes

- E2e test
  - The `api_2` is splitted to `api_11` with the most time-consuming case (turn on conflict resolution and then turn off)
  - Fixed `iterative-rollback` tests not catching the exit code issue. In codebuild, the `CtrlC` is not sending the 130 but 2 exit code
- Cleanup
  - Fixed the issue of cleanup job always cleaning all stale resources in the e2e workflow. It should only clean the resources within the batch
  - `stale-resources.json` are correctly generated for each individual account
  - The info of stale resources are also uploaded as artifacts for cleanup job
- Artifact
  - Test artifacts are now correctly uploaded for those run with retries. The `RUN_INDEX` is added in the path. Previously only the second run is uploaded.
  - Also fix the view artifact script to support this change
  - Use `yarn view-test-artifact <code-build-id>` instead of s3 URI to view the artifact of single build

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes


packages/amplify-e2e-tests/src/__tests__/api_2.test.ts
```diff
@@ -129,43 +127,6 @@ describe('amplify add api (GraphQL)', () => {
     expect(updateResultData.updateNote.note).toEqual(updateInput.input.note);
   });
 
-  it('init a project with conflict detection enabled and toggle disable', async () => {
```
AaronZyLee, line 132: This case takes the majority time of the whole test suites(around 30~40mins). This is moved to a new test file `api_11`

scripts/view-test-artifacts.ts
```diff
@@ -115,7 +115,8 @@ const downloadSingleTestArtifact = async (tempDir: string, artifact: Required<Te
 
   return Promise.all(
     listObjectsResponse.Contents.filter(hasKey).map(({ Key }) => {
-      const filePath = path.join(artifactDownloadPath, Key.split('/').pop()!);
+      const filePath = path.join(artifactDownloadPath, Key.split('/').slice(3).join('/'));
```
AaronZyLee, line 118: This change is to support the artifact change, which have the run index (0 & 1) in the path

shared-scripts.sh
```diff
@@ -269,7 +269,7 @@ function _cleanupE2EResources {
   echo "Running clean up script"
   build_batch_arn=$(aws codebuild batch-get-builds --ids $CODEBUILD_BUILD_ID | jq -r -c '.builds[0].buildBatchArn')
   echo "Cleanup resources for batch build $build_batch_arn"
-  yarn clean-e2e-resources --buildBatchArn $build_batch_arn
+  yarn clean-e2e-resources buildBatchArn $build_batch_arn
```
AaronZyLee, line 272: This is to fix the bug that the cleanup job in e2e workflow always cleaning all the stale resources. It should only cleanup the resources generated within the batch

packages/amplify-e2e-core/src/categories/auth.ts
```diff
@@ -127,7 +127,7 @@ interface AddApiOptions {
 
 const defaultOptions: AddApiOptions = {
   apiName: '\r',
-  testingWithLatestCodebase: true,
+  testingWithLatestCodebase: false,
```
alharris-at, line 130: ? Seems like we want to leave this on.

alharris-at (commented):
Overall LGTM - just a question on flipping default behavior to 'testNewCode: false'

packages/amplify-e2e-core/src/categories/auth.ts
phani-srikar: This is set to `true` only for migration tests where we install the older CLI version globally and use the locally built binary as the latest. Unless any migration tests are using these functions without explicitly passing in `testingWithLatestCodebase=true`, we should be good.

phani-srikar (approved):
Have you run e2e on this change atleast to confirm that none of the migration tests are impacted? Otherwise LGTM.

alharris-at (approved):

packages/amplify-e2e-core/src/categories/auth.ts
AaronZyLee: When it is true, the test consumes the binary installed within the node modules of repo root. In this test the binary installed from local registry will be used and it cannot be consumed by test unless the value is false.

AaronZyLee:
> Have you run e2e on this change atleast to confirm that none of the migration tests are impacted? Otherwise LGTM.

Yes the migration tests all look good

PR #1756, opened 2023-08-02
Add remaining L1 resources, as well as limited L2 resources.

alharris-at:
#### Description of changes
Return L1 and L2 CFN Resources, tweaking shape slightly, and adding L2 GraphqlApi, Tables, Roles, and Functions.

N/A

#### Issue #, if available
N/A

Unit Tests

#### Checklist
- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-graphql-api-construct/API.md
+++ b/packages/amplify-graphql-api-construct/API.md
@@ -19,7 +19,9 @@ import { Construct } from 'constructs';
 import { Duration } from 'aws-cdk-lib';
 import { IBucket } from 'aws-cdk-lib/aws-s3';
 import { IFunction } from 'aws-cdk-lib/aws-lambda';
+import { IGraphqlApi } from 'aws-cdk-lib/aws-appsync';
 import { IRole } from 'aws-cdk-lib/aws-iam';
+import { ITable } from 'aws-cdk-lib/aws-dynamodb';
 import { IUserPool } from 'aws-cdk-lib/aws-cognito';
 import { SchemaFile } from 'aws-cdk-lib/aws-appsync';
 import { TransformerPluginProvider } from '@aws-amplify/graphql-transformer-interfaces';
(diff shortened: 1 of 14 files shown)


packages/amplify-graphql-transformer-core/src/transformation/sync-utils.ts
```diff
@@ -47,6 +48,8 @@ function createSyncIAMRole(context: TransformerContext, scope: Construct, tableN
     roleName: context.resourceHelper.generateIAMRoleName(SyncResourceIDs.syncIAMRoleName),
     assumedBy: new iam.ServicePrincipal('appsync.amazonaws.com'),
   });
+  setResourceName(role.node.defaultChild!, SyncResourceIDs.syncIAMRoleName);
+  setResourceName(role, SyncResourceIDs.syncIAMRoleName);
```
phani-srikar, line 52: qq: why do we need to set the name on the `defaultChild` as well as the object itself?

Since this pattern is repeating, handling it centrally inside the `setResourceName` method with an optional flag `setOnChild` if required, can save us some lines of code.

phani-srikar (approved):
Non-blocking comment.

packages/amplify-graphql-transformer-core/src/transformation/sync-utils.ts
AaronZyLee: I think it is because it is setting the name for both L1 & L2 construct. But I agree to wrap this into one method to save the lines.

AaronZyLee (approved):

packages/amplify-graphql-api-construct/API.md
```diff
@@ -42,6 +44,20 @@ export class AmplifyGraphqlApi<SchemaType = AmplifyGraphqlApiResources> extends
     readonly resources: AmplifyGraphqlApiResources;
 }
 
+// @public
+export type AmplifyGraphqlApiL1Resources = {
```
sobolk, line 48:
```suggestion
export type AmplifyGraphqlApiCfnResources = {
```

Issue #1714, opened 2023-07-14
Cannot mutate update on model without @auth delete


krosiek:
### Description

Looks similar to this https://github.com/aws-amplify/amplify-cli/issues/10087 but this issue does not provide any answer.
I have a model:
```
type UserData @model @auth(rules: [{allow: owner, operations: [create, read, update]}]) {
  email: String
  nick: String
}
```
I list entries with `ModelQueries.list` and then try to update:
```
final userData = items.first!;
final updatedUserData = userData.copyWith(nick: 'NickiNick');

final updateRequest = ModelMutations.update(updatedUserData);
final updateResponse = await Amplify.API.mutate(request: updateRequest).response;
if (updateResponse.errors.isNotEmpty) {
      return Failure.unexpected();
}
```
I cannot update my entry due to `"errorType": "Unauthorized"`.  
When the entry was added from code and I try to update it, it says: `"message": "Unauthorized on [id]"`
when from Appsync web client: `"message": "Unauthorized on [email, id]"`

The error disappears when I add `delete` to `@auth rules`, but then  even though `Amplify.API.mutate` runs without any error it does change anything in db.

Other models do not have this problem, just this`UserData`. Before I simplified `UserData` model to email and nick it had also members of enums and another model `prediction`. The `prediction` model was missing `update` rule in `@Auth` but even after removing dependency on `prediction` model and enums Unauthorized error appears.

`UserData` is also the only model that previously had `{allow: groups, groups: ["admin"], operations: [create, read, update, delete]}`

As Authorization Mode I use Cognito Pools

I can get, list, create from AppSync and from code without any error.

Any hints if it is a mistake on my part or a bug?

### Categories

- [ ] Analytics
- [ ] API (REST)
- [X] API (GraphQL)
- [X] Auth
- [ ] Authenticator
- [ ] DataStore
- [ ] Notifications (Push)
- [ ] Storage

### Platforms

- [X] iOS
- [ ] Android
- [ ] Web
- [ ] macOS
- [ ] Windows
- [ ] Linux

### Flutter Version

3.10.6

### Amplify Flutter Version

1.2.0

### Deployment Method

Amplify CLI

### Schema

```GraphQL
type UserData @model @auth(rules: [{allow: owner, operations: [create, read, update]}]) {
  email: String
  nick: String
}
```

PR #1762, opened 2023-08-03
fix: allow update on operations create, read, update

dpilch:
#### Description of changes

Allow update on operations create, read, update.

See https://github.com/aws-amplify/amplify-category-api/issues/1714 for details.

Include primary key in allowed fields on update. The primary key won't be updated, but this indicates that including the primary key in the input (which it always will be) should not result in unauthorized error.

Given the case `@auth(rules: [{allow: owner, operations: [create, read, update]}])`.

* The `nullAllowedFields` will be empty: https://github.com/aws-amplify/amplify-category-api/blob/db6c64570235f67deb3068998d6d8cb73ad5e15b/packages/amplify-graphql-auth-transformer/src/graphql-auth-transformer.ts#L920
* Resulting in `areAllFieldsNullAllowed` to be `false`: https://github.com/aws-amplify/amplify-category-api/blob/db6c64570235f67deb3068998d6d8cb73ad5e15b/packages/amplify-graphql-auth-transformer/src/graphql-auth-transformer.ts#L923
* `isAuthorizedOnAllFields` will be `false` in the resolver: https://github.com/aws-amplify/amplify-category-api/blob/db6c64570235f67deb3068998d6d8cb73ad5e15b/packages/amplify-graphql-auth-transformer/src/resolvers/mutation.update.ts#L155
  * when `isAuthorizedOnAllFields` is `true` the auth check on each field is short-circuited.
* Then the resolver checks all fields for auth included in the input
  * The primary key will always be included in the input because it is necessary to indicate which record should be updated.
  * https://github.com/aws-amplify/amplify-category-api/blob/db6c64570235f67deb3068998d6d8cb73ad5e15b/packages/amplify-graphql-auth-transformer/src/resolvers/mutation.update.ts#L331-L339
  * If the primary key is not in the allowedFields the resolver will give unauth error.

N/A

#### Issue #, if available

Resolves https://github.com/aws-amplify/amplify-category-api/issues/1714

* Unit testting

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-graphql-auth-transformer/src/graphql-auth-transformer.ts
```diff
@@ -921,7 +921,9 @@ export class AuthTransformer extends TransformerAuthBase implements TransformerA
       const roleDefinition = this.roleMap.get(role)!;
       roleDefinition.areAllFieldsAllowed = allowedFields.length === fields.length;
       roleDefinition.areAllFieldsNullAllowed = nullAllowedFields.length === fields.length;
-      roleDefinition.allowedFields = this.addAutoGeneratedFields(ctx, def, allowedFields, fields);
+      // include primary key in allowed fields on update
+      // the primary key won't be updated, but this indicates that including the primary key in the input should not result in unauth err
+      roleDefinition.allowedFields = this.addAutoGeneratedFields(ctx, def, [...allowedFields, getObjectPrimaryKey(def).name.value], fields);
```
sundersc, line 926: When customer explicitly define a primary key, it may be added twice to the array `[...allowedFields, getObjectPrimaryKey(def).name.value]`. Does it impact any functionality?

sundersc:
Can we add an E2E test for this scenario? That's more reliable than a snapshot for this case.

packages/amplify-graphql-auth-transformer/src/graphql-auth-transformer.ts
dpilch: `addAutoGeneratedFields` first makes a set out of the supplied `allowedFields`. I can add a test to ensure this behavior.

sundersc (approved):

packages/amplify-e2e-tests/src/schema-api-directives/tests/auth-owner12.ts
```diff
@@ -0,0 +1,73 @@
+// schema
+export const schema = `
+type Todo @model @auth(rules: [{allow: owner, operations: [create, read, update]}]) {
+  id: ID!
+  updatedAt: AWSDateTime!
+  content: String!
+}
+
+##auth/owner5`;
```
alharris-at, line 9: owner12?

alharris-at (dismissed):

packages/amplify-e2e-tests/src/schema-api-directives/tests/auth-owner12.ts
dpilch: Oops, quick copy and paste job.

alharris-at (approved):

AaronZyLee (approved):

sundersc (approved):

PR #1816, opened 2023-08-23
feat(api): custom queries support for rds using sql directive

sundersc:
#### Description of changes

Add custom queries support for RDS using @sql directive.
- `@sql(statement: "xyx")` allows to define SQL queries inline in RDS schema file.
- `@sql(reference: "abc")` allows to define SQL statements in a file under api/sql-statements directory.

NA

#### Issue #, if available

NA

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-graphql-sql-transformer/src/graphql-sql-transformer.ts
```diff
@@ -0,0 +1,148 @@
...
+        const { RDSLambdaDataSourceLogicalID: dataSourceId } = ResourceConstants.RESOURCES;
+        const dataSource = context.api.host.getDataSource(dataSourceId);
+        const MISSING_QUERY_STATEMENT = 'MISSING_CUSTOM_QUERY';
+        if (config.reference && !context.customQueries.has(config.reference)) {
+          throw new InvalidDirectiveError(
+            `@sql directive 'reference' argument must be a valid custom query name. Check type "${config.resolverTypeName}" and field "${config.resolverFieldName}". The custom query "${config.reference}" does not exist in "sql-statements" directory.`,
+          );
+        }
+
+        const statement = config.statement ?? (config.reference ? context.customQueries.get(config.reference) : MISSING_QUERY_STATEMENT);
```
phani-srikar, line 98: Instead of falling back to a `MISSING_QUERY_STATEMENT` and proceeding with resolver generation, shouldn't the case of not having valid `statement` or a pre-loaded `reference` SQL file be an Error instead? If you already have those checks, is `MISSING_QUERY_STATEMENT` default needed at all?

packages/amplify-graphql-sql-transformer/src/graphql-sql-transformer.ts
```diff
@@ -0,0 +1,148 @@
...
+export const generateSqlLambdaRequestTemplate = (statement: string, operation: string, operationName: string): string => {
+  return printBlock('Invoke RDS Lambda data source')(
+    compoundExpression([
+      set(ref('lambdaInput'), obj({})),
+      set(ref('lambdaInput.parameters'), obj({})),
+      set(ref('lambdaInput.statement'), str(statement)),
+      set(ref('lambdaInput.operation'), str(operation)),
+      set(ref('lambdaInput.operationName'), str(operationName)),
+      set(ref('lambdaInput.parameters'), methodCall(ref('util.defaultIfNull'), ref('context.arguments'), obj({}))),
+      obj({
```
phani-srikar, line 131: Is there a clean way to detect and fail fast if the variables expected by the SQL statement do not exist in the input? This will save customers invalid lambda invocations.

packages/amplify-graphql-sql-transformer/src/graphql-sql-transformer.ts
sundersc: MISSING_QUERY_STATEMENT has been added to make the compiler happy. Otherwise `statement` would be `string | undefined` instead of a `string`. I could have used `context.customQueries.get(config.reference)!` but preferred this approach better.

phani-srikar: There are couple of cases that we should check: 
- When the `statements` here is an empty string.
- When both statements and reference are provided.

phani-srikar (approved):

packages/amplify-graphql-sql-transformer/src/graphql-sql-transformer.ts
```diff
@@ -0,0 +1,164 @@
...
+
+  generateResolvers = (context: TransformerContextProvider): void => {
+    if (this.sqlDirectiveFields.size === 0) {
+      return;
+    }
+
+    const stack: cdk.Stack = context.stackManager.createStack(SQL_DIRECTIVE_STACK);
+    const env = context.synthParameters.amplifyEnvironmentName;
+
+    stack.templateOptions.templateFormatVersion = '2010-09-09';
```
alharris-at, line 80: Why is this required?

alharris-at (approved):

packages/amplify-graphql-sql-transformer/src/graphql-sql-transformer.ts
sundersc: This is to handle an edge case where customer hasn't used any `@model` types but using the new SQL directive alone. This makes sure that the stack is created.

PR #1823, opened 2023-08-24
fix: reading admin role names

phani-srikar:
#### Description of changes

 In CLI `11.1.1`, specifying the admin role names as a string in the `custom-roles.json` file as shown below is supported.
```
{
"adminRoleNames": "amplify-123456-unauthRole"
}
```
While in the latest version `12.3.0`, we fail with error message `customRoleConfig.adminRoleNames.map is not a function`.
This PR fixes this regression and restores the original behavior.

Added unit tests,
Tested in a JS app that a single string input for admin role names compiles.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-category-api/src/__tests__/graphql-transformer/utils.test.ts
+++ b/packages/amplify-category-api/src/__tests__/graphql-transformer/utils.test.ts
@@ -1,8 +1,8 @@
 import * as path from 'path';
 import { DatasourceType } from '@aws-amplify/graphql-transformer-core';
 import * as fs from 'fs-extra';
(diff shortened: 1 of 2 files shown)


AaronZyLee (approved):

sundersc:
Can we add a test to make sure that the generated auth resolvers contain the admin roles?

palpatim (approved):

phani-srikar:
> Can we add a test to make sure that the generated auth resolvers (after schema transformation) contain the admin roles?

There are a couple of E2E's that check this already, one of them is the `api_10.test.ts` that's passing. This change only impacts how we read the `adminRoles` into transformer options which happens even before any transformer is invoked, so we should be good. Tested it with my local app.

PR #1824, opened 2023-08-24
feat(api): specify include or exclude tables option to generate schema

sundersc:
#### Description of changes
Include/Exclude Option
- Customers can specify in the input AMPLIFY to exclude or only include certain tables from being re-generated.
- Customers can not have "include" and "exclude" statements at the same time. they are mutually exclusive.

**Example:**
```
input AMPLIFY {
   exclude: [String] = ["todos", "hello"]
}
```
or
```
input AMPLIFY {
   include: [String] = ["todos", "hello"]
}
```

This change also includes preserving customer added relational fields on schema regeneration.

NA

#### Issue #, if available

NA

- Manual test
- New unit tests
- New E2E test

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-graphql-schema-generator/src/schema-generator/schema-overrides.ts
+++ b/packages/amplify-graphql-schema-generator/src/schema-generator/schema-overrides.ts
@@ -37,9 +37,6 @@ const preserveRelationalDirectives = (document: DocumentNode, existingDocument: 
   const RELATIONAL_DIRECTIVES = ['hasOne', 'hasMany', 'belongsTo'];
   const documentWrapper = document as any;
 
-  if (!existingDocument) {
-    return document;
-  }
   existingDocument.definitions
     .filter((def) => def.kind === 'ObjectTypeDefinition' && def.directives.find((dir) => dir.name.value === MODEL_DIRECTIVE_NAME))
     .forEach((existingObject: ObjectTypeDefinitionNode) => {
--- a/packages/amplify-graphql-schema-generator/src/schema-generator/schema-overrides.ts
+++ b/packages/amplify-graphql-schema-generator/src/schema-generator/schema-overrides.ts
@@ -1,15 +1,17 @@
 import { DocumentNode, FieldDefinitionNode, ObjectTypeDefinitionNode, visit } from 'graphql';
 import { isArrayOrObject, findMatchingField, getNonModelTypes, isOfType, isNonNullType } from 'graphql-transformer-common';
 import { printer } from '@aws-amplify/amplify-prompts';
+import { FieldWrapper, ObjectDefinitionWrapper } from '@aws-amplify/graphql-transformer-core';
 
 export const applySchemaOverrides = (document: DocumentNode, existingDocument?: DocumentNode | undefined): DocumentNode => {
   if (!existingDocument) {
     return document;
   }
 
+  let updatedDocument = preserveRelationalDirectives(document, existingDocument) as any;
   const schemaVisitor = {
     FieldDefinition: {
-      leave(node: FieldDefinitionNode, key, parent, path, ancestors): any {
+      leave: (node: FieldDefinitionNode, key, parent, path, ancestors) => {
         const parentObjectType = getParentNode(ancestors);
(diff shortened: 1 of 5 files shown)


packages/amplify-graphql-schema-generator/src/schema-generator/schema-overrides.ts
```diff
@@ -18,16 +20,54 @@ export const applySchemaOverrides = (document: DocumentNode, existingDocument?:
...
 };
 
+const preserveRelationalDirectives = (document: DocumentNode, existingDocument: DocumentNode): DocumentNode => {
+  const MODEL_DIRECTIVE_NAME = 'model';
+  const RELATIONAL_DIRECTIVES = ['hasOne', 'hasMany', 'belongsTo'];
+  const documentWrapper = document as any;
+
+  if (!existingDocument) {
+    return document;
+  }
```
alharris-at, line 42: This check shouldn't be required, since you don't specify as optional, and also explicitly already check this in the caller.

packages/amplify-graphql-schema-generator/src/schema-generator/schema-overrides.ts
```diff
@@ -18,16 +20,54 @@ export const applySchemaOverrides = (document: DocumentNode, existingDocument?:
...
+      }
+      const relationalFields = existingObject.fields.filter((field) =>
+        field.directives.find((dir) => RELATIONAL_DIRECTIVES.includes(dir.name.value)),
+      );
+      const newObjectWrapper = new ObjectDefinitionWrapper(newObject);
+      relationalFields.forEach((relationalField: FieldDefinitionNode) => {
+        newObjectWrapper.fields.push(new FieldWrapper(relationalField));
+      });
+
+      if (relationalFields.length > 0) {
```
alharris-at, line 60: This branch creates a potentially bizarre modal behavior. If I'm reading this right if I go make changes to my Model after import, then upon the next import those changes are scrubbed, but if those changes ALSO include a relational update then those changes are maintained (not just the relationship).

Is that the desired behavior, or should we be doing the merge one layer deeper (e.g. only maintain relational fields in the colliding models?

sundersc: That a great question. This logic copies only the relationship. `newObject` is the regenerated GraphQL type and `existingObject` is the current state of the GraphQL type from schema.rds.graphql file. We only copy the relational fields from existing type to the new type in this method (line 56-58).

I'm adding a test specific to this scenario.

However, I had other concerns that I will review with Rene and address in a follow up if required.
- What if the table belongs to the related type is deleted? Should we still retain the relationship?
- What if one of the references fields removed?

alharris-at: Got it - thank you for explaining. I missed that on my first read, but makes sense after reviewing.

alharris-at (approved):

packages/amplify-graphql-schema-generator/src/schema-generator/schema-overrides.ts
```diff
@@ -1,15 +1,17 @@
...
 import { isArrayOrObject, findMatchingField, getNonModelTypes, isOfType, isNonNullType } from 'graphql-transformer-common';
 import { printer } from '@aws-amplify/amplify-prompts';
+import { FieldWrapper, ObjectDefinitionWrapper } from '@aws-amplify/graphql-transformer-core';
 
 export const applySchemaOverrides = (document: DocumentNode, existingDocument?: DocumentNode | undefined): DocumentNode => {
   if (!existingDocument) {
     return document;
   }
 
+  let updatedDocument = preserveRelationalDirectives(document, existingDocument) as any;
```
phani-srikar, line 11: One alternative here is to extend the schema visitor pattern we use for `FieldDefinition` to `ObjectTypeDefinition` as well, this could save us from multiple mutations of `updatedDocument`. Non-blocking suggestion.

phani-srikar (approved):

PR #1826, opened 2023-08-25
fix: Fix schema import for RDS clusters in VPCs

palpatim:
#### Description of changes

Fixes schema imports for RDS clusters in VPCs:

1. Added `await`s to `getHostVpc`. The null coalescing evaluation never evaluated the right hand side since `checkHostInDBInstances` returns a perfectly non-null Promise that rejects. :)
1. Added a couple of tests for `getHostVpc` using mocked data to represent instance & cluster configs. These aren't comprehensive by any stretch, but get us to test 1. It's reasonable to add e2e tests for the cluster case as well, but I'll do that under a separate PR if desired
1. Added regions to all IAMClient constructions. IAMClient calls fail if there isn't a default region set, which at least my config doesn't have.

N/A

#### Issue #, if available

N/A

1. New unit test
2. Local testing

#### Checklist

- [X] PR description included
- [X] `yarn test` passes
- [X] Tests are [changed or added](https://github.com/aws-amplify/amplify-cli/blob/master/CONTRIBUTING.md#tests)
- [ ] ~~Relevant documentation is changed or added (and PR referenced)~~
- [ ] ~~New AWS SDK calls or CloudFormation actions have been added to relevant test and service IAM policies~~
- [ ] ~~Any CDK or CloudFormation parameter changes are called out explicitly~~

### DIFF
--- a/packages/amplify-graphql-schema-generator/src/utils/__tests__/vpc-helper.test.ts
+++ b/packages/amplify-graphql-schema-generator/src/utils/__tests__/vpc-helper.test.ts
@@ -1,0 +1,248 @@
+import {
+  RDSClient,
+  DescribeDBClustersCommandOutput,
+  DescribeDBInstancesCommandOutput,
+  DescribeDBSubnetGroupsCommandOutput,
+} from '@aws-sdk/client-rds';
+
+import {
+  getHostVpc,
+} from '../vpc-helper';
+
+
+const sendSpy = jest.spyOn(RDSClient.prototype, 'send');
+
+describe('detect VPC settings', () => {
+
+  it('should detect VPC settings for an RDS instance', async () => {
+    // TS complains about resolving values in the spy.
+    // Cast it through never to overcome the compile error.
+    sendSpy
+      .mockResolvedValueOnce(instanceResponse as never);
+
+    const result = await getHostVpc(
+      'mock-rds-cluster-instance-1.aaaaaaaaaaaa.us-west-2.rds.amazonaws.com',
+      'us-west-2'
+    );
+
+    expect(result).toBeDefined();
+  });
+
+  it('should detect VPC settings for an RDS cluster', async () => {
+
+    // TS complains about resolving values in the spy.
+    // Cast it through never to overcome the compile error.
+    sendSpy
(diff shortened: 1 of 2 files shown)


packages/amplify-graphql-schema-generator/src/utils/__tests__/vpc-helper.test.ts
```diff
@@ -0,0 +1,248 @@
...
+    // Cast it through never to overcome the compile error.
+    sendSpy
+      .mockResolvedValueOnce(instanceResponse as never);
+
+    const result = await getHostVpc(
+      'mock-rds-cluster-instance-1.aaaaaaaaaaaa.us-west-2.rds.amazonaws.com',
+      'us-west-2'
+    );
+
+    expect(result).toBeDefined();
```
phani-srikar, line 28: is there a reason to not have a stronger assertion about the output here that it should match the VPC config that's mocked in the response?

packages/amplify-graphql-schema-generator/src/utils/__tests__/vpc-helper.test.ts
```diff
@@ -0,0 +1,248 @@
...
+      .mockResolvedValueOnce(instanceResponse as never)
+      .mockResolvedValueOnce(clusterResponse as never)
+      .mockResolvedValueOnce(subnetResponse as never);
+
+    const result = await getHostVpc(
+      'mock-rds-cluster.cluster-abc123.us-west-2.rds.amazonaws.com',
+      'us-west-2'
+    );
+
+    expect(result).toBeDefined();
```
phani-srikar, line 45: same as above.

packages/amplify-graphql-schema-generator/src/utils/__tests__/vpc-helper.test.ts
palpatim: :sigh: I forgot to go back and assert the actual data. :(  Thanks for catching, will update.

phani-srikar (approved):

sundersc (approved):

sundersc (approved):

PR #1830, opened 2023-08-28
feat: add refersTo directive transformer for model renaming

phani-srikar:
#### Description of changes

- Add a new `@refersTo` directive for RDS models to enable renaming them.
- Refactor the name mappings handling in transformers based on the datasource.

Unit, Integration and E2E tests added.
Added E2E tests pass locally.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


depcheck.config.js
```diff
@@ -13,7 +13,7 @@ const GQL_V2_TRANSFORMER_PACKAGES = [
   'amplify-graphql-function-transformer',
   'amplify-graphql-http-transformer',
   'amplify-graphql-index-transformer',
-  'amplify-graphql-maps-to-transformer',
+  'amplify-graphql-name-mapping-transformer',
```
sundersc, line 16: Is there a reason why we can't use the existing package name?

phani-srikar: The NPM package name is still the same, but the folder is renamed to more appropriately convey that it now holds 2 transformers related to name mapping.

packages/amplify-graphql-searchable-transformer/src/graphql-searchable-transformer.ts
```diff
@@ -447,6 +448,9 @@ export class SearchableModelTransformer extends TransformerPluginBase {
     // but if mappings are defined this will ensure the mapping is also applied to the search results
     for (const def of this.searchableObjectTypeDefinitions) {
       const modelName = def.node.name.value;
+      if (isRDSModel(ctx as TransformerContextProvider, modelName)) {
```
palpatim, line 451: I'm assuming we will not support `@searchable` for RDS at all--should we throw an error rather than silently ignoring?

phani-srikar: Converted this to an Error and added an unit test. We actually have a check for unsupported directives even before the transformers are invoked and a plan to move it into the transfotmers before release. So, this is one step towards it.

codebuild_specs/e2e_workflow.yml
```diff
@@ -146,35 +146,34 @@ batch:
...
+    - identifier: api_6_api_lambda_auth_api_9
       buildspec: codebuild_specs/run_e2e_tests.yml
       env:
         compute-type: BUILD_GENERAL1_MEDIUM
         variables:
           TEST_SUITE: >-
-            src/__tests__/graphql-v2/api_lambda_auth.test.ts|src/__tests__/api_9.test.ts
-          CLI_REGION: eu-central-1
+            src/__tests__/api_6.test.ts|src/__tests__/graphql-v2/api_lambda_auth.test.ts|src/__tests__/api_9.test.ts
+          CLI_REGION: us-east-1
```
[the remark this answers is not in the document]
phani-srikar: We have a pre-commit Git hook that randomizes the region assignments of tests to ensure we test from multiple supported regions. So, you'll find this file updated in most PRs.

palpatim (approved):

alharris-at (approved):

packages/amplify-graphql-name-mapping-transformer/src/__tests__/graphql-refers-to-transformer.test.ts
```diff
@@ -0,0 +1,166 @@
...
+
+  it('throws if a model name mapping is duplicate', () => {
+    const [stubDefinition, stubDirective, stubTransformerContext] = getTransformerInputsFromSchema(
+      duplicateNameMappingModelSchema,
+      modelName,
+    );
+    expect(() =>
+      refersToTransformer.object(stubDefinition as ObjectTypeDefinitionNode, stubDirective as DirectiveNode, stubTransformerContext),
+    ).toThrowErrorMatchingInlineSnapshot(
+      `"Cannot apply @refersTo with name \\"OriginalName\\" on type \\"TestName\\" because \\"DuplicateName\\" model already has the same name mapping."`,
```
sundersc, line 154: Do we throw an error if customer is using refersTo on DDB models?

phani-srikar: Yes we do, the test pointed out below i.e `can be applied only on RDS models` does that check. Updated the below test.

sundersc (approved):

PR #1840, opened 2023-09-01
feat: enable jsii builds for the api construct

alharris-at:
#### Description of changes
Building the `AmplifyGraphqlApi` construct using JSII instead of TSC. This is a required for indexing in [Construct Hub](https://constructs.dev/).

Input prop shapes for the CDK construct have changed slightly, but the underlying system is unaffected by this PR.

#### Issue #, if available
N/A

Unit + E2E Tests pass, build + package all work.

[E2E Test Run](https://us-east-1.console.aws.amazon.com/codesuite/codebuild/594813022831/projects/amplify-category-api-e2e-workflow/batch/amplify-category-api-e2e-workflow%3Ae1d28d71-4829-4c2a-a279-5b5a30c6c465/?region=us-east-1) (only excluding prettierignore change)

#### Checklist
- [x] PR description included
- [x] `yarn test` passes


packages/amplify-graphql-api-construct/tsconfig.json
```diff
@@ -1,11 +1,34 @@
...
+    "strictNullChecks": true,
+    "strictPropertyInitialization": true,
+    "stripInternal": false,
+    "target": "ES2020",
+    "composite": false,
+    "tsBuildInfoFile": "lib/tsconfig.tsbuildinfo"
   },
-  "exclude": ["coverage", "lib", "src/__tests__"]
+  "include": ["src/**/*.ts"],
+  "exclude": ["node_modules", "/Users/alharris/workplace/amplify-category-api/packages/amplify-graphql-api-construct/lib/.types-compat"],
```
AaronZyLee, line 32: The absolute local path should be fixed

AaronZyLee (commented):
Just check the code build. Looks like the build job does not succeed for the e2e tests. 
```
$ tsc --build tsconfig.tests.json
--
250 | lerna ERR! yarn run build-tests exited 1 in 'amplify-category-api-e2e-tests'
251 | lerna ERR! yarn run build-tests stdout:
252 | $ tsc --build tsconfig.tests.json
253 | src/__tests__/cdk/backends/base-cdk/app.ts(15,3): error TS2322: Type 'string' is not assignable to type 'IAmplifyGraphqlSchema'.
254 | src/__tests__/cdk/backends/custom-logic/app.ts(18,3): error TS2322: Type 'string' is not assignable to type 'IAmplifyGraphqlSchema'.
255 | src/__tests__/cdk/backends/relationships/app.ts(14,3): error TS2322: Type 'string' is not assignable to type 'IAmplifyGraphqlSchema'.

```

packages/amplify-graphql-api-construct/tsconfig.json
alharris-at: Oh, this file actually should no longer be committed, it's generated during the jsii build process.

phani-srikar: remove local workspace reference. Also, any reason to not exclude the `__tests__`?

packages/amplify-graphql-api-construct/.jsii
```diff
@@ -0,0 +1,5770 @@
...
+  "author": {
+    "name": "Amazon Web Services",
+    "roles": [
+      "author"
+    ]
+  },
+  "bundled": {
+    "@aws-amplify/graphql-transformer": "1.1.2",
+    "@aws-amplify/graphql-transformer-core": "2.1.2",
+    "@aws-amplify/graphql-transformer-interfaces": "3.1.2",
```
phani-srikar, line 11: How do we ensure these get bumped in every release? When do we intend to refresh this file?

packages/amplify-graphql-api-construct/tsconfig.json
alharris-at: Yeah, similar to the response to Aaron's comment above, but this file is now auto-generated. I forgot to delete it in the first pass.

packages/amplify-graphql-api-construct/.jsii
alharris-at: This file is generated by the `build` step, so it'll be up-to-date.

alharris-at:
E2E Tests are all passing now except for the existing one that's known, failing.

packages/amplify-e2e-tests/src/__tests__/cdk/base-cdk.test.ts
```diff
@@ -21,65 +21,67 @@ describe('CDK GraphQL Transformer', () => {
     deleteProjectDir(projRoot);
   });
 
-  test('CDK base case', async () => {
-    const templatePath = path.resolve(path.join(__dirname, 'backends', 'base-cdk'));
-    const name = await initCDKProject(projRoot, templatePath);
-    const outputs = await cdkDeploy(projRoot, '--all');
-    const { GraphQLAPIEndpointOutput: apiEndpoint, GraphQLAPIKeyOutput: apiKey } = outputs[name];
+  ['2.80.0', 'latest'].forEach((cdkVersion) => {
```
sundersc, line 24: Is there a reason why we are testing it on CDK 2.80 in addition to the latest?

sundersc:
Is there a way to add a test to make sure that the construct we build here (using JSII) is compatible with CDK on other programming languages? I don't think we need to necessarily do it in this PR, something to consider later.

phani-srikar (approved):
Does JSII provide any guarantee out of the box that the packaged construct will be compatible with the target platforms? In other words, how do we ensure the changes we make don't break Java consumers for example.

packages/amplify-e2e-tests/src/__tests__/cdk/base-cdk.test.ts
alharris-at: Yeah, this makes sure that our min/max peerDep versions both work.

alharris-at:
Re: the questions about testing other platforms, we'll want to add that, but this out of the box doesn't support other platforms.

dpilch (commented):
I am getting a build error on a clean repo.

```
yarn setup-dev && cd packages/amplify-graphql-api-construct/ && yarn build
```

```
[2023-09-08T10:33:42.539] [ERROR] jsii/compiler - Compilation errors prevented the JSII assembly from being created
src/amplify-graphql-api.ts:95:7 - error TS2322: Type 'this' is not assignable to type 'Construct'.
...
```

The `yarn build` fails in `packages/amplify-graphql-api-construct/`. I need to run `yarn` in `packages/amplify-graphql-api-construct/` before `yarn build` will work.

PR #1841, opened 2023-09-01
feat(api): always use vpc for aws rds instances and clusters 

sundersc:
#### Description of changes
- Deploy the SQL lambda in VPC if the database is in VPC regardless whether the database is accessible through internet or not.
- Add VPC endpoints required to read the secrets if the database is in a VPC.
- Inject the SSM VPC endpoints to SQL lambda environment variable.
- SQL lambda uses SSM endpoint from the environment variable.

NA

#### Issue #, if available
NA

- Manual test
- E2E tests

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-category-api/src/graphql-transformer/transform-graphql-schema-v2.ts
+++ b/packages/amplify-category-api/src/graphql-transformer/transform-graphql-schema-v2.ts
@@ -10,8 +10,9 @@ import {
   AppSyncAuthConfiguration,
   TransformerLog,
   TransformerLogLevel,
-  VpcConfig,
+  VpcSubnetConfig,
   RDSLayerMapping,
+  SubnetAvailabilityZone,
 } from '@aws-amplify/graphql-transformer-interfaces';
 import fs from 'fs-extra';
 import { ResourceConstants } from 'graphql-transformer-common';
@@ -36,6 +37,7 @@ import { TransformerProjectOptions } from './transformer-options-types';
 import { DeploymentResources } from './cdk-compat/deployment-resources';
 import { TransformManager } from './cdk-compat/transform-manager';
 import { checkForUnsupportedDirectives } from '../provider-utils/awscloudformation/utils/rds-resources/utils';
+import { getAvaliabilityZoneOfSubnets } from '../provider-utils/vpc-utils';
 
 const PARAMETERS_FILENAME = 'parameters.json';
 const SCHEMA_FILENAME = 'schema.graphql';
@@ -207,9 +209,9 @@ const buildAPIProject = async (context: $TSContext, opts: TransformerProjectOpti
   const { modelToDatasourceMap } = opts.projectConfig;
   const datasourceSecretMap = await getDatasourceSecretMap(context);
   const datasourceMapValues: Array<DatasourceType> = modelToDatasourceMap ? Array.from(modelToDatasourceMap.values()) : [];
-  let sqlLambdaVpcConfig: VpcConfig | undefined;
+  let sqlLambdaVpcConfig: VpcSubnetConfig | undefined;
   if (datasourceMapValues.some((value) => value.dbType === MYSQL_DB_TYPE && !value.provisionDB)) {
-    sqlLambdaVpcConfig = await isSqlLambdaVpcConfigRequired(context, getSecretsKey(), ImportedRDSType.MYSQL);
+    sqlLambdaVpcConfig = await isSqlLambdaVpcConfigRequired(context);
   }
   const rdsLayerMapping = await getRDSLayerMapping();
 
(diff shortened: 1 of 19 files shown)


packages/amplify-category-api/src/graphql-transformer/transform-graphql-schema-v2.ts
```diff
@@ -302,12 +302,19 @@ const getDatasourceSecretMap = async (context: $TSContext): Promise<Map<string,
   return outputMap;
 };
 
-const getSQLLambdaVpcConfig = async (context: $TSContext): Promise<VpcConfig | undefined> => {
+const getSQLLambdaVpcConfig = async (context: $TSContext): Promise<VpcSubnetConfig> => {
```
palpatim, line 305: Although you intend to use this for Lambda, you're really getting the VPC config for the RDS host here, right? A rename might clarify that your intent is to install the Lambda into the RDS host's VPC. I can see an eventual requirement to install the Lambda into its own VPC, and let customers set up a private link between the Lambda VPC & the RDS VPC, so the rename will make it clearer to future devs (i.e., us) where that change needs to happen :)

packages/amplify-e2e-tests/src/__tests__/rds-model-v2.test.ts
```diff
@@ -616,7 +647,7 @@ describe('RDS Model Directive', () => {
     return listResult;
   };
 
-  const checkGenericError = async (errorMessage?: string) => {
+  const checkGenericError = async (errorMessage?: string): Promise<void> => {
```
palpatim, line 650: Why was this async?

packages/amplify-graphql-model-transformer/rds-lambda/handler.ts
```diff
@@ -18,7 +18,11 @@ export const run = async (event): Promise<any> => {
 };
 
 const createSSMClient = (): void => {
-  secretsClient = new SSMClient({});
+  const DNS_SEPERATOR = ':';
+  const endpoint = process.env.SSM_ENDPOINT?.split(DNS_SEPERATOR).pop();
+  secretsClient = new SSMClient({
+    endpoint: `https://${endpoint}`,
```
palpatim, line 24: I had some problems using the SDK in my local setup if I didn't pass region to IAMClient. Do we need to plumb region through here, or will the constructor automatically pick it up from Lambda environment vars?

packages/amplify-graphql-model-transformer/src/resolvers/rds/resolver.ts
```diff
@@ -152,13 +169,60 @@ export const createRdsLambda = (
...
+    serviceName: Fn.join('', [serviceEndpointPrefix, '.', Fn.ref('AWS::Region'), '.', serviceSuffix]), // Sample: com.amazonaws.us-east-1.ssmmessages
+    vpcEndpointType: 'Interface',
+    vpcId: sqlLambdaVpcConfig.vpcConfig.vpcId,
+    subnetIds: extractSubnetForVpcEndpoint(sqlLambdaVpcConfig.subnetAvailabilityZoneConfig),
+    securityGroupIds: sqlLambdaVpcConfig.vpcConfig.securityGroupIds,
+    privateDnsEnabled: false,
+  });
+};
+
+const addVpcEndpointForSecretsManager = (
```
palpatim, line 194: naming nit: You're setting up more endpoints than just secretsmanager here

palpatim (approved):
A couple of questions & some naming nits, but nothing blocking

packages/amplify-category-api/src/graphql-transformer/transform-graphql-schema-v2.ts
```diff
@@ -276,19 +278,17 @@ const getRDSLayerMapping = async (): Promise<RDSLayerMapping> => {
...
-  secretsKey: string,
-  engine: ImportedRDSType,
-): Promise<VpcConfig | undefined> => {
-  const { secrets } = await getConnectionSecrets(context, secretsKey, engine);
-  const isDBPublic = await testDatabaseConnection(secrets);
-  if (isDBPublic) {
-    // No need to deploy the SQL Lambda in VPC if the DB is public
+const isSqlLambdaVpcConfigRequired = async (context: $TSContext): Promise<VpcSubnetConfig | undefined> => {
+  // If the database is in VPC, we will use the same VPC configuration for the SQL lambda.
+  // Customers are required to add inbound rule for port 443 from the private subnet in the Security Group.
```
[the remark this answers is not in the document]
sundersc: Since we are now using VPC endpoints specific to the subnet, I didn't have to add any inbound rule. But with my setup, my subnets are all connected (means all traffic allowed from security group).

I will review this again with few more setups and update the comments as required.

packages/amplify-e2e-tests/src/__tests__/rds-model-v2.test.ts
sundersc: Oops, it doesn't have to be. This is an existing code, caught by lint fix.

packages/amplify-category-api/src/graphql-transformer/transform-graphql-schema-v2.ts
```diff
@@ -276,19 +278,17 @@ const getRDSLayerMapping = async (): Promise<RDSLayerMapping> => {
...
-const isSqlLambdaVpcConfigRequired = async (
-  context: $TSContext,
-  secretsKey: string,
-  engine: ImportedRDSType,
-): Promise<VpcConfig | undefined> => {
-  const { secrets } = await getConnectionSecrets(context, secretsKey, engine);
-  const isDBPublic = await testDatabaseConnection(secrets);
-  if (isDBPublic) {
-    // No need to deploy the SQL Lambda in VPC if the DB is public
+const isSqlLambdaVpcConfigRequired = async (context: $TSContext): Promise<VpcSubnetConfig | undefined> => {
```
phani-srikar, line 281: nit: `isSqlLambdaVpcConfigRequired` sounds like it would return a `boolean`.

phani-srikar (approved):

packages/amplify-graphql-model-transformer/src/resolvers/rds/resolver.ts
sundersc: Though we are creating endpoints for different services, all these are required for secrets manager to work E2E.

Refer here:
https://docs.aws.amazon.com/systems-manager/latest/userguide/setup-create-vpc.html#sysman-setting-up-vpc-create

packages/amplify-graphql-model-transformer/rds-lambda/handler.ts
sundersc: I haven't noticed any of the internal customers tried the product so far run into any issues. Does it happen if you don't have a default region?

I'm adding an item to QA finding so that this gets tested before GA.

PR #1849, opened 2023-09-06
feat: bootstrap for adding amplify managed table

AaronZyLee:
#### Description of changes
Bootstrap for adding the new table management for DynamoDB tables capable of multiple GSIs update. 

There are still features to come like the construct wrapping the custom resource, handling non-gsi table update in lambda, etc. Those are marked as TODO and will be implemented later.

Test included in the PR:
- `getNextGSIUpdate`
-  New table resource generator (partially)

Other unit and e2e tests will be added in separate PR. 

- New table resource generator
  - Custom resource with DynamoDB table properties
  - Use Custom Resource Provider Framework
  - The params & conditions are kept in the template
  - New nested Stack `AmplifyTableManager` for the framework resources & lambda functions
- Custom resource lambda functions
  - `onEvent`
    - Handles CFN event of `Create`, `Update` and `Delete`
    - Parsed the event to the table definition input
    - Make DynamoDB SDK calls based on event type
  - `isComplete`
    - Check if the table/indexes are ready
    - Make follow-up updates if there are any
- New transformer parameter: `useAmplifyManagedTableResources`
  - ONLY for internal usage

The changes are validated by my local CDK app. The `AmplifyGraphqlApi` construct package is linked locally for test.

So far I've successfully tested the new table management with the following scenarios:

- Create table
- Update table with adding 2 GSIs
- Update table with deleting 2 GSIs
- Change key schema to replace the table (delete then create)
- Delete table

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes


packages/amplify-graphql-model-transformer/src/resources/custom-resource-lambda/custom-resource-handler.ts
```diff
@@ -0,0 +1,437 @@
...
+          Data: {
+            TableArn: describeTableResult.Table.TableArn,
+            TableStreamArn: describeTableResult.Table.LatestStreamArn,
+            TableName: describeTableResult.Table.TableName
+          }
+        }
+        return result; // nothing to update
+      } 
+
+      // TODO: merge gsi update with other table updates
```
AaronZyLee, line 84: TODO: merge gsi update with other table updates

packages/amplify-graphql-model-transformer/src/resources/iterative-dynamo-model-resource-generator.ts
```diff
@@ -0,0 +1,371 @@
...
+    });
+    const usePointInTimeRecovery = new cdk.CfnCondition(scope, ResourceConstants.CONDITIONS.ShouldUsePointInTimeRecovery, {
+      expression: cdk.Fn.conditionEquals(pointInTimeRecovery, 'true'),
+    });
+
+    const removalPolicy = this.options.EnableDeletionProtection ? cdk.RemovalPolicy.RETAIN : cdk.RemovalPolicy.DESTROY;
+
+    // TODO: The following lines for custom resource creation will be wrapped into a new internal construct
+    // TODO: The attribute of encryption and TTL should be added
+    const defaultTableState = {
```
AaronZyLee, line 191: TODO: The following lines for custom resource creation will be wrapped into a new internal construct
TODO: The attribute of encryption and TTL should be added

packages/amplify-graphql-api-construct/src/internal/default-parameters.ts
```diff
@@ -14,4 +14,5 @@ export const defaultSchemaTranslationBehavior: SchemaTranslationBehavior = {
   enableAutoIndexQueryNames: true,
   respectPrimaryKeyAttributesOnConnectionField: true,
   enableSearchNodeToNodeEncryption: false,
+  useAmplifyManagedTableResources: false, // TODO: flip to true once all implementations are finished
```
AaronZyLee, line 17: TODO: flip to true once all implementations are finished

packages/amplify-graphql-model-transformer/src/resources/iterative-dynamo-model-resource-generator.ts
```diff
@@ -0,0 +1,368 @@
...
+  }
+
+  createCustomProviderResource(scope: Construct, context: TransformerContextProvider): void {
+    // Policy that grants access to Create/Update/Delete DynamoDB tables
+    const ddbManagerPolicy = new aws_iam.Policy(scope, 'CreateUpdateDeleteTablesPolicy');
+    ddbManagerPolicy.addStatements(
+      new aws_iam.PolicyStatement({
+        actions: ['dynamodb:CreateTable', 'dynamodb:UpdateTable', 'dynamodb:DeleteTable', 'dynamodb:DescribeTable'],
+        // TODO: have more restricted scope
+        resources: ['*'],
```
AaronZyLee, line 87: TODO: have more restricted scope

packages/amplify-graphql-model-transformer/src/resources/custom-resource-lambda/custom-resource-handler.ts
```diff
@@ -0,0 +1,440 @@
...
+  if (event.RequestType === 'Delete') {
+    // nothing else to do on delete
+    console.log('Delete is finished');
+    return finished;
+  }
+  if (!event.PhysicalResourceId) {
+    throw new Error('PhysicalResourceId not set in call to isComplete');
+  }
+  console.log('Fetching current table state');
+  const describeTableResult = await retry(
```
palpatim, line 129: Similar to above comment: we retry the describeTable here, but don't above when we check for existing state. Why the discrepancy?

packages/amplify-graphql-model-transformer/src/resources/custom-resource-lambda/custom-resource-handler.ts
```diff
@@ -0,0 +1,440 @@
...
+  const endStateGSIs = endState.GlobalSecondaryIndexes || [];
+  const endStateGSINames = endStateGSIs.map((gsi) => gsi.IndexName);
+
+  const currentStateGSIs = currentState.GlobalSecondaryIndexes || [];
+  const currentStateGSINames = currentStateGSIs.map((gsi) => gsi.IndexName);
+
+  // function to identify any GSIs that need to be removed
+  const gsiRequiresReplacementPredicate = (currentGSI: DynamoDB.GlobalSecondaryIndexDescription): boolean => {
+    // check if the index has been removed entirely
+    if (!endStateGSINames.includes(currentGSI.IndexName!)) return true;
```
palpatim, line 183: nit: do our linting rules allow if statements to have no-brace conditions?

packages/amplify-graphql-model-transformer/src/resources/custom-resource-lambda/custom-resource-handler.ts
```diff
@@ -0,0 +1,440 @@
...
+        },
+      ],
+    };
+  }
+
+  // no more updates necessary
+  return undefined;
+};
+
+type CreateTableResponse = {
```
palpatim, line 235: Is this factorable into a common type def, since the CDK code is going to operate on the type as well?

packages/amplify-graphql-model-transformer/src/resources/custom-resource-lambda/custom-resource-handler.ts
```diff
@@ -0,0 +1,440 @@
...
+    return true; // Table exists
+  } catch (error) {
+    if (error.code === 'ResourceNotFoundException') {
+      return false; // Table does not exist
+    }
+    throw error; // Handle other errors
+  }
+};
+
+const isProjectionModified = (currentProjection: Projection, endProjection: Projection): boolean => {
```
palpatim, line 333: ANy reason this isn't simply an "equals" function?

packages/amplify-graphql-model-transformer/src/resources/custom-resource-types.d.ts
```diff
@@ -0,0 +1,125 @@
+import { DynamoDB } from 'aws-sdk';
+
+export as namespace CustomDDB;
+
+export type Input = DynamoDB.CreateTableInput;
+
+// TYPES BELOW ARE COPIED FROM CDK
```
palpatim, line 7: Would it be better to declare a dev-dependency on @types/aws-cdk ?

packages/amplify-graphql-model-transformer/src/resources/iterative-dynamo-model-resource-generator.ts
```diff
@@ -0,0 +1,260 @@
...
+      }),
+    );
+
+    const lambdaCode = aws_lambda.Code.fromAsset(path.join(__dirname, '..', '..', 'lib', 'resources', 'custom-resource-lambda'));
+
+    // lambda that will handle DDB CFN events
+    const gsiOnEventHandler = new aws_lambda.Function(scope, ResourceConstants.RESOURCES.TableOnEventHandlerLogicalID, {
+      runtime: aws_lambda.Runtime.NODEJS_16_X,
```
palpatim, line 93: Should we go ahead & use Node 18?

packages/amplify-graphql-model-transformer/src/resources/iterative-dynamo-model-resource-generator.ts
```diff
@@ -0,0 +1,260 @@
...
+      code: lambdaCode,
+      handler: 'custom-resource-handler.onEvent',
+      timeout: Duration.minutes(14),
+    });
+
+    // lambda that will poll for provisioning to complete
+    const gsiIsCompleteHandler = new aws_lambda.Function(scope, ResourceConstants.RESOURCES.TableIsCompleteHandlerLogicalID, {
+      runtime: aws_lambda.Runtime.NODEJS_16_X,
```
palpatim, line 101: Should we go ahead & use Node 18?

packages/amplify-graphql-model-transformer/src/resources/iterative-dynamo-model-resource-generator.ts
```diff
@@ -0,0 +1,260 @@
...
+    return new DynamoDBModelVTLGenerator();
+  }
+
+  protected createModelTable(scope: Construct, def: ObjectTypeDefinitionNode, context: TransformerContextProvider): void {
+    const modelName = def!.name.value;
+    const tableLogicalName = ModelResourceIDs.ModelTableResourceID(modelName);
+    const tableName = context.resourceHelper.generateTableName(modelName);
+
+    // Add parameters.
+    const readIops = new cdk.CfnParameter(scope, ResourceConstants.PARAMETERS.DynamoDBModelTableReadIOPS, {
```
palpatim, line 130: Any opportunity to reuse the code above from `generateResources`?

packages/amplify-graphql-model-transformer/src/resources/iterative-dynamo-model-resource-generator.ts
```diff
@@ -0,0 +1,260 @@
...
+      resourceType: CUSTOM_DDB_CFN_TYPE,
+      properties: {
+        ...defaultTableState,
+      },
+      removalPolicy,
+    });
+    setResourceName(tableResource, { name: modelName, setOnDefaultChild: true });
+
+    // construct a wrapper around the custom table to allow normal CDK operations on top of it
+    const table = Table.fromTableAttributes(scope, `CustomTable${tableLogicalName}`, {
```
palpatim, line 215: We should pick something a bit more descriptive than "CustomTable" -- you refer to it as "AmplifyManagedTable" in the feature flag. That might get a bit long, but maybe "ManagedTable"?

packages/amplify-graphql-model-transformer/src/resources/custom-resource-lambda/custom-resource-handler.ts
```diff
@@ -0,0 +1,440 @@
+import { DynamoDB } from 'aws-sdk';
```
alharris-at, line 1: nit on overall naming. I see you using `custom-resource`, `managed-table`, and `iterative-deployment`. Let's coalesce around a single name scheme for class names, resource names, and filenames (To make it easier to understand how this all works together).

I think I might prefer `DynamicTable` or something like that, to indicate it's not static within the context of the CDK application definition. @palpatim if you have any suggestions too.

packages/amplify-graphql-model-transformer/src/resources/custom-resource-lambda/custom-resource-handler.ts
```diff
@@ -0,0 +1,440 @@
...
+import type { CreateTableInput, KeySchema, Projection, TableDescription, UpdateTableInput } from 'aws-sdk/clients/dynamodb';
+
+const ddbClient = new DynamoDB();
+
+const log = (msg: string, ...other: any[]) => {
```
alharris-at, line 6: Please generally add comments and docs on all the methods here.

packages/amplify-graphql-model-transformer/src/resources/custom-resource-lambda/custom-resource-handler.ts
alharris-at: Discussed offline, and at least temporarily going w/ `AmplifyTable`. here.

packages/amplify-graphql-model-transformer/src/resources/custom-resource-lambda/custom-resource-handler.ts
```diff
@@ -0,0 +1,440 @@
...
+  console.log(
+    msg,
+    other.map((o) => (typeof o === 'object' ? JSON.stringify(o, undefined, 2) : o)),
+  );
+};
+
+export const onEvent = async (
+  event: AWSCDKAsyncCustomResource.OnEventRequest,
+): Promise<AWSCDKAsyncCustomResource.OnEventResponse | void> => {
```
alharris-at, line 15: When do we return a `void` here?

packages/amplify-graphql-model-transformer/src/resources/custom-resource-lambda/custom-resource-handler.ts
```diff
@@ -0,0 +1,440 @@
...
+  for (const key in obj) {
+    if (obj[key] === undefined) {
+      // Use the delete operator to remove the attribute if it's undefined
+      delete obj[key];
+    }
+  }
+  return obj;
+};
+
+const toCreateTableInput = (props: any): CreateTableInput => {
```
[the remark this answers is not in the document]
AaronZyLee: Mark this as TODO. This will be addressed by the latter PR where the `CustomDDB.Input` is well defined. The input will be different from current as the new table construct is in use. (There are some additional property parse process such as using PascalCase, converting table props to CFN properties, etc)

packages/amplify-graphql-model-transformer/src/resources/custom-resource-lambda/custom-resource-handler.ts
```diff
@@ -0,0 +1,440 @@
...
+        },
+      };
+      return result;
+    case 'Delete':
+      console.log('Initiating table deletion');
+      if (!event.PhysicalResourceId) {
+        throw new Error(`Could not find the physical ID for the resource`);
+      }
+      try {
+        await ddbClient.deleteTable({ TableName: event.PhysicalResourceId }).promise();
```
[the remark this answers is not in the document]
AaronZyLee: For the table replacement the confirmation of table being deleted is required as the next step is to create the table again with the exact physical id(table name). For the CFN delete event, there is no follow up steps after the deletion so we just kick off the API call here. If the error is thrown during deletion, the CFN will tell user the resource is failed to delete and let them to delete manually. 

The initial thought is to save the cost as the retry listener will take a lot of lambda execution time. This can be mitigated by adding listener in `isComplete`. Mark this as TODO to address in the latter PR.

alharris-at (commented):
Still reading through the actual lambda code, but dropping these now.

packages/amplify-graphql-model-transformer/src/resources/custom-resource-lambda/custom-resource-handler.ts
AaronZyLee: This `describe` is inside `isComplete`. The `retry` here is to avoid the delay of `Create` event not already populating the new table. 

For the `check for existing state` mentioned, it is in the `onEvent` handler during an `Update` event. The retry is not used but instead a error will be thrown if no table returned in response, which means the table does not exist or the physical Id is not correct, in which case the retry is not needed.

packages/amplify-graphql-model-transformer/src/resources/custom-resource-lambda/custom-resource-handler.ts
AaronZyLee: I think so as there are no warning msgs when I run the lint check

packages/amplify-graphql-model-transformer/src/resources/custom-resource-lambda/custom-resource-handler.ts
AaronZyLee: This type is used by lambda only. What do you mean by the CDK code usage?

packages/amplify-graphql-model-transformer/src/resources/custom-resource-lambda/custom-resource-handler.ts
AaronZyLee: What is the difference here by the `equals` function you mentioned?

packages/amplify-graphql-model-transformer/src/resources/custom-resource-types.d.ts
AaronZyLee: I think this part is to avoid the import of the cdk module. I will try to add it as the dev dependency and see if the cdk module is bundled in the lambda zip. Marked as TODO and will address in next PR.

packages/amplify-graphql-model-transformer/src/resources/custom-resource-types.d.ts
```diff
@@ -0,0 +1,125 @@
+import { DynamoDB } from 'aws-sdk';
```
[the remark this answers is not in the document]
AaronZyLee: It is better to use V3 once the CLI bumps all SDK versions. Mark this as TODO

packages/amplify-graphql-model-transformer/src/resources/custom-resource-lambda/custom-resource-handler.ts
AaronZyLee: When a `Delete` event is processed

AaronZyLee: Removed the void here

packages/amplify-graphql-model-transformer/src/resources/iterative-dynamo-model-resource-generator.ts
AaronZyLee: I refactored them to a method `createDynamoParameters` in the parent class

packages/amplify-graphql-model-transformer/src/resources/amplify-dynamodb-table/amplify-dynamo-model-resource-generator.ts
```diff
@@ -5,22 +5,19 @@ import { ObjectTypeDefinitionNode } from 'graphql';
...
+import { DynamoModelResourceGenerator } from '../dynamo-model-resource-generator';
 import * as path from 'path';
 
 export const ITERATIVE_TABLE_STACK_NAME = 'AmplifyTableManager';
-export const CUSTOM_DDB_CFN_TYPE = 'Custom::AmplifyManagedDynamoDBTable';
+export const CUSTOM_DDB_CFN_TYPE = 'Custom::AmplifyDynamoDBTable';
 /**
- * DynamoModelResourceGenerator is an implementation of ModelResourceGenerator,
- * providing necessary utilities to generate the DynamoDB resources for models
+ * AmplifyDynamoModelResourceGenerator is an subclass of DynamoModelResourceGenerator,
```
palpatim, line 17: nit to incorporate into your next PR (no need to go through the entire review/approval/testing process again on this PR)
```suggestion
 * AmplifyDynamoModelResourceGenerator is a subclass of DynamoModelResourceGenerator,
```

packages/amplify-graphql-model-transformer/src/resources/custom-resource-lambda/custom-resource-handler.ts
palpatim: Doesn't the CDK code that used the custom resource to create the table need the CreateTableResponse (e.g., the table name) in order to complete its work setting up resolvers, etc? If so, is that information delivered in some other structure than the CreateTableResponse?

packages/amplify-graphql-model-transformer/src/resources/custom-resource-lambda/custom-resource-handler.ts
palpatim: No functional difference, it's just a naming thing: traditionally a function that checks for equality would be named to reflect that. Not a big deal now, but if you end up using it elsewhere I'd suggest renaming (and changing the condition check) to make its usage and form more conventional. Feel free to resolve this conversation.

packages/amplify-graphql-model-transformer/src/resources/custom-resource-lambda/custom-resource-handler.ts
```diff
@@ -0,0 +1,440 @@
...
+  if (currentNonKeyAttributes.length !== endNonKeyAttributes.length) return true;
+
+  // if an attribute has been swapped
+  if (currentNonKeyAttributes.some((currentNonKeyAttribute) => !endNonKeyAttributes.includes(currentNonKeyAttribute))) return true;
+
+  // nothing is different
+  return false;
+};
+
+const isKeySchemaModified = (currentSchema: KeySchema, endSchema: KeySchema): boolean => {
```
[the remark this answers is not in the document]
palpatim: Resolving in favor of convo below

packages/amplify-graphql-model-transformer/src/resources/amplify-dynamodb-table/amplify-dynamo-model-resource-generator.ts
```diff
@@ -0,0 +1,192 @@
...
+    this.generateResolvers(ctx);
+  }
+
+  protected createCustomProviderResource(scope: Construct, context: TransformerContextProvider): void {
+    // Policy that grants access to Create/Update/Delete DynamoDB tables
+    const ddbManagerPolicy = new aws_iam.Policy(scope, 'CreateUpdateDeleteTablesPolicy');
+    ddbManagerPolicy.addStatements(
+      new aws_iam.PolicyStatement({
+        actions: ['dynamodb:CreateTable', 'dynamodb:UpdateTable', 'dynamodb:DeleteTable', 'dynamodb:DescribeTable'],
+        // TODO: have more restricted scope
```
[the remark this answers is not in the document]
AaronZyLee: Just added a task in the project

packages/amplify-graphql-model-transformer/src/resources/amplify-dynamodb-table/amplify-dynamo-model-resource-generator.ts
```diff
@@ -0,0 +1,192 @@
...
+      return;
+    }
+
+    if (this.isProvisioned()) {
+      // add model related-parameters to the root stack
+      const rootStack = cdk.Stack.of(ctx.stackManager.scope);
+      this.createDynamoDBParamters(rootStack, false);
+
+      const tableManagerStack = ctx.stackManager.getScopeFor('AmplifyTableCustomProvider', ITERATIVE_TABLE_STACK_NAME);
+      this.createCustomProviderResource(tableManagerStack, ctx);
```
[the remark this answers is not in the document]
AaronZyLee: There are 23 resources generated in the manager stack. I can try to add them into the API root stack in the follow up PR

packages/amplify-graphql-model-transformer/src/resources/amplify-dynamodb-table/amplify-dynamo-model-resource-generator.ts
```diff
@@ -0,0 +1,192 @@
...
+        {
+          AttributeName: 'id',
+          KeyType: 'HASH',
+        },
+      ],
+      StreamSpecification: {
+        StreamViewType: 'NEW_AND_OLD_IMAGES',
+      },
+    };
+    const tableResource = new CustomResource(scope, tableLogicalName, {
```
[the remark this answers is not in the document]
AaronZyLee: The resource logical name stays the same as previous, which is `${modelName}Table`. The resource type is `Custom::AmplifyDynamoDBTable`. Which part are you suggesting the change for?

packages/amplify-graphql-model-transformer/src/resources/amplify-dynamodb-table/amplify-dynamo-model-resource-generator.ts
```diff
@@ -0,0 +1,192 @@
...
+    const removalPolicy = this.options.EnableDeletionProtection ? cdk.RemovalPolicy.RETAIN : cdk.RemovalPolicy.DESTROY;
+
+    // TODO: The following lines for custom resource creation will be wrapped into a new internal construct
+    // TODO: The attribute of encryption and TTL should be added
+    const defaultTableState = {
+      TableName: tableName,
+      AttributeDefinitions: [
+        {
+          AttributeName: 'id',
+          AttributeType: 'S',
```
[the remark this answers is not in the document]
AaronZyLee: Yes. This part will be rewritten in the follow-up PR with the new internal construct. Also the directives such as `@primaryKey`, `@index`, `@hasMany` will be respected

alharris-at (approved):

packages/amplify-graphql-model-transformer/src/resources/custom-resource-lambda/custom-resource-handler.ts
AaronZyLee: There are two steps happening. First is that the create table response is converted to the CFN response in the lambda which is as following
```
      const response = await createNewTable(createTableInput);
      result = {
        PhysicalResourceId: response.tableName,
        Data: {
          TableArn: response.tableArn,
          TableStreamArn: response.streamArn,
          TableName: response.tableName,
        },
      };
```
The next step is that a `Table.fromTableAttributes` is called to construct a table wrapper so that it can be passed into the resolver generation functions. The way to extract the attributes from CFN response is by using intrinsic function `Fn::GetAttr` (https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/crpg-ref-responses.html)
```
    const table = Table.fromTableAttributes(scope, `AmplifyTable${tableLogicalName}`, {
      tableArn: tableResource.getAttString('TableArn'),
      tableStreamArn: tableResource.getAttString('TableStreamArn'),
    });
```

AaronZyLee: The related type here is the `OnEventResponse`. Currently it is more of a generic type. I can add more specific type definitions for the tables in the upcoming PRs.

palpatim (approved):

Issue #1808, opened 2023-08-17
How do I extend Query using when using multiple files in schema/*.graphql?


Zamithal:
### Before opening, please confirm:

- [X] I have checked to see if my question is addressed in the [FAQ](https://github.com/aws-amplify/amplify-hosting/blob/master/FAQ.md).
- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-hosting/issues?q=is%3Aissue+).
- [X] I have read the guide for [submitting questions](https://github.com/aws-amplify/amplify-hosting/blob/master/CONTRIBUTING.md).
- [X] I have removed any sensitive information from my code snippets and submission.

### App Id

d5wysh8nju8x2

### AWS Region

us-west-2

### Amplify Hosting feature

Backend builds

### Question

I have reached the point in my amplify project where having a single `schema.graphql` file is becoming unwieldy. I would like to divide my schema up into multiple files, and have done so:

`<project>/amplify/backend/api/<myApi>/schema/a.graphql` (name is non-ideal name but this causes it to be joined first.)

This contains only a single empty definition for Query:
```graphql
type Query
```
`<project>/amplify/backend/api/<myApi>/schema/post.graphql`

This contains a type and extends Query.
```graphql
enum PostType {
  TYPE_A,
  TYPE_B
}

extend type Query {
  createPost(
    sender: String!,
    subject: String!,
    body: String!,
    type: PostType!
  ): String!
  @auth( rules: [
    { allow: public }
  ])
  @function(name: "createPost-${env}")
```
`<project>/amplify/backend/api/<myApi>/schema/contact.graphql`

 This also contains a type and extends Query.
```graphql
enum ContactReasonType {
  CONTACT_US,
  BETA_SIGN_UP
}

extend type Query {
  sendContactEmail(
    sender: String!,
    subject: String!,
    body: String!,
    type: ContactReasonType!
  ): String!
  @auth( rules: [
    { allow: public }
  ])
  @function(name: "sendContactEmail-${env}")
}
```

Running `amplify api gql-compile` and Including just one of these will successfully compile. Including both of these will result in the following:

`🛑 Object type extension 'Query' cannot redeclare field sendContactEmail`

Using `Query` instead of `extend type Query`results in the following:

`There can be only one type named "Query".`

How can I define Queries when using multiple graphql files?

phani-srikar:
[...]

PR #1863, opened 2023-09-14
fix: iterate over each document definition once

dpilch:
#### Description of changes

Iterate over each document definition once. Previously documents were iterated over `numTotalDocuments - index` times.

This fixes a bug that arises when multiple schema files are used with multiple `extends`.

`schema/a.graphql`
```
type Query {
    queryA: String!
}
```

`schema/b.graphql`
```
extend type Query {
    queryB: String!
}
```

`schema/c.graphql`
```
extend type Query {
    queryC: String!
}
```

The definition in `schema/b.graphql` was processed twice resulting in error: `🛑 Object type extension 'Query' cannot redeclare field queryB`

N/A

#### Issue #, if available

Resolves https://github.com/aws-amplify/amplify-category-api/issues/1808

Unit testing

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-graphql-transformer-core/src/transformer-context/output.ts
```diff
@@ -297,13 +344,17 @@ export class TransformerOutput implements TransformerContextOutputProvider {
...
+    this.ensureSubscription();
     const subscriptionTypeName = this.getSubscriptionTypeName();
     if (subscriptionTypeName) {
       if (!this.getType(subscriptionTypeName)) {
         this.addType(blankObject(subscriptionTypeName));
       }
       let subscriptionType = objectExtension(subscriptionTypeName, fields);
       this.addObjectExtension(subscriptionType);
+    } else {
+      this.addDefaultSubscription();
```
alharris-at, line 356: This should be covered by the call to `this.ensureSubscription()`, right?

dpilch: Oops, yes. This is the original implementation before I changed it use `ensureSubscription`.

AaronZyLee (approved):
Overall LGTM. Could you also double check by `amplify-dev` with the multiple schema files using `extend`? (same as the case in the issue)

alharris-at (approved):

dpilch:
Holding off on merging until construct release.

PR #1865, opened 2023-09-15
feat: refersTo supports field name mappings on RDS models

phani-srikar:
#### Description of changes

- The `field` handling in `refersTo` transformer populates the field mappings for a model in `resourceHelper` present on the transformer context object.
- Other transformers like `index` for index queries, `model` for CRUD operations, `relational` for relational field queries check for those field mappings and add resolver references in the `modelFieldMap` present on the transformer context object.
- These resolver references are finally collected in the `after` step of `refersTo` transformer that adds appropriate `preUpdate` or `preDataLoad` slots to the references resolvers. These slots will set the field mappings in the resolver stash to be passed down to the RDS lambda.
- One side effect of this PR is that for the index queries, we were adding the resolver as `ModelName.IndexQueryName` which is now corrected to `Query.IndexQueryName` in Index and Auth transformers. The generated resolvers should materially be the same.

Added unit and e2e tests. [E2E link](https://us-east-1.console.aws.amazon.com/codesuite/codebuild/594813022831/projects/amplify-category-api-e2e-workflow/batch/amplify-category-api-e2e-workflow%3A0dfea598-3256-406a-815e-b9467d8c3c60?region=us-east-1)

#### Checklist

- [x] PR description included
- [ ] `yarn test` passes


packages/amplify-e2e-tests/src/__tests__/rds-refers-to-fields.test.ts
```diff
@@ -0,0 +1,806 @@
...
+  let host = 'localhost';
+  const identifier = `integtest${db_identifier}`;
+  const projName = 'rdsreferstoapitest2';
+
+  let projRoot;
+  let appSyncClient;
+
+  beforeAll(async () => {
+    projRoot = await createNewProjectDir('rdsreferstoapi2');
+    await setupDatabase();
```
sundersc, line 45: Modify this step to use the VPC configuration and reduce the amount of time the ports are kept open. Refer to the below PR for more details.
https://github.com/aws-amplify/amplify-category-api/pull/1875/files

packages/amplify-graphql-relational-transformer/src/__tests__/relational-auth.test.ts
```diff
@@ -70,7 +71,14 @@ test('ModelXConnection type is getting the directives added, when a field has @h
...
     message: String
   }`;
-  const transformer = getTransformer(withAuthModes(iamDefaultConfig, ['AMAZON_COGNITO_USER_POOLS']));
+  const modelToDatasourceMap = new Map<string, DatasourceType>();
+  ['User', 'Post'].map((modelName) => {
+    modelToDatasourceMap.set(modelName, {
+      dbType: 'DDB',
+      provisionDB: true,
+    });
+  });
```
sundersc, line 80: Is this now required for DDB datasource? Ideally DDB should be default if mapping is not specified.

packages/amplify-graphql-index-transformer/src/__tests__/__snapshots__/amplify-graphql-primary-key-transformer.test.ts.snap
```diff
@@ -33,6 +33,8 @@ $util.qr($ctx.stash.put(\\"keys\\", $keys))
 #set( $lambdaInput.args.metadata = {} )
 #set( $lambdaInput.args.metadata.keys = [] )
 #set( $lambdaInput.args.metadata.nonScalarFields = [] )
+#set( $lambdaInput.args.metadata.fieldMap = {} )
+$util.qr($lambdaInput.args.metadata.fieldMap.putAll($util.defaultIfNull($context.stash.fieldMap, {})))
```
sundersc, line 37: Where is `context.stash.fieldMap` set? Can we add some snapshot to verify & test its datastructure?

phani-srikar: It's set in the `preDataLoad` for queries and `preUpdate` slot for mutations. These resolver slots are generated from the `refersTo` transformer. Will add a snapshot test there.

packages/amplify-graphql-relational-transformer/src/__tests__/relational-auth.test.ts
phani-srikar: It shouldn't be required in the end-to-end experience since we construct this map as part of the `loadSchema`. Since this test doesn't go through that flow, I added these. I'll take another look at this test setup.

packages/amplify-graphql-auth-transformer/src/graphql-auth-transformer.ts
```diff
@@ -410,7 +410,7 @@ export class AuthTransformer extends TransformerAuthBase implements TransformerA
       if (context.metadata.has(indexKeyName)) {
         context.metadata.get<Set<string>>(indexKeyName)!.forEach((index) => {
           const [indexName, indexQueryName] = index.split(':');
-          this.protectListResolver(context, def, def.name.value, indexQueryName, acm, indexName);
+          this.protectListResolver(context, def, 'Query', indexQueryName, acm, indexName);
```
alharris-at, line 413: What's the context on this change?

alharris-at: Or rather, I understand teh change, but don't understand how things work today if that's not changed.

alharris-at (approved):

packages/amplify-graphql-auth-transformer/src/graphql-auth-transformer.ts
phani-srikar: Today the resolvers that are generated from various transformers are managed by the [TransformerResolversManagerProvider](https://github.com/aws-amplify/amplify-category-api/blob/25b65238b5e35e03b1275f432693d8dd6c8fc966/packages/amplify-graphql-transformer-interfaces/src/transformer-context/transformer-resolver-provider.ts#L25) where you can see that we treat `typeName.fieldName` as the key to [add/get any resolver](https://github.com/aws-amplify/amplify-category-api/blob/25b65238b5e35e03b1275f432693d8dd6c8fc966/packages/amplify-graphql-transformer-core/src/transformer-context/resolver.ts#L92). However, if we look at the [resolver interface itself](https://github.com/aws-amplify/amplify-category-api/blob/25b65238b5e35e03b1275f432693d8dd6c8fc966/packages/amplify-graphql-transformer-core/src/transformer-context/resolver.ts#L136), this is where we actually store which typeName, fieldName it corresponds to. For [most other transformers](https://github.com/aws-amplify/amplify-category-api/blob/25b65238b5e35e03b1275f432693d8dd6c8fc966/packages/amplify-graphql-model-transformer/src/resources/model-resource-generator.ts#L136), the typeName, fieldName used in the key matches the ones in the resolver itself, except for index where I found the mis-match. In this PR, we're changing that key for index query resolver which is basically used to get the resolver and append to it's slots etc from the Auth transformer and refersTo transformer.

sundersc (approved):

PR #1867, opened 2023-09-18
fix(graphql): disable default value validation for rds

sundersc:
#### Description of changes

Disable default value validation for RDS datasource. The default value exists to understand the table configuration, we do not use it in our resolvers.

When customer define defaults in the database, they can use function calls or certain database specific syntax to define the defaults (eg: CURRENT_TIMESTAMP for datetime type). In such cases, we do not want to validate or set the default value in the transformers. We will leave it up to the database to set the default value if it is not present in the input.

NA

#### Issue #, if available

NA

- Manual test
- Unit tests

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-graphql-default-value-transformer/src/graphql-default-value-transformer.ts
+++ b/packages/amplify-graphql-default-value-transformer/src/graphql-default-value-transformer.ts
@@ -140,7 +140,6 @@ export class DefaultValueTransformer extends TransformerPluginBase {
     const context = ctx as TransformerContextProvider;
 
     for (const typeName of this.directiveMap.keys()) {
-
       // Set the default value only for DDB datasource. For RDS, the database will set the value.
       const isDynamoDB = isDynamoDBDatasource(ctx, typeName);
       if (!isDynamoDB) {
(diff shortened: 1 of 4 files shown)


packages/amplify-graphql-default-value-transformer/src/graphql-default-value-transformer.ts
```diff
@@ -79,7 +80,18 @@ const validate = (ctx: TransformerSchemaVisitStepContextProvider, config: Defaul
...
+
+  // Validate the default values only for the DynamoDB datasource.
+  // For RDS, the database determines and sets the default value. We will not validate the value in transformers.
+  const isDynamoDB = isDynamoDBDatasource(ctx, config.object.name.value);
+  if (isDynamoDB) {
+    validateDefaultValueType(ctx, config);
+  }
+};
+
+const isDynamoDBDatasource = (ctx: TransformerSchemaVisitStepContextProvider, modelName: string): boolean => {
```
phani-srikar, line 92: nit: This is something we could re-use across transformers if we put this in the transformer core.

phani-srikar (approved):

palpatim (approved):

PR #1875, opened 2023-09-20
test(api): restrict e2e rds instances public access

sundersc:
#### Description of changes

1. Restrict inbound security role added by E2E tests to smaller CIDR range (example: x.y.0.0/16). Also, now the ports are open for a very short time to execute the initial DDLs to create tables. The total time the port kept open has decreased from an average of 15 minutes to less than 5 seconds.

2. Improve SSM read parameter timeout issue. Instead of throwing an exception from wait method, now it returns a string which is used to determine whether to throw an error or not. (This avoids incorrect logging of exception message in the cloudwatch logs).

NA

#### Issue #, if available
NA

- Manual test.
- E2E test in local.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-e2e-tests/src/__tests__/rds-model-v2.test.ts
```diff
@@ -14,28 +14,33 @@ import {
...
 import gql from 'graphql-tag';
+import axios from 'axios';
 
 // to deal with bug in cognito-identity-js
 (global as any).fetch = require('node-fetch');
 
 const CDK_FUNCTION_TYPE = 'AWS::Lambda::Function';
 const CDK_VPC_ENDPOINT_TYPE = 'AWS::EC2::VPCEndpoint';
+const IPIFY_URL = 'https://api.ipify.org/';
+const AWSCHECKIP_URL = 'https://checkip.amazonaws.com/';
```
palpatim, line 33: Under what conditions would we expect these to return different values? Or are you using one as a fallback for the other?

packages/amplify-graphql-model-transformer/rds-lambda/handler.ts
```diff
@@ -44,11 +44,23 @@ const getSSMValue = async (key: string | undefined): Promise<string> => {
...
-  const data = await Promise.race([secretsClient.send(parameterCommand), wait10SecondsAndThrowError()]);
-  if ((data?.$metadata?.httpStatusCode && data?.$metadata?.httpStatusCode >= 400) || !data?.Parameter?.Value) {
+  const data = await Promise.race([secretsClient.send(parameterCommand), wait10Seconds()]);
+
+  // If string is returned, throw error.
+  if (
+    (typeof data === 'string' || data instanceof String) &&
+    data === WAIT_COMPLETE
+  ) {
+    console.log('Unable to retrieve secret for database connection from SSM. If your database is in VPC, verify that you have VPC endpoints for SSM defined and the security group\'s inbound rule for port 443 is defined.');
```
palpatim, line 54:
```suggestion
    console.log('Unable to retrieve secret for database connection from SSM. If your database is in a VPC, verify that you have VPC endpoints for SSM defined and the security group\'s inbound rule for port 443 is defined.');
```

palpatim: Also, it's actually like 4 or 5 endpoints right? We should add a TODO to add a doc link for that so customers can go to setup instructions.

palpatim (commented):
A lot of repetition in the setup code, any chance to centralize & reuse?

packages/amplify-e2e-tests/src/__tests__/rds-model-v2.test.ts
sundersc: IPIFY Url works when I'm not connected to AWS network (or VPN). `checkip` works in AWS network. Having both to allow list enables us to run the tests from anywhere.

packages/amplify-graphql-model-transformer/rds-lambda/handler.ts
sundersc: Since we are using local VPC endpoint URLs, this inbound rule shouldn't be required now. This is meant to be more generic message if we are unable to connect to the SSM for any reason, we have a backlog item to return more meaningful message.

packages/amplify-e2e-tests/src/__tests__/rds-refers-to.test.ts
```diff
@@ -96,11 +96,17 @@ describe('RDS Relational Directives', () => {
...
+    ipAddresses.push(...(await getIpRanges()));
+    await Promise.all(
+      ipAddresses.map((ip) =>
+        addRDSPortInboundRule({
+          region,
+          port: db.port,
+          cidrIp: ip,
+        }),
+      ),
+    );
```
phani-srikar, line 109: nit: since we're doing similar setup for multiple tests, we could consider adding another method to the utils like `addRDSInboundRules(region)` which computes the ipAddresses and adds inbound rule for each of them. Similarly for outbound rules as well. `port` is 3306 in all tests, so it could either be an optional to those methods or skip.

sundersc: We need to store the IP address here so that we can use the same values for remove rule. We can retrieve IP address independent of each function call as it could change in between these 2 invocations.

sundersc:
> A lot of repetition in the setup code, any chance to centralize & reuse?

Addressed it. Moved the complete database setup and port opening logic to the E2E core.

phani-srikar (approved):

packages/amplify-e2e-tests/src/__tests__/rds-refers-to.test.ts
```diff
@@ -45,9 +39,9 @@ describe('RDS Relational Directives', () => {
 
   beforeAll(async () => {
     projRoot = await createNewProjectDir('rdsreferstoapi1');
-    await setupDatabase();
     await initProjectAndImportSchema();
     await amplifyPush(projRoot);
+    await sleep(2 * 60 * 1000); // Wait for 2 minutes for the VPC endpoints to be live.
```
palpatim, line 44: Does [`DescribeVpcEndpoints`](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeVpcEndpoints.html) correctly report status? I'd love not to spend 2 minutes in every test waiting for an endpoint to get created

sundersc: Based on last few E2E test runs, it is observed that it takes about 20-30 seconds for the VPC endpoints to be active even though the status shows 'AVAILABLE'. This delay was added to avoid flaky tests.

packages/amplify-e2e-core/src/utils/rds.ts
```diff
@@ -268,3 +323,13 @@ export const getResource = (resources: Map<string, any>, resourcePrefix: string,
   }
   return undefined;
 };
+
+export const getIpRanges = async (): Promise<string[]> => {
+  return Promise.all(
+    [IPIFY_URL, AWSCHECKIP_URL].map(async (url) => {
```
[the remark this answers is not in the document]
sundersc: Are you connected to VPN? If yes, then the IPs would differ. We can't create a duplicate rule. If there is already an entry, the second API call would fail and we ignore the error.

palpatim (approved):

PR #1883, opened 2023-09-23
Field mapping schema edits

phani-srikar:
#### Description of changes

- Infer and add field name mappings based on the column names of a table in MySQL. This happens when customers import RDS schema.
- Retain any edits to the schema with respect to field name mappings when customers re-generate their RDS schema.
- Throw error during re-generation when multiple models map to same tableName in the edited schema.

Added unit and E2E tests. [E2E link](https://us-east-1.console.aws.amazon.com/codesuite/codebuild/594813022831/projects/amplify-category-api-e2e-workflow/batch/amplify-category-api-e2e-workflow%3Ac44d815c-2032-40c7-85ca-770a82520698?region=us-east-1)

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-e2e-tests/src/__tests__/rds-model-v2.test.ts
+++ b/packages/amplify-e2e-tests/src/__tests__/rds-model-v2.test.ts
@@ -125,10 +125,10 @@ describe('RDS Model Directive', () => {
       region,
     };
     const queries = [
-      'CREATE TABLE Contact (id VARCHAR(40) PRIMARY KEY, FirstName VARCHAR(20), LastName VARCHAR(50))',
-      'CREATE TABLE Person (personId INT PRIMARY KEY, FirstName VARCHAR(20), LastName VARCHAR(50))',
-      'CREATE TABLE Employee (ID INT PRIMARY KEY, FirstName VARCHAR(20), LastName VARCHAR(50))',
-      'CREATE TABLE Student (studentId INT NOT NULL, classId CHAR(1) NOT NULL, FirstName VARCHAR(20), LastName VARCHAR(50), PRIMARY KEY (studentId, classId))',
+      'CREATE TABLE Contact (id VARCHAR(40) PRIMARY KEY, firstName VARCHAR(20), lastName VARCHAR(50))',
+      'CREATE TABLE Person (personId INT PRIMARY KEY, firstName VARCHAR(20), lastName VARCHAR(50))',
+      'CREATE TABLE Employee (id INT PRIMARY KEY, firstName VARCHAR(20), lastName VARCHAR(50))',
+      'CREATE TABLE Student (studentId INT NOT NULL, classId CHAR(1) NOT NULL, firstName VARCHAR(20), lastName VARCHAR(50), PRIMARY KEY (studentId, classId))',
     ];
 
     const db = await setupRDSInstanceAndData(dbConfig, queries);
@@ -181,8 +181,8 @@ describe('RDS Model Directive', () => {
 
     // Verify the fields in the generated schema on type 'Contacts'
     const contactsIdFieldType = contactObjectType.fields.find((f) => f.name.value === 'id');
-    const contactsFirstNameFieldType = contactObjectType.fields.find((f) => f.name.value === 'FirstName');
-    const contactsLastNameFieldType = contactObjectType.fields.find((f) => f.name.value === 'LastName');
+    const contactsFirstNameFieldType = contactObjectType.fields.find((f) => f.name.value === 'firstName');
+    const contactsLastNameFieldType = contactObjectType.fields.find((f) => f.name.value === 'lastName');
 
     expect(contactsIdFieldType).toBeDefined();
     expect(contactsFirstNameFieldType).toBeDefined();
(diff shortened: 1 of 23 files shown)


packages/amplify-e2e-tests/src/__tests__/rds-model-v2.test.ts
```diff
@@ -125,10 +125,10 @@ describe('RDS Model Directive', () => {
       region,
     };
     const queries = [
-      'CREATE TABLE Contact (id VARCHAR(40) PRIMARY KEY, FirstName VARCHAR(20), LastName VARCHAR(50))',
-      'CREATE TABLE Person (personId INT PRIMARY KEY, FirstName VARCHAR(20), LastName VARCHAR(50))',
-      'CREATE TABLE Employee (ID INT PRIMARY KEY, FirstName VARCHAR(20), LastName VARCHAR(50))',
-      'CREATE TABLE Student (studentId INT NOT NULL, classId CHAR(1) NOT NULL, FirstName VARCHAR(20), LastName VARCHAR(50), PRIMARY KEY (studentId, classId))',
+      'CREATE TABLE Contact (id VARCHAR(40) PRIMARY KEY, firstName VARCHAR(20), lastName VARCHAR(50))',
```
phani-srikar, line 128: these formatting on existing setup is to ensure that these tests actually test what they're intending to w/o field mappings interfering. We have separate E2Es added for field mappings.

packages/amplify-e2e-tests/src/__tests__/rds-v2-generate-schema.test.ts
```diff
@@ -74,9 +74,10 @@ describe('RDS Generate Schema tests', () => {
       region,
     };
     const queries = [
-      'CREATE TABLE Contact (ID INT PRIMARY KEY, FirstName VARCHAR(20), LastName VARCHAR(50))',
-      'CREATE TABLE Person (ID INT PRIMARY KEY, Info JSON NOT NULL)',
-      'CREATE TABLE tbl_todos (ID INT PRIMARY KEY, description VARCHAR(20))',
+      'CREATE TABLE Contact (id INT PRIMARY KEY, firstName VARCHAR(20), lastName VARCHAR(50))',
+      'CREATE TABLE Person (id INT PRIMARY KEY, info JSON NOT NULL)',
+      'CREATE TABLE tbl_todos (id INT PRIMARY KEY, description VARCHAR(20))',
+      'CREATE TABLE Task (Id INT PRIMARY KEY, Description VARCHAR(20), task_name VARCHAR(20))',
```
sundersc, line 80: Why are we changing the field names from PascalCase to camelCase?

packages/amplify-e2e-tests/src/__tests__/rds-v2-generate-schema.test.ts
```diff
@@ -74,9 +74,10 @@ describe('RDS Generate Schema tests', () => {
       region,
     };
     const queries = [
-      'CREATE TABLE Contact (ID INT PRIMARY KEY, FirstName VARCHAR(20), LastName VARCHAR(50))',
-      'CREATE TABLE Person (ID INT PRIMARY KEY, Info JSON NOT NULL)',
-      'CREATE TABLE tbl_todos (ID INT PRIMARY KEY, description VARCHAR(20))',
+      'CREATE TABLE Contact (id INT PRIMARY KEY, firstName VARCHAR(20), lastName VARCHAR(50))',
```
sundersc, line 77: Why are we changing the field names from PascalCase to camelCase?

phani-srikar: Like I mentioned in the above comment, some of the existing tests have table setups that will now be generated with a `refersTo` if they are not camel cased which increases the scope of what they're testing. So I added additional tests for `refersTo` on fields cases and kept the existing ones simpler.

packages/amplify-graphql-schema-generator/src/__tests__/schema-overrides.test.ts
```diff
@@ -397,6 +398,229 @@ describe('apply schema overrides for models with refersTo', () => {
...
+                  name: String @refersTo(name: "nameField")
+              }
+          `;
+    const editedDocument = parse(editedSchema);
+    const updatedDocument = applySchemaOverrides(document, editedDocument);
+    stringsMatchWithoutWhitespace(print(updatedDocument), editedSchema);
+  });
+
+  it('should retain other directives on mapped fields', () => {
+    const documents = [
```
sundersc, line 505: Instead of relying on array index order, can we be more explicit using key value pair?

packages/graphql-transformer-common/package.json
```diff
@@ -36,8 +36,8 @@
     "collectCoverage": true,
     "coverageThreshold": {
       "global": {
-        "branches": 30,
-        "functions": 19,
+        "branches": 14,
```
sundersc, line 39: It is a significant coverage change. Can we improve it?

packages/graphql-transformer-common/src/__tests__/defnition.test.ts
```diff
@@ -31,28 +31,3 @@ describe('gets Non-Model Types', () => {
     expect((nonModelTypes[0] as ObjectTypeDefinitionNode).name?.value).toEqual('NonModel');
   });
 });
-
-describe('finds matching field', () => {
```
sundersc, line 35: Are these tests moved to a different package?

packages/amplify-graphql-schema-generator/src/__tests__/schema-overrides.test.ts
```diff
@@ -397,6 +398,229 @@ describe('apply schema overrides for models with refersTo', () => {
     const updatedDocument = applySchemaOverrides(document, editedDocument);
     stringsMatchWithoutWhitespace(print(updatedDocument), editedSchema);
   });
+
+  it('should not allow duplicate model name mappings in edited schema', () => {
```
sundersc, line 402: Can we add a test case when refersTo is used on a model which is used as output type on a custom query using @sql directive?

palpatim: To pile onto this: we should have a positive case (refersTo maps to an output type that doesn't conflict) as well as failure case (refersTo maps to an output type that conflicts)

packages/amplify-graphql-schema-generator/src/__tests__/schema-overrides.test.ts
```diff
@@ -397,6 +398,229 @@ describe('apply schema overrides for models with refersTo', () => {
...
+                  id: ID! @refersTo(name: "my_id")
+                  name_field: String
+              }
+          `;
+    const editedDocument = parse(editedSchema);
+    const updatedDocument = applySchemaOverrides(document, editedDocument);
+    stringsMatchWithoutWhitespace(print(updatedDocument), editedSchema);
+  });
+
+  it('should allow adding refersTo to fields', () => {
```
palpatim, line 486: Can you add a test for field level conflicts (2 GraphQL fields referring to the same SQL field)

packages/amplify-graphql-schema-generator/src/schema-generator/generate-schema.ts
```diff
@@ -370,6 +380,14 @@ export const convertToGraphQLTypeName = (modelName: string): string => {
   return singular(toPascalCase(cleanedInput?.split('_')));
 };
 
+export const convertToGraphQLFieldName = (fieldName: string): string => {
```
palpatim, line 383: What happens if the incoming name begins with a number?

packages/graphql-transformer-common/src/__tests__/defnition.test.ts
phani-srikar: Yes, you can find them here: https://github.com/aws-amplify/amplify-category-api/blob/cb370f2fbcbbedc077483d9b66414ea2119577be/packages/amplify-graphql-schema-generator/src/__tests__/schema-overrides.test.ts#L600

packages/graphql-transformer-common/package.json
phani-srikar: Added additional tests to the method I updated and one more similar one. Bumping this number all at once takes more time, will do it incrementally as we update this package.

packages/amplify-graphql-schema-generator/src/schema-generator/generate-schema.ts
phani-srikar: Updated to trim non-alphabetic characters from beginning of the field name.

packages/amplify-graphql-schema-generator/src/__tests__/schema-overrides.test.ts
```diff
@@ -397,6 +398,308 @@ describe('apply schema overrides for models with refersTo', () => {
...
+              id: String! @primaryKey
+              name: String
+              profile: Profile @hasOne(references: "userId")
+        }`;
+    const editedDocument = parse(editedSchema);
+    const updatedDocument = applySchemaOverrides(document, editedDocument);
+    stringsMatchWithoutWhitespace(print(updatedDocument), editedSchema);
+  });
+
+  it('should retain relational directives', () => {
```
palpatim, line 614: Please rename this test -- you're asserting the error condition, and it wasn't until I read the assertion at the bottom that I realized you weren't actually going to allow a refersTo on a relational field :)

packages/graphql-transformer-common/package.json
```diff
@@ -36,9 +36,9 @@
     "collectCoverage": true,
     "coverageThreshold": {
       "global": {
-        "branches": 30,
-        "functions": 19,
-        "lines": 30
+        "branches": 17,
```
palpatim, line 39: Per your convo https://github.com/aws-amplify/amplify-category-api/pull/1883#discussion_r1339237643 , are we committing to raising this before release? I'm pretty uncomfortable with the drop otherwise.

packages/amplify-graphql-schema-generator/src/schema-generator/generate-schema.ts
palpatim: Please see new comment about leading underscore

packages/amplify-graphql-schema-generator/src/__tests__/generate-schema.test.ts
```diff
@@ -37,6 +51,55 @@ describe('Type name conversions', () => {
...
+  it('GraphQL idiomatic field name conversions', () => {
+    // leave as-is
+    expect(convertToGraphQLFieldName('posts')).toEqual('posts');
+    expect(convertToGraphQLFieldName('postscolumn')).toEqual('postscolumn');
+    // Camel case
+    expect(convertToGraphQLFieldName('employees_salary')).toEqual('employeesSalary');
+
+    // Remove special characters not supported in GraphQL
+    expect(convertToGraphQLFieldName('Employees_salaries-Column')).toEqual('employeesSalariesColumn');
+    expect(convertToGraphQLFieldName('_employee_Salaries-column')).toEqual('employeeSalariesColumn');
```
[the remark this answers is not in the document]
phani-srikar: underscore prefix in GraphQL is commonly used for server added/internal fields but in MySQL customers use it for wide variety of reasons that are specific to their use-case. [some examples in comments](https://www.sqlservercentral.com/forums/topic/column-name-starting-with-_-underscore). If we retain these without knowing the underlying use-case, we would falsely imply these as internal fields. After initial generation, customers can choose to keep them by making explicit edits based on their use-case, and we preserve them.

packages/graphql-transformer-common/package.json
phani-srikar: The **effective** change in this package is a single line which I added tests for, not really sure what's causing this drop, but will add a task for this.

palpatim (approved):

Issue #907, opened 2022-10-20
OpenSearch instances should be created with amplify env suffix


hisham:
### Is this feature request related to a new or existing Amplify category?

api

### Is this related to another service?

OpenSearch

### Describe the feature you'd like to request

Currently opensearch instances are named randomly (e.g. amplify-opense-foobar). You should include the env name in this name to help us easily distinguish which opensearch instance is for which amplify environment. 

![CleanShot 2022-10-20 at 12 42 57@2x](https://user-images.githubusercontent.com/81853/197031800-8e0de907-d1e0-47d2-be5c-79fefb27e697.png)

### Describe the solution you'd like

You should include the env name in this name to help us easily distinguish which opensearch instance is for which amplify environment. 

### Describe alternatives you've considered

none

### Is this something that you'd be interested in working on?

- [ ] 👋 I may be able to implement this feature request

### Would this feature include a breaking change?

- [ ] ⚠️ This feature might incur a breaking change

AnilMaktala:
We should also include the Project - API - Environment in the search domain name if the character limit supports all three in the domain name.

PR #1916, opened 2023-10-04
fix(graphql-searchable-transformer): formatting domain name

snsarij:
#### Description of changes
- Created a new resource helper function that will generate the domain name for open search domain
- Referred to this function in the CDK
- Completed the local testing

No parameters have been changed

#### Issue #, if available
Solves the issue: https://github.com/aws-amplify/amplify-category-api/issues/907

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-graphql-searchable-transformer/src/cdk/create-searchable-domain.ts
+++ b/packages/amplify-graphql-searchable-transformer/src/cdk/create-searchable-domain.ts
@@ -2,7 +2,7 @@ import { TransformerContextProvider } from '@aws-amplify/graphql-transformer-int
 import { EbsDeviceVolumeType } from 'aws-cdk-lib/aws-ec2';
 import { CfnDomain, Domain, ElasticsearchVersion } from 'aws-cdk-lib/aws-elasticsearch';
 import { IRole, Role, ServicePrincipal } from 'aws-cdk-lib/aws-iam';
-import { CfnParameter, Fn, RemovalPolicy } from 'aws-cdk-lib';
+import { CfnParameter, RemovalPolicy } from 'aws-cdk-lib';
 import { Construct } from 'constructs';
 import { ResourceConstants } from 'graphql-transformer-common';
 import { setResourceName } from '@aws-amplify/graphql-transformer-core';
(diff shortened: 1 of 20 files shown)


dpilch:
I think the change makes sense to me. I'll need to investigate if this could have ill effects on existing apps.

There is a failing test on the mock util. `cd packages/amplify-util-mock` then `yarn e2e` to see the failure.

Also, can you re-run `yarn extract-api`? Seems to just be a formatting issue.

AaronZyLee:
The domain name is an attribute which requires replacement, during which process the new search domain with the given name is generated and the old one being deleted as follow up. This is definitely a breaking change for the existing users. Thus this PR will be held until a migration plan is figured out. 

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-opensearchservice-domain.html#cfn-opensearchservice-domain-domainname

As a workaround for the linked issue, the `amplify override api` can be used to change the naming of search domain. The example `override.ts` is as following.

```TypeScript
import { AmplifyApiGraphQlResourceStackTemplate, AmplifyProjectInfo } from '@aws-amplify/cli-extensibility-helper';

export function override(resources: AmplifyApiGraphQlResourceStackTemplate, amplifyProjectInfo: AmplifyProjectInfo) {
  resources.opensearch.OpenSearchDomain.domainName = `amplify-opensearch-${amplifyProjectInfo.envName}`;
}
```

packages/amplify-graphql-searchable-transformer/src/cdk/create-searchable-domain.ts
```diff
@@ -29,7 +28,7 @@ export const createSearchableDomain = (
     zoneAwareness: {
       enabled: false,
     },
-    domainName: Fn.conditionIf(HasEnvironmentParameter, Fn.ref('AWS::NoValue'), `d${apiId}`).toString(),
+    domainName: context.resourceHelper.generateDomainName(),
```
sundersc, line 31: Changing the domain name may be destructive for the existing apps. We need to verify that this change result in loss of data.

PR #1919, opened 2023-10-04
fix: iam auth values can be passed as cdk tokens

alharris-at:
#### Description of changes
A shortcoming of the transformer generated resolvers was identified. If any token value is passed into a generated file, then the VTL will be invalid, and will fail to execute.
By shifting these CDK token-able values into the stash (which makes sense, since they're config anyway) we end up allowing those to work, the before/after templates on pipeline transformers are inline in the Cloudformation, so this is fine.
Then we just need to reference the stash values instead of injecting directly into the pipeline stages themselves.

I've added and updated 2 integration tests to help identify that we're generating valid resolvers and stacks, and moved the adminRoles and identityPoolId into the `synthParameters` so they're available during resolver generation.

Resolver templates will be updated.

#### Issue #, if available
N/A

Unit + E2E Tests, and manual testing w/ a CDK app on the packaged tarball.

https://us-east-1.console.aws.amazon.com/codesuite/codebuild/594813022831/projects/amplify-category-api-e2e-workflow/batch/amplify-category-api-e2e-workflow%3A0a82da8d-26f5-44b9-a11d-4220180d04e2?region=us-east-1

#### Checklist
- [x] PR description included
- [x] `yarn test` passes


packages/amplify-graphql-auth-transformer/src/__tests__/resolvers/subscriptions.test.ts
```diff
@@ -105,33 +104,15 @@ describe('subscriptions', () => {
       });
 
       it('renders for iam auth with no admin roles', () => {
-        const adminRolesDisabledExpr1 = generateAuthExpressionForSubscriptions(
-          configFromPartial({
```
alharris-at, line 109: These three parameterized ways of representing admin roles (which we were verifying had the same effect here) are now collapsed into one, so the comparison between them is simplified.

packages/amplify-graphql-auth-transformer/src/__tests__/__snapshots__/amplify-admin-auth.test.ts.snap
```diff
@@ -4,8 +4,7 @@ exports[`admin roles should be return the field name inside field resolvers 1`]
 "## [Start] Field Authorization Steps. **
 #set( $isAuthorized = false )
 #if( $util.authType() == \\"IAM Authorization\\" )
-  #set( $adminRoles = [\\"us-fake-1_uuid_Full-access/CognitoIdentityCredentials\\",\\"us-fake-1_uuid_Manage-only/CognitoIdentityCredentials\\"] )
```
AaronZyLee, line 7: Is it possible to add unit tests to verify the admin roles are added to stash as expected since they are being removed from old snapshots? Looks like they are moved to the inline statement of `before` stage.

alharris-at: I've updated and added positive/negative stash checks to these units.

alharris-at:
I need to dig in more tomorrow, but this may not work with field-level resolvers, since they're maybe not pipelines? I need to double check that, but it's leading to the mock tests to fail (but that could just be a bad test rig).

AaronZyLee (approved):

packages/amplify-graphql-api-construct-tests/src/__tests__/backends/all-auth-modes/app.ts
```diff
@@ -16,8 +16,7 @@ const stack = new Stack(app, packageJson.name.replace(/_/g, '-'), {
 });
 
 const userPool = new UserPool(stack, 'TestPool', {});
-// const identityPoolId = new CfnIdentityPool(stack, 'TestIdentityPool', { allowUnauthenticatedIdentities: true }).attrId;
-const identityPoolId = 'us-west-2:54ee61c9-8d51-46aa-8898-33e8578d2f75'; // Random id, value not used in this test
+const identityPoolId = new CfnIdentityPool(stack, 'TestIdentityPool', { allowUnauthenticatedIdentities: true }).logicalId;
```
palpatim, line 19: Just making super sure this is supposed to be `logicalId` not `physicalId` or `attrId` of the construct? Logical IDs are allowed to be overridden, which I'm not expecting for the actual resolved Identity Pool ID. (Basically I was assuming that you'd be using the physical id)

packages/amplify-graphql-auth-transformer/src/resolvers/helpers.ts
```diff
@@ -173,16 +174,11 @@ export const iamExpression = (
...
-  fieldName?: string,
-  adminCheckExpression?: Expression,
-): Expression => {
+export const iamAdminRoleCheckExpression = (fieldName?: string, adminCheckExpression?: Expression): Expression => {
   const returnStatement = fieldName ? raw(`#return($context.source.${fieldName})`) : raw('#return($util.toJson({}))');
   const fullReturnExpression = adminCheckExpression ? compoundExpression([adminCheckExpression, returnStatement]) : returnStatement;
   return compoundExpression([
-    set(ref('adminRoles'), raw(JSON.stringify(adminRoles))),
-    forEach(/* for */ ref('adminRole'), /* in */ ref('adminRoles'), [
+    forEach(/* for */ ref('adminRole'), /* in */ ref('ctx.stash.adminRoles'), [
```
palpatim, line 181: OOC, why the commented out for/in?

packages/amplify-graphql-transformer/src/graphql-transformer.ts
```diff
@@ -176,8 +176,15 @@ export const executeTransform = (config: ExecuteTransformConfig): void => {
     assetProvider,
     synthParameters,
     parameterProvider,
+    transformersFactoryArgs,
   } = config;
 
+  const mergedSynthParameters = {
```
palpatim, line 182: Since I don't fully understand the construct arch I'm confused why we define the roles & IDs as both synth params and this factory args override?

packages/amplify-graphql-api-construct-tests/src/__tests__/backends/all-auth-modes/app.ts
alharris-at: Ahh yeah, this should actually be `ref` which I tested locally, but haven't set here (just to make the test accurate).

packages/amplify-graphql-auth-transformer/src/resolvers/helpers.ts
alharris-at: I think just readability.

palpatim (approved):

AaronZyLee (approved):

motravo:
I'm seeing some broken auth starting in 12.8.2 that I think are related to this change. Filed #2060 with details.

PR #1920, opened 2023-10-04
test: add userpool, OIDC and custom claims auth tests for RDS (MySQL)

phani-srikar:
#### Description of changes

Contains following fixes:
- Auth provider value correctly set in the resolver.
- Added missing field mapping to the `GET` call in Auth resolver.
- Should not set `idenitityClaim` for default cognito multi-claim.
- Use `groupClaim` instead of `identiyClaim` for Group Auth in resolver.
- Prevent models without primary key from compiling, to avoid leaking unwanted records to the customers.

Added the E2E tests and test helpers for following Auth scenarios in RDS:
- Cognito user pools as Auth provider.
- OIDC as Auth provider.
- RefersTo and Custom claims with Cognito user pools.
- [Test Helper] Generate table DDL from schema definitions. This will be improved over time as we expand our tests.
- [Test Helper] Create model operation helpers based on the schema.

One callout is that between the test cases are similar b/w userpools and OIDC. Although I have abstracted out few things into helpers, there is still room for refinement, that we can refactor later.

E2E tests pass locally. Will run these on CI and post any fixes necessary when the CI account is setup.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-graphql-model-transformer/src/graphql-model-transformer.ts
+++ b/packages/amplify-graphql-model-transformer/src/graphql-model-transformer.ts
@@ -176,6 +176,16 @@ export class ModelTransformer extends TransformerModelBase implements Transforme
         `'${definition.name.value}' is a reserved type name and currently in use within the default schema element.`,
       );
     }
+
+    if (!isDynamoDB) {
+      const containsPrimaryKey = definition.fields?.some((field) => {
+        return field?.directives?.some((directive) => directive.name.value === 'primaryKey');
+      });
+      if (!containsPrimaryKey) {
+        throw new InvalidDirectiveError(`RDS model "${definition.name.value}" must contain a primary key field`);
+      }
+    }
+
     // todo: get model configuration with default values and store it in the map
     const typeName = definition.name.value;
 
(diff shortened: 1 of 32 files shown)


palpatim (approved):

packages/amplify-graphql-model-transformer/src/graphql-model-transformer.ts
```diff
@@ -176,6 +176,16 @@ export class ModelTransformer extends TransformerModelBase implements Transforme
...
     }
+
+    if (!isDynamoDB) {
+      const containsPrimaryKey = definition.fields?.some((field) => {
+        return field?.directives?.some((directive) => directive.name.value === 'primaryKey');
+      });
+      if (!containsPrimaryKey) {
+        throw new InvalidDirectiveError(`RDS model "${definition.name.value}" must contain a primary key field`);
+      }
+    }
```
alharris-at, line 187: Instead of pivoting on !isDynamoDB, can we have the generator per-platform have platform-specific `validate` methods which we invoke?

alharris-at (approved):

packages/amplify-graphql-model-transformer/src/graphql-model-transformer.ts
phani-srikar: That's a good idea. We only have platform specific generators now, will take this as future enhancement. Thanks.

PR #1937, opened 2023-10-10
fix: convert single action to list of actions

dpilch:
#### Description of changes

Sometimes the policies have a single action as a string instead of a list of string. I'm not sure what cases this happens in. When it is a string `iam.PolicyStatement` will split the string as if it were an array resulting in the following error.

```
Action 's' is invalid. An action string consists of a service namespace, a colon, and the name of an action. Action names can include wildcards.
```

#### Issue #, if availabl

N/A

Unit testings

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-category-api/src/__tests__/provider-utils/awscloudformation/ecs-apigw-stack.test.ts
+++ b/packages/amplify-category-api/src/__tests__/provider-utils/awscloudformation/ecs-apigw-stack.test.ts
@@ -59,6 +59,10 @@ describe('ecs stack', () => {
             },
           ],
         },
+        {
+          Effect: 'Allow',
+          Action: 's3:ListBucket',
+        },
       ],
     });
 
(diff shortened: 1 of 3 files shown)


packages/amplify-category-api/src/__tests__/provider-utils/awscloudformation/ecs-apigw-stack.test.ts
```diff
@@ -59,6 +59,10 @@ describe('ecs stack', () => {
             },
           ],
         },
+        {
+          Effect: 'Allow',
+          Action: 's3:ListBucket',
```
dpilch, line 64: This will cause the test to throw an error without the change to `jsonPolicyToCdkPolicyStatement`.

alharris-at (approved):

PR #1938, opened 2023-10-10
feat: add support for a model introspection provider as an input property

alharris-at:
#### Description of changes
Adding a new input parameter where a 'modelIntrospectionSchemaProvider' can be passed in. This is a workaround for the fact that bundling deps is quite painful, and after spending some time on it here, we still weren't able to get the right matching deps for the full appsync-modelgen-plugin library. Instead, rely on that being passed in from the client use-location.

N/A

#### Issue #, if available
N/A

Unit + Functional tests on the changes and the outputs.

#### Checklist
- [x] PR description included
- [x] `yarn test` passes


alharris-at:
Note: this may not end up working in any capacity. We will end up needing to invoke an async method for the injected codegen, so it'll need to pass in a `Promise<string>` though this won't work since we do all this work in the `ctor`. We may need some additional method to asynchronously build the meta.

PR #1957, opened 2023-10-16
test(api): e2e for iam provider on rds models

sundersc:
Adding E2E tests for Auth Directive IAM provider support.
- Covers IAM public and private rules. (Public rule strictly gives access to unauthRule and private rule gives access to the logged in users).
- Mutli-auth scenarios with API key, lambda and IAM for queries, mutations and subscriptions.

### DIFF
--- a/packages/amplify-e2e-core/src/init/amplifyPush.ts
+++ b/packages/amplify-e2e-core/src/init/amplifyPush.ts
@@ -32,9 +32,13 @@ export type LayerPushSettings = {
 /**
  * Function to test amplify push with verbose status
  */
-export function amplifyPush(cwd: string, testingWithLatestCodebase = false, settings?: {
-  skipCodegen?: boolean;
-}): Promise<void> {
+export function amplifyPush(
+  cwd: string,
+  testingWithLatestCodebase = false,
+  settings?: {
+    skipCodegen?: boolean;
+  },
+): Promise<void> {
   return new Promise((resolve, reject) => {
     // Test detailed status
     spawn(getCLIPath(testingWithLatestCodebase), ['status', '-v'], { cwd, stripColors: true, noOutputTimeout: pushTimeoutMS })
@@ -50,20 +54,20 @@ export function amplifyPush(cwd: string, testingWithLatestCodebase = false, sett
       .sendConfirmYes();
 
     if (!settings?.skipCodegen) {
-      pushCommands.wait('Do you want to generate code for your newly created GraphQL API')
+      pushCommands
+        .wait('Do you want to generate code for your newly created GraphQL API')
         .sendConfirmNo()
         .wait('Do you want to generate code for your newly created GraphQL API')
         .sendConfirmNo();
     }
 
-    pushCommands.wait(/.*/)
-      .run((err: Error) => {
-        if (!err) {
-          resolve();
-        } else {
-          reject(err);
-        }
-      });
+    pushCommands.wait(/.*/).run((err: Error) => {
+      if (!err) {
+        resolve();
+      } else {
+        reject(err);
+      }
+    });
   });
 }
 
--- a/packages/amplify-e2e-core/src/init/amplifyPush.ts
+++ b/packages/amplify-e2e-core/src/init/amplifyPush.ts
@@ -32,7 +32,9 @@ export type LayerPushSettings = {
 /**
  * Function to test amplify push with verbose status
  */
-export function amplifyPush(cwd: string, testingWithLatestCodebase = false): Promise<void> {
+export function amplifyPush(cwd: string, testingWithLatestCodebase = false, settings?: {
+  skipCodegen?: boolean;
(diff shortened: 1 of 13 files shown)


packages/amplify-graphql-auth-transformer/src/vtl-generator/rds/resolvers/common.ts
```diff
@@ -95,7 +95,7 @@ const convertAuthRoleToVtl = (
           type: str(role.strategy),
           provider: str('iam'),
           roleArn: role.strategy === 'public' ? ref('ctx.stash.unauthRole') : ref('ctx.stash.authRole'),
-          cognitoIdentityPoolId: hasIdentityPoolId ? ref('ctx.stash.identityPoolId') : nul(),
+          ...(role.strategy === 'private' && { cognitoIdentityPoolId: identityPoolId ? str(identityPoolId) : nul() }),
```
alharris-at, line 98: You need to reference the identityPoolId in stash, right? I don't think that value `identityPoolId` you're referencing exists. This should fail to build if I'm not mistaken.

alharris-at: Yeah, just double-checked, and it looks like the build is failing on this.

packages/amplify-e2e-core/src/init/amplifyPush.ts
```diff
@@ -43,12 +45,18 @@ export function amplifyPush(cwd: string, testingWithLatestCodebase = false): Pro
...
+    const pushCommands = spawn(getCLIPath(testingWithLatestCodebase), ['push'], { cwd, stripColors: true, noOutputTimeout: pushTimeoutMS })
       .wait('Are you sure you want to continue?')
-      .sendConfirmYes()
-      .wait('Do you want to generate code for your newly created GraphQL API')
-      .sendConfirmNo()
-      .wait(/.*/)
+      .sendConfirmYes();
+
+    if (!settings?.skipCodegen) {
+      pushCommands.wait('Do you want to generate code for your newly created GraphQL API')
```
phani-srikar, line 53: There is an existing method already for this called `amplifyPushWithoutCodegen` that's used in other tests.

packages/amplify-graphql-auth-transformer/src/vtl-generator/rds/resolvers/common.ts
sundersc: Ah, looks rebase didn't go well. Looking into it.

packages/amplify-e2e-tests/src/utils/api.ts
```diff
@@ -0,0 +1,52 @@
+import { API } from 'aws-amplify';
+
+export const reconfigureAmplifyAPI = (url: string, region: string, appSyncAuthType: string, apiKey?: string): void => {
+  if (appSyncAuthType === 'API_KEY') {
```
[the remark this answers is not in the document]
sundersc: No strong preference here. Like the more explicit check on the auth type here.

packages/amplify-e2e-core/src/init/amplifyPush.ts
sundersc: Not sure why we have a separate method for these minor differences. Still I feel like this is a good addition as we can eventually get rid of the `amplifyPushWithoutCodegen` with this change.

packages/amplify-graphql-auth-transformer/src/vtl-generator/rds/resolvers/common.ts
sundersc: PR has been updated now with the correct changes.

packages/amplify-e2e-core/src/init/amplifyPush.ts
phani-srikar: Can you just add a task for this to migrate the usage of `amplifyPushWithoutCodegen`? Don't really want two methods that do the same thing.

alharris-at (approved):

packages/amplify-e2e-tests/src/__tests__/rds-mysql-auth-apikey-lambda.test.ts
```diff
@@ -27,7 +27,7 @@ describe('RDS Relational Directives', () => {
   // Generate settings for RDS instance
   const username = db_user;
   const password = db_password;
-  let region = 'ap-northeast-2';
+  let region = 'us-east-1';
```
phani-srikar, line 30: We're talking about allow listing our E2E account to test this in ICN first. Can keep this the same.

phani-srikar:
The build job is still failing

sundersc:
All the tests are passing now.

alharris-at (approved):

phani-srikar (approved):

PR #1958, opened 2023-10-16
chore: flip back from pinned and resolved versions of CLI everywhere

alharris-at:
#### Description of changes
Unpinning CLI packages, and pushing to run E2E flows as well, will continue iterating to address remaining failing tests after that pass is complete.

Changes and verifications:
- [x] Remove CI and Target steps targeting pinned Version
- [x] Use `yarn update-cli-packages` to fix all nested `package.json` files (and ensure we're back on caret deps).
- [x] Verify that `yarn setup-dev` creates a functional `amplify-dev` target for testing and debugging.
- [x] Verify that `yarn cloud-e2e` runs an e2e suite and is functionality.
- [x] Fix any newly failing e2e tests.
- [x] Address specifically the `api_rest_lambda_migration_v10` test that has been known to fail.

##### CDK / CloudFormation Parameters Changed
N/A

#### Issue #, if available
N/A

#### Description of how you validated changes
Build, Unit, E2E Tests

#### Checklist
- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are [changed or added](https://github.com/aws-amplify/amplify-cli/blob/master/CONTRIBUTING.md#tests)

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1968, opened 2023-10-19
merge main

alharris-at:
Merge `main` into feature branch, pulling in the following change:

- chore: refactor auth to accept params only via context
- fix: update tests for api change
- chore: flip back from pinned and resolved versions of CLI everywhere (#1958)


AaronZyLee (approved):

alharris-at:
Failed on stale `yarn.lock`. Refreshing that file before committing.

PR #1972, opened 2023-10-19
chore: add datasource provision strategy to replace transformer config value

AaronZyLee:
#### Description of changes

Added `DatasourceProvisionStrategy` top level attribute to replace the usage of `useAmplifyManagedTableResources` transformer config value. The strategies include

- `DEFAULT`: use CFN type `AWS::DynamoDB::Table`
- `AMPLIFY_TABLE`: use custom type `Custom::AmplifyDynamoDBTable`

The way to pass in the strategy is via the `AmplifyGraphqlDefinition` static methods `fromString`, `fromFiles`. Also a new method `combine` is added to merge the definitions.

One thing to note the RDS related changes are not included. They will be implemented in the RDS feature branch by @palpatim 

Example code snippet:
```TypeScript
const schema = `
  input AMPLIFY { globalAuthRule: AuthRule = { allow: public } } # FOR TESTING ONLY!
  
   type Todo @model {
    id: ID!
    name: String @index(name: "byName")
   }
  type Author @model {
    id: ID!
    name: String
  }
`
export class TempTestStackMain extends Stack {
  constructor(scope: Construct, id: string, props?: StackProps) {
    super(scope, id, props);
    const amplifyTableStrategy: ModelDataSourceDefinition = {
      name: 'customDDB',
      strategy: {
        dbType: 'DYNAMODB',
        provisionStrategy: 'AMPLIFY_TABLE',
      },
    };
    const definition1 = AmplifyGraphqlDefinition.fromString(schema, amplifyTableStrategy);
    const definition2 = AmplifyGraphqlDefinition.fromFiles('schema.graphql');
    const definition = AmplifyGraphqlDefinition.combine(definition1, definition2);
    const api = new AmplifyGraphqlApi(this, 'DemoAPI', {
      definition,
      authorizationModes: {
        apiKeyConfig: {
          expires: Duration.days(30),
        },
      },   
    });
}
```
Other minor changes:

- Reverted the custom resources adding to the root stack change to avoid the dependency issue (when destroying the stack, the IAM policies are deleted before the dynamoDB tables which fails the permission check)

Unit tests
Manual test via local CDK app (the code example above is deployed successfully)
#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes


packages/amplify-graphql-api-construct/API.md
```diff
@@ -57,6 +57,14 @@ export interface AddFunctionProps {
     readonly runtime?: FunctionRuntime;
 }
 
+// @public
+export interface AmplifyDynamoDBTableStrategy {
+    // (undocumented)
+    readonly dbType: 'DDB';
+    // (undocumented)
+    readonly provisionStrategy: 'AMPLIFY_TABLE';
+}
```
alharris-at, line 66: How would I configure for `replacement` instead of `iteration`?

alharris-at (dismissed):

packages/amplify-graphql-api-construct/API.md
AaronZyLee: The param for enabling replacement is not in scope of this PR. Whether to include that in the strategy or not is still TBD. Mark this as TODO.

alharris-at (approved):

packages/amplify-graphql-api-construct/src/amplify-graphql-definition.ts
```diff
@@ -1,32 +1,90 @@
...
       functionSlots: [],
+      dataSourceProvisionConfig: {
+        project: modelDataSourceDefinition,
+      },
+    };
+  }
+
+  /**
+   * Combines multiple IAmplifyGraphqlDefinitions into a single definition.
+   * @param definitions the definitions to combine
```
palpatim, line 60: We should document the error conditions. Off the top of my head, the only one I can think of is conflicting model definitions (if we find the same model in multiple definitions). For SQL support, we'll also disallow cross-type many-to-many relationships.

packages/amplify-graphql-api-construct/src/amplify-graphql-definition.ts
```diff
@@ -1,32 +1,90 @@
...
+   */
+  static combine(...definitions: IAmplifyGraphqlDefinition[]): IAmplifyGraphqlDefinition {
+    if (definitions.length === 0) {
+      throw new Error('The definitions of amplify GraphQL cannot be empty.');
+    }
+    if (definitions.length === 1) {
+      return definitions[0];
+    }
+    const projectStrategy = definitions[0].dataSourceProvisionConfig.project;
```
palpatim, line 69: It looks like you're using the `project` config of the first definition to apply to all subsequent definitions, but that's not how it should work. Each definition should be able to supply its own configuration, which will be independently resolved. If we really want a top-level config (which I would argue against), then it should be a separate argument to the `combine` method.

packages/amplify-graphql-api-construct/src/types.ts
```diff
@@ -541,6 +531,63 @@ export interface IBackendOutputStorageStrategy {
...
 }
 
+/**
+ * Project level configuration for datasource provision strategy.
+ */
+export interface DataSourceProvisionConfig {
+  /**
+   * Project-wide config for datasource provision. Applies to all non-overridden models.
+   */
+  readonly project?: ModelDataSourceDefinition;
```
palpatim, line 541: See note above re: "default" vs "project"

packages/amplify-graphql-transformer-core/src/utils/provision-strategy-utils.ts
```diff
@@ -0,0 +1,32 @@
...
+  TransformerBeforeStepContextProvider,
+} from '@aws-amplify/graphql-transformer-interfaces';
+
+/**
+ * Get the project-level datasource provision strategy. When typeName is provided, the model level strategy will be fetched
+ * @param ctx transfomer before step context
+ * @param typeName model name defined in GraphQL schema defintion
+ * @returns Datasource provision strategy for the provided model. Undefined if not found
+ */
+export function getDatasourceProvisionStrategy(
```
palpatim, line 13: This shouldn't return `undefined`; instead we should ensure that we assign the default dynamoDB model strategy to each model, and then we can be sure that each model will use the default DynamoDB provisioning strategy unless otherwise overridden.

alharris-at (commented):
Revoking approval, per Tim's comment on the breaking change to `fromFiles`.

alharris-at (changes requested):
Attempting to revoke approval pending interface update.

packages/amplify-graphql-api-construct/src/amplify-graphql-definition.ts
AaronZyLee: The first definition is not applied to all subsequent definitions unless the latter strategy is different from the first one(either `name` or `provisionStrategy`, or other fields related to RDS usage). When the difference is detected, the corresponding config will be keyed by the model type name.

AaronZyLee: The initial thought on this is that if the user only uses one strategy (like the default DDB), we don't need to map every model type name to the same value but only use a default value for all.

packages/amplify-graphql-api-construct/src/amplify-graphql-definition.ts
AaronZyLee: I think this is better to address after the RDS changes are merged. Mark this as TODO

alharris-at (approved):

packages/amplify-graphql-api-construct/src/amplify-graphql-definition.ts
```diff
@@ -1,32 +1,90 @@
 import * as os from 'os';
 import { SchemaFile } from 'aws-cdk-lib/aws-appsync';
-import { IAmplifyGraphqlDefinition } from './types';
+import { IAmplifyGraphqlDefinition, ModelDataSourceDefinition } from './types';
+import { getModelTypeNames } from './internal';
 
+export const DEFAULT_MODEL_DATA_SOURCE_DEFINITION: ModelDataSourceDefinition = {
+  name: 'defaultDDB',
```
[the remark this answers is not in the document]
AaronZyLee: Ah I forgot this point. Since it is only a renaming, I will make this change in the follow up PR if no other major changes required for the current one

packages/amplify-graphql-api-construct/src/amplify-graphql-definition.ts
palpatim: But if I'm reading this right, you're losing all of the model-specific overrides in def 1 since you `slice()` the definition array but never store the explicitly mapped models?

E.g. in this code, you'd lose the override on the T2 model:

```typescript
const def1 = AmplifyGraphQlApiDefinition.fromString('type T1 @model {id: ID!} type T2 @model {id: ID!}', {
  models: {
    T2: {
      dbType: 'DYNAMODB',
      provisioningStrategy: 'AMPLIFY_MANAGED'
    }
  }
});

const def2 = AmplifyGraphQlApiDefinition.fromString('type T3 @model {id: ID!} type T4 @model {id: ID!}', {
  default: {
      dbType: 'DYNAMODB',
      provisioningStrategy: 'AMPLIFY_MANAGED'
  },
  models: {
    T4: {
      dbType: 'DYNAMODB',
      provisioningStrategy: 'DEFAULT'
    }
  }
});
```

Further, our internal implementation is going to be simpler if we can assume at every level other than the UI that there is an explicit mapping of model to data source. The factory methods, including `combine`, should just go through each definition and create an explicit map of model to ModelDataSourceDefinition, and pass that through the rest of the system. That's also the place where we can quickly detect model conflicts (e.g., if def2 above had attempted to redefine the `T1` model), allowing us to fail fast.

palpatim (commented):
A few minor comments, but lgtm overall

alharris-at (approved):
Agreed w/ Tim's comments, but dropping a check as well.

alharris-at (approved):

palpatim (approved):

PR #1979, opened 2023-10-20
feat: add postgres engine and update types as needed

phani-srikar:
#### Description of changes

- Add support for `Postgres` engine for the RDS models.
- Refactor the transformers where we directly use `MYSQL_DB_TYPE` to make some decisions to now use common methods for RDS data sources.
- Add a new environment variable for engine to be set in the RDS lambda.
- Add array type fields in the metadata inside the resolver stash so that we take advantage of the array specific operators supported in Postgres.

Added unit tests
E2Es will be added in subsequent PR

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-graphql-relational-transformer/src/graphql-has-many-transformer.ts
```diff
@@ -53,6 +53,7 @@ import {
   validateRelatedModelDirective,
 } from './utils';
 import { getGenerator } from './resolver/generator-factory';
+import { is } from 'immer/dist/internal';
```
palpatim, line 56: Is this an accidental import? (I had one of those earlier today, so if so, I sympathize. :) )

packages/amplify-graphql-relational-transformer/src/resolver/generator-factory.ts
```diff
@@ -1,10 +1,11 @@
-import { DBType, MYSQL_DB_TYPE } from '@aws-amplify/graphql-transformer-core';
+import { DBType, MYSQL_DB_TYPE, POSTGRES_DB_TYPE } from '@aws-amplify/graphql-transformer-core';
 import { RDSRelationalResolverGenerator } from './rds-generator';
 import { DDBRelationalResolverGenerator } from './ddb-generator';
 import { RelationalResolverGenerator } from './generator';
 
 export const getGenerator = (dbType: DBType): RelationalResolverGenerator => {
   switch (dbType) {
+    case POSTGRES_DB_TYPE:
```
palpatim, line 8: Just checking that our linting rules allow fallthrough cases? It's a common source of errors, so a lot of projects are configured to disallow it.

packages/amplify-graphql-transformer-core/API.md
```diff
@@ -19,6 +19,7 @@ import { CfnResource } from 'aws-cdk-lib';
 import { Construct } from 'constructs';
 import { DataSourceInstance } from '@aws-amplify/graphql-transformer-interfaces';
 import { DataSourceProvider } from '@aws-amplify/graphql-transformer-interfaces';
+import { DBType as DBType_2 } from '@aws-amplify/graphql-transformer-interfaces';
```
palpatim, line 22: Why would we alias this?

packages/amplify-graphql-transformer-core/src/types/import-appsync-api-types.ts
```diff
@@ -29,5 +29,6 @@ export type RDSConnectionSecrets = TransformerSecrets & {
 
 export const MYSQL_DB_TYPE = 'MySQL';
 export const DDB_DB_TYPE = 'DDB';
+export const POSTGRES_DB_TYPE = 'Postgres';
 
-export type ModelDatasourceType = 'MySQL' | 'DDB';
+export type ModelDatasourceType = 'MySQL' | 'DDB' | 'Postgres';
```
palpatim, line 34: nit: Please alphabetize. This has the added effect of putting the most commonly used/default option (DDB) first
```suggestion
export type ModelDatasourceType = 'DDB' | 'MySQL' | 'Postgres';
```

packages/amplify-graphql-model-transformer/rds-lambda/handler.ts
```diff
@@ -83,3 +83,10 @@ const getDBConfig = async (): DBConfig => {
 
   return config;
 };
+
+const getDBEngine = (): string => {
+  if (!process.env.engine) {
+    throw Error('Missing database engine configuration');
+  }
+  return process.env.engine;
```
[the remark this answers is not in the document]
phani-srikar: I thought we concluded on not doing that for now and stick to the current design of reading engine from the schema input. In case of CDK construct, the `bindingType` in `modelDataSourceBinding` input will be used to deduce the engine.

packages/amplify-graphql-model-transformer/src/resolvers/rds/query.ts
```diff
@@ -1,6 +1,11 @@
 import { TransformerContextProvider } from '@aws-amplify/graphql-transformer-interfaces';
 import { compoundExpression, list, methodCall, obj, printBlock, qref, ref, set, str } from 'graphql-mapping-template';
-import { constructAuthFilterStatement, constructNonScalarFieldsStatement, constructFieldMappingInput } from './resolver';
+import {
+  constructAuthFilterStatement,
+  constructNonScalarFieldsStatement,
```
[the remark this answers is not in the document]
phani-srikar: We are still storing the array fields as part of the non-scalar fields metadata since adapters like MySQL do not need the information about which is array or object.

packages/amplify-graphql-relational-transformer/src/graphql-has-many-transformer.ts
phani-srikar: Yes, this sneaked in my commit. Removed.

packages/amplify-graphql-transformer-core/API.md
phani-srikar: Wrong import. Fixed.

packages/amplify-graphql-relational-transformer/src/resolver/generator-factory.ts
phani-srikar: They do, but I also feel comfortable just being explicit. Updated.

packages/amplify-graphql-model-transformer/rds-lambda/handler.ts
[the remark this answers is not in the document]
palpatim: Sorry for the confusing comment: I'm referring to how we store the engine in the Lambda's environment variable. You are storing it as a string literal, rather than an SSM parameter path, which is what I recall us discussing.

sundersc (approved):

packages/amplify-graphql-model-transformer/rds-lambda/handler.ts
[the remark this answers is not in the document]
phani-srikar: Engine doesn't disclose any sensitive DB information and was previously reviewed by AppSec to be allowed to exist in the schema. Also it can be deduced based on the modelToDatasource mapping wherever needed and not needed to be persisted and fetched from SSM.
Discussed with Chris and will sync with you on this at a later time and refactor in separate PR if needed.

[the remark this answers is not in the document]
palpatim: Sounds good. My concern isn't about security (Lambda Env vars are encrypted, so they're already secure) but rather about access patterns: we now have 5 parameters that we retrieve from SSM, but this special snowflake that we grab directly from the env var. I think it's fine for now; we can talk about overall storage later--we might even consider a single stringified JSON param rather than individual params, but that's a conversation for another day.

palpatim (approved):

PR #1983, opened 2023-10-23
feat: Invoke graphql transformer to generate SQL Lambda from AmplifyGraphqlApi construct

palpatim:
#### Description of changes

Adds plumbing to invoke the V2 graphql transformer to generate the SQL Lambda, required VPC Endpoints, and default auto-update infrastructure. Ensured the Lambda is exposed on the `AmplifyGraphqlApi.resources` path.

- Added unit tests to assert that the Lambda & VPC endpoints are present
- Cannot seem to assert Lambda environment variables: CDK doesn't seem to provide a property for them
- CDK `IFunction.isBoundToVpc` does not correctly report that the Lambda is bound to a VPC even though the configuration is present. The unit test casts through `CfnFunction`, which is how customers would have to inspect as well if they care to. Since customers must provide the VPC configuration as an input, this shouldn't be a problem, but it's something to keep in mind if we extend support to greenfield scenarios.
- Auto-update unit tests will come with a separate change to allow for more granular management
- Future changes will update the test logic to ensure we have valid naming for the SQL Lambdas associated with each schema snippet
- E2E tests will come in a separate PR

* No changes to CDK/CFN params
* Added new static helper method `AmplifyGraphqlApiDefinition.fromBindingAndFiles`
* Added new `modelDataSourceBinding` parameter to `AmplifyGraphqlApiDefinition.fromString`
* Added public types to allow customers to invoke above APIs

* Manual testing: Deployed new CDK app using local install of this code
* Unit testing: Asserted transformer-generated resources are present in synthesized stack

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-graphql-api-construct/API.md
```diff
@@ -119,8 +119,9 @@ export interface AmplifyGraphqlApiResources {
 
 // @public
 export class AmplifyGraphqlDefinition {
+    static fromBindingAndFiles(modelDataSourceBinding: ModelDataSourceBinding, ...filePaths: string[]): IAmplifyGraphqlDefinition;
     static fromFiles(...filePaths: string[]): IAmplifyGraphqlDefinition;
```
alharris-at, line 123: I think if we don't just have a pure vararg I'd avoid the spread on filePaths, and explicitly pass in an array.

packages/amplify-graphql-api-construct/src/amplify-graphql-api.ts
```diff
@@ -188,7 +190,13 @@ export class AmplifyGraphqlApi extends Construct {
         ...defaultTranslationBehavior,
         ...(translationBehavior ?? {}),
       },
-    });
+    };
+
+    if (isSqlModelDataSourceBinding(definition.modelDataSourceBinding)) {
+      this.extendTransformConfig(executeTransformConfig, definition.modelDataSourceBinding);
+    }
```
alharris-at, line 197: This appears to be a side-effecting function, but also returns the updated object. Let's adhere to one or the other. I think longer-term going to a 'transformConfigBuilder' may make this more readable, but I'm fine keep that OOS for this change.

palpatim: D'oh! Thanks, I intended to move away from the side effect in favor of immutable objects, and didn't close the task. Will fix.

packages/amplify-graphql-api-construct/API.md
palpatim: Went back & forth on this. I generally disfavor varargs in any case since it makes extensions hard (which is why we can't just add an argument to the `fromFiles` static factory method. I'll rework the signature to `fromFilesAndBinding(filePaths: string[], modelDataSourceBinding: ModelDataSourceBinding)` to keep it parallel with `fromString` above.

packages/amplify-graphql-transformer-core/src/transformer-context/index.ts
```diff
@@ -29,6 +29,7 @@ import { ResolverManager } from './resolver';
 import { TransformerResourceHelper } from './resource-helper';
 import { StackManager } from './stack-manager';
 import { assetManager } from './asset-manager';
+import { DataSource } from 'aws-cdk-lib/aws-stepfunctions-tasks';
```
palpatim, line 32: Huh. Not sure where this came from...I'm assuming I was intending to import the `DatasourceType` utility type and VSCode tried to be helpful. Will remove or replace with correct type.

palpatim (commented):
marker review

palpatim:
> Last macro comment, I see you've waffled back from adding the SqlBinding stuff from the GraphqlDefinition. I'm not immediately sure that I have a strong opinion either way, but if this increases concept count, and having that builder be implicit in the definition builder, we may want to do that. Food for thought.

I don't follow your comment. What I intended to do with this iteration was to make the "data source" an explicit part of the API definition:

```
GraphQLAPI
  |
  +-- Data source
      |
      +-- Model [, model2...modelN]
```

I think this sets up a reasonably easy-to-understand hierarchy that maps pretty closely to the way the system works. The types are now basically support types, rather than standalone construct or factory types, which I think is a cleaner DX overall since you don't have to import a new class to get to the factory: it's just an option on the existing factory method.

Is that addressing anything close to your comment?

packages/amplify-graphql-api-construct/API.md
alharris-at: Yeah, I have pretty immediate regrets on making that a varargs in the first place, since we can't easily change it now w/o impact to callers.

packages/amplify-graphql-api-construct/src/amplify-graphql-api.ts
palpatim: Resolved in bd11853f3

packages/amplify-graphql-api-construct/src/sql-model-datasource-binding.ts
```diff
@@ -0,0 +1,40 @@
...
+ * PostgreSQL
+ * @param obj the object to inspect
+ * @returns true if the object is shaped like a SqlModelDataSourceBinding
+ */
+export const isSqlModelDataSourceBinding = (obj: any): obj is SqlModelDataSourceBinding => {
+  return (typeof obj === 'object' || typeof obj === 'function') && typeof obj.bindingType === 'string' && obj.bindingType === 'MySQL';
+};
+
+/**
+ * Class exposing utilities to produce ISqlModelDataSourceBinding objects given various inputs.
```
alharris-at, line 16:
```suggestion
 * Class exposing utilities to produce SqlModelDataSourceBinding objects given various inputs.
```

packages/amplify-graphql-api-construct/src/sql-model-datasource-binding.ts
```diff
@@ -0,0 +1,40 @@
...
+ */
+export class SqlModelDataSourceBindingFactory {
+  /**
+   * Creates an ISqlModelDataSourceBinding where the binding's `customSqlStatements` are populated from `sqlFiles`. The key of the
```
alharris-at, line 20:
```suggestion
   * Creates a SqlModelDataSourceBinding where the binding's `customSqlStatements` are populated from `sqlFiles`. The key of the
```

packages/amplify-graphql-api-construct/src/sql-model-datasource-binding.ts
```diff
@@ -0,0 +1,40 @@
...
+   * `customSqlStatements` record is the file's base name (that is, the name of the file minus the directory and extension).
+   * @param sqlFiles the list of files to load SQL statements from.
+   * @param options the remaining ISqlModelDataSourceBinding options.
```
alharris-at, line 23:
```suggestion
   * @param options the remaining SqlModelDataSourceBinding options.
```

packages/amplify-graphql-transformer-core/src/transformer-context/index.ts
palpatim: Resolved in a82f5a05e

alharris-at:
> > Last macro comment, I see you've waffled back from adding the SqlBinding stuff from the GraphqlDefinition. I'm not immediately sure that I have a strong opinion either way, but if this increases concept count, and having that builder be implicit in the definition builder, we may want to do that. Food for thought.
> 
> I don't follow your comment. What I intended to do with this iteration was to make the "data source" an explicit part of the API definition:
> 
> ```
> GraphQLAPI
>   |
>   +-- Data source
>       |
>       +-- Model [, model2...modelN]
> ```
> 
> I think this sets up a reasonably easy-to-understand hierarchy that maps pretty closely to the way the system works. The types are now basically support types, rather than standalone construct or factory types, which I think is a cleaner DX overall since you don't have to import a new class to get to the factory: it's just an option on the existing factory method.
> 
> Is that addressing anything close to your comment?

I think that was a misreading on my part, I'd seen references to an ISql*Binding interface, and thought this was another top-level builder, but checking your example this makes sense to me.. Dropped a few change recommendations to update those references to remove the `I`.

palpatim:
> Dropped a few change recommendations to update those references to remove the `I`.

I don't see any change recommendations, but I also don't see any remaining exported types with a leading `I` in `types.ts`, so it's possible you were seeing an older change?

phani-srikar (approved):

alharris-at (approved):

PR #2002, opened 2023-10-30
chore: expand scope of adminRole test

alharris-at:
#### Description of changes
Improve functional coverage of adminRole e2e test for CDK construct. Now we deploy a lambda whose executionRole is registered as an adminRole, and flex that functionality.

Also added new lambdaInvoke method to e2e tests, and updated `init` command to support flags a bit more consistently.

N/A

#### Issue #, if available
N/A

E2E Test

#### Checklist
- [x] PR description included


alharris-at:
E2Es are still running (and a bunch failed for ERR:SocketTimeout), but the admin roles one has passed.

packages/amplify-graphql-api-construct-tests/package.json
```diff
@@ -21,9 +21,11 @@
     "build-tests": "tsc --build tsconfig.tests.json"
   },
   "dependencies": {
+    "@aws-sdk/client-lambda": "^3.438.0",
```
phani-srikar, line 24: This is pulling in several additional dependencies with it which could eventually cause the CLI size to increase. I see that we already use `3.338.0` version for the SDK clients including the lambda client. Can we use that version instead?

alharris-at: This shouldn't affect library size, this is is just test package dependencies, I can flip to use 338 though.

AaronZyLee (commented):
Looks like the build job fails in CI due to some type issue. Otherwise LGTM

AaronZyLee (approved):

PR #2004, opened 2023-10-31
feat: add graphqlSchemaFromRDSSchema for unauth schema import

dpilch:
#### Description of changes

Allow customers to generate a GraphQL schema from an RDS database without authenticating the Amplify CLI.

Add a new function `graphqlSchemaFromRDSSchema` to `@aws-amplify/graphql-schema-generator`. 

```
graphqlSchemaFromRDSSchema: string (
  rdsSchema: string,
  engineType: ImportedRDSType,
)
```

* rdsSchema - The RDS schema representation. Customer will get this by running a query on the database.
* engineType - The database engine the customer is using (mysql, postgress)
* return - the generated GraphQL schema to be written to the file system

MySQL query to get the rdsSchema:

```
SELECT INFORMATION_SCHEMA.COLUMNS.*,
  INFORMATION_SCHEMA.STATISTICS.INDEX_NAME,
  INFORMATION_SCHEMA.STATISTICS.NON_UNIQUE,
  INFORMATION_SCHEMA.STATISTICS.SEQ_IN_INDEX,
  INFORMATION_SCHEMA.STATISTICS.NULLABLE
FROM INFORMATION_SCHEMA.COLUMNS
LEFT JOIN INFORMATION_SCHEMA.STATISTICS
  ON INFORMATION_SCHEMA.COLUMNS.TABLE_NAME=INFORMATION_SCHEMA.STATISTICS.TABLE_NAME
  AND INFORMATION_SCHEMA.COLUMNS.COLUMN_NAME=INFORMATION_SCHEMA.STATISTICS.COLUMN_NAME
WHERE INFORMATION_SCHEMA.COLUMNS.TABLE_SCHEMA = '<database-name>';
```   

CLI Command

```
npx @aws-amplify/cli api generate-schema --rds-schema <rds-schema.csv> --engine-type mysql --out schema.graphql
```

N/A

* Unit testing
* Manual testing

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


alharris-at (approved):

packages/amplify-category-api/src/commands/api/generate-schema.ts
```diff
@@ -20,47 +21,72 @@ const subcommand = 'generate-schema';
...
+      if (!sqlSchema) {
+        printer.error('A SQL schema must be provided with --sql-schema');
+      }
+      if (!engineType) {
+        printer.error('An engine type must be provided with --engine-type');
+      }
+      if (!out) {
+        printer.error('An output path must be provided with --out');
+      }
+      return;
```
palpatim, line 41: Do we have a pattern for abnormal exits (e.g., exiting with a non-zero status) to indicate errors? Can we throw an error rather than just returning?

packages/amplify-category-api/src/provider-utils/awscloudformation/utils/graphql-schema-utils.ts
```diff
@@ -118,3 +78,46 @@ const parseSchema = (schemaContent: string | undefined, pathToSchemaFile: string
...
+    case ImportedRDSType.MYSQL:
+      adapter = new MySQLDataSourceAdapter(databaseConfig as MySQLDataSourceConfig);
+      schema = new Schema(new Engine('MySQL'));
+      break;
+    case ImportedRDSType.POSTGRESQL:
+      adapter = new PostgresDataSourceAdapter(databaseConfig as PostgresDataSourceConfig);
+      schema = new Schema(new Engine('Postgres'));
+      break;
+    default:
+      printer.error('Only MySQL and Postgres Data Sources are supported.');
```
palpatim, line 99: Need to throw, unless I'm misunderstanding that `printer.error` throws in addition to printing the error message.

packages/amplify-graphql-schema-generator/src/datasource-adapter/datasource-adapter.ts
```diff
@@ -11,14 +12,15 @@ export abstract class DataSourceAdapter {
...
 
-  public abstract mapDataType(datatype: string, nullable: boolean, tableName: string, fieldName: string, columnType: string): FieldType;
-
   public abstract initialize(): Promise<void>;
 
   public abstract cleanup(): void;
 
   public abstract test(): Promise<boolean>;
 
+  // todo correct function signature
```
palpatim, line 21: What's incorrect about it?

packages/amplify-graphql-schema-generator/src/generate.ts
```diff
@@ -0,0 +1,53 @@
...
+      adapter = new MySQLStringDataSourceAdapter(stringSchema);
+      schema = new Schema(new Engine('MySQL'));
+      break;
+    case ImportedRDSType.POSTGRESQL:
+      // todo change to postgres
+      adapter = new MySQLStringDataSourceAdapter(stringSchema);
+      schema = new Schema(new Engine('Postgres'));
+      break;
+    default:
+      throw new Error('Only MySQL and Postgres Data Sources are supported');
```
palpatim, line 23: As above, need to throw from here to prevent execution from proceeding past this error

packages/amplify-graphql-schema-generator/src/generate.ts
```diff
@@ -0,0 +1,53 @@
...
+    os.EOL +
+    os.EOL +
+    generateGraphQLSchema(schema, existingSchema)
+  );
+};
+
+export const graphqlSchemaFromRDSSchema = (sqlSchema: string, engineType: ImportedRDSType): string => {
+  const schema = buildSchemaFromString(sqlSchema, engineType);
+
+  const transformerVersion = 2;
```
palpatim, line 48: I'm surprised we don't have a centralized place to store transformer versions. How will we remember to catch this when we update the version at some point in future?

packages/amplify-category-api/src/commands/api/generate-schema.ts
dpilch: This is what is used for other failures. See the migrate command: https://github.com/aws-amplify/amplify-category-api/blob/main/packages/amplify-category-api/src/commands/api/migrate.ts

This probably should be changed though because it give a 0 exit code on the cli. Throwing an error will still call `printer.error`, so I'll change to that.

packages/amplify-graphql-schema-generator/src/datasource-adapter/datasource-adapter.ts
dpilch: Nothing is wrong. I forgot to remove the comment after updating.

packages/amplify-graphql-schema-generator/src/generate.ts
dpilch: An alternative here would be to change the function signature to `graphqlSchemaFromRDSSchema = (sqlSchema: string, engineType: ImportedRDSType, includeAuthRule: boolean = false)`

The transformer version is only used here: https://github.com/aws-amplify/amplify-category-api/pull/2004/files#diff-fbea81850cb6d8d9f8b12d224eaebab2b6377705a8799dcc1e7aa9de03537183R25

dpilch: I think I'll actually change to this. I think it makes more sense

palpatim: Looks like that would also require you to remove the transformerversion check from that block though, which may have unintended consequences later if we change the schema structure & associated version. I don't object to passing the version, per se, I am concerned by hardcoding it at arbitrary places through the code, rather than centralizing known versions into a single place that we can refer to throughout the codebase. If we don't already have such a centralization, lets defer the work to create it rather than bundling it into this change.

dpilch: Ok, I'll revert to when transformerVersion is hard coded so that it leaves the function as minimal as possible.

packages/amplify-graphql-schema-generator/src/input.ts
```diff
@@ -0,0 +1,88 @@
...
+): AmplifyInputEntry[] => {
+  const inputs: AmplifyInputEntry[] = [
+    {
+      name: 'engine',
+      type: 'String',
+      default: dataSourceType,
+    },
+  ];
+
+  if (includeAuthRule && transformerVersion === 2) {
```
phani-srikar, line 25: Since this is the only place we're using this `transformerVersion` and we only support this flow for V2, we can get rid of this input if we throw an error early on.

packages/amplify-category-api/src/commands/api/generate-schema.ts
```diff
@@ -20,47 +21,70 @@ const subcommand = 'generate-schema';
...
+      if (!out) {
+        throw new AmplifyError('UserInputError', { message: 'An output path must be provided with --out' });
+      }
+    }
+    if (!Object.values(ImportedRDSType).includes(engineType)) {
+      throw new AmplifyError('UserInputError', { message: `${engineType} is not a supported engine type.` });
+    }
+    const transformerVersion = await ApiCategoryFacade.getTransformerVersion(context);
+    const schema = await graphqlSchemaFromRDSSchema(fs.readFileSync(sqlSchema, 'utf8'), engineType, transformerVersion);
+    writeSchemaFile(out, schema);
```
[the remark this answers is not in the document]
dpilch: `writeSchemaFile` calls `fs.ensureFileSync` before writing.

palpatim (dismissed):

palpatim (approved):

phani-srikar (approved):

alharris-at (approved):

PR #2005, opened 2023-10-31
test: add Postgres E2E tests

phani-srikar:
#### Description of changes

- Expands the existing tests for MySQL to Postgres by abstracting out the test suites that are common across engines to a directory.
- Added the following tests for Postgres:
-- Model tests with primary and composite keys
-- Model re-naming using `refersTo`
-- Model fields re-naming using `refersTo`.
-- Regenerate Schema tests.

CI checks - E2E and PR flows passing

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-e2e-tests/src/__tests__/rds-pg-model-v2.test.ts
```diff
@@ -0,0 +1,16 @@
...
+import { testRDSModel } from '../rds-v2-tests-common/rds-model-v2';
+
+// to deal with bug in cognito-identity-js
+(global as any).fetch = require('node-fetch');
+
+describe('RDS Postgres Model Directive', () => {
+  const queries = [
+    'CREATE TABLE "Contact" (id text PRIMARY KEY, "firstName" text, "lastName" text)',
+    'CREATE TABLE "Person" ("personId" integer PRIMARY KEY, "firstName" text, "lastName" text)',
+    'CREATE TABLE "Employee" (id integer PRIMARY KEY, "firstName" text, "lastName" text)',
```
sundersc, line 11: Where do we run the queries and validate the result? For postgres, the fields must be in lower case or have appropriate 'refersTo' directive to work properly. I'm unable to find where we handle this logic.

phani-srikar: We handle the quoted identifiers in the lambda layer. We don't need a refersTo in this case since the names are idiomatic to GraphQL. For example, when we define a filter on say `firstName`, we will quote it before sending to DB like `"firstName" = "John"` for example. This applies to cases where we generate the query conditions ourselves. Otherwise the table and column names are automatically quoted by Knex. This lambda layer change hasn't been deployed to all regions yet, but I tested with my personal account that the added test works.

packages/amplify-e2e-tests/src/__tests__/rds-mysql-v2-generate-schema.test.ts
```diff
@@ -0,0 +1,16 @@
+import { ImportedRDSType } from '@aws-amplify/graphql-transformer-core';
+import { testRDSGenerateSchema } from '../rds-v2-tests-common/rds-v2-generate-schema';
+
+// to deal with bug in cognito-identity-js
+(global as any).fetch = require('node-fetch');
```
palpatim, line 5: Do we need this even though we're not doing any auth in the test?

palpatim (approved):
Question about the cognito-identity-js workaround, but otherwise LGTM

packages/amplify-e2e-tests/src/__tests__/rds-mysql-v2-generate-schema.test.ts
phani-srikar: Will remove this and see in subsequent PR.

sundersc (approved):

PR #2007, opened 2023-11-01
fix: pass in and use correct lambda arn for lambdaauthorizers

alharris-at:
#### Description of changes
Lambda authorizers were still computing an ARN which included `-NONE` at the end, to align w/ older amplify convention. This is no longer guaranteed, given the lambda can be created in a normal CDK app, instead pass in the IFunction ARN in a CDK context, and use that when available.

Updated the all-auth-roles e2e test for CDK to flex this functionality.

Yes, GraphqlApi auth config for lambda will use the arn if possible.

#### Issue #, if available
N/A

Updated E2E test and manually.

#### Checklist
- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-graphql-api-construct-tests/src/graphql-request.ts
+++ b/packages/amplify-graphql-api-construct-tests/src/graphql-request.ts
@@ -5,24 +5,13 @@ export type GraphqlResponse = {
   body: any;
 };
 
-export const graphql = async (apiEndpoint: string, apiKey: string, query: string): Promise<GraphqlResponse> => {
-  const options = {
-    method: 'POST',
-    headers: {
-      'x-api-key': apiKey,
-      'Content-Type': 'application/json',
-    },
-    body: JSON.stringify({ query }),
-  };
-
-  const request = new Request(apiEndpoint, options);
-
+const graphqlRequest = async (apiEndpoint: string, payload: any): Promise<GraphqlResponse> => {
   let statusCode = 200;
   let body;
   let response;
 
   try {
-    response = await fetch(request);
+    response = await fetch(new Request(apiEndpoint, payload));
     body = await response.json();
     if (body.errors) statusCode = 400;
   } catch (error) {
@@ -43,3 +32,23 @@ export const graphql = async (apiEndpoint: string, apiKey: string, query: string
     body: body,
   };
 };
+
+export const graphql = async (apiEndpoint: string, apiKey: string, query: string): Promise<GraphqlResponse> =>
+  graphqlRequest(apiEndpoint, {
+    method: 'POST',
+    headers: {
+      'x-api-key': apiKey,
+      'Content-Type': 'application/json',
+    },
+    body: JSON.stringify({ query }),
+  });
+
+export const graphqlRequestWithLambda = async (apiEndpoint: string, authToken: string, query: string): Promise<GraphqlResponse> =>
+  graphqlRequest(apiEndpoint, {
+    method: 'POST',
+    headers: {
+      Authorization: authToken,
+      'Content-Type': 'application/json',
+    },
+    body: JSON.stringify({ query }),
+  });
--- a/packages/amplify-graphql-transformer-core/src/utils/authType.ts
+++ b/packages/amplify-graphql-transformer-core/src/utils/authType.ts
@@ -76,6 +76,7 @@ export const adoptAuthMode = (
       return {
         authorizationType: authType,
         lambdaAuthorizerConfig: {
+          lambdaArn: entry.lambdaAuthorizerConfig!.lambdaArn,
           lambdaFunction: entry.lambdaAuthorizerConfig!.lambdaFunction,
           ttlSeconds: strToNumber(entry.lambdaAuthorizerConfig!.ttlSeconds),
         },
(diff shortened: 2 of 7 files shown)


alharris-at:
Note: all e2e tests are still running, but the one I touched here `all-auth-roles.test.ts` has succeeded.

packages/amplify-graphql-api-construct-tests/src/graphql-request.ts
```diff
@@ -5,24 +5,13 @@ export type GraphqlResponse = {
...
+const graphqlRequest = async (apiEndpoint: string, payload: any): Promise<GraphqlResponse> => {
   let statusCode = 200;
   let body;
   let response;
 
   try {
-    response = await fetch(request);
+    response = await fetch(new Request(apiEndpoint, payload));
     body = await response.json();
     if (body.errors) statusCode = 400;
```
palpatim, line 16: (Side note: I know this is pre-existing code, but it seems odd that we're overriding the status code from the service. I recognize that the service returning 200 in error conditions is weird behavior driven by the need to support partial responses, but it's strange that our tests suppress that behavior rather than dealing with it.)

packages/amplify-graphql-transformer-interfaces/src/graphql-api-provider.ts
```diff
@@ -57,6 +57,7 @@ export interface OpenIDConnectConfig {
 
 export interface LambdaConfig {
   lambdaFunction: string;
+  lambdaArn?: string;
```
palpatim, line 60: As a customer, I'd interpret this to mean that I could reuse an existing function (which is in line with your description of the change), but since the config still requires me to provide a `lambdaFunction`, I'm not sure how to deal with it. How can we clarify the intent of the optional `lambdaArn` field on this interface, and specifically how it interacts with the `lambdaFunction` field?

packages/amplify-graphql-api-construct-tests/src/graphql-request.ts
alharris-at: I agree, this took me a little while to figure out why the response code wasn't what I expected. I don't want to go through the 2-4 hour cycle for running e2e's again, so I'll track that as a fast-follow today.

packages/amplify-graphql-transformer-interfaces/src/graphql-api-provider.ts
alharris-at: Yeah, `lambdaFunction` as the name isn't super clear, I'll drop a comment that this is the expected function name (excluding optional env).

packages/amplify-graphql-transformer-interfaces/src/graphql-api-provider.ts
```diff
@@ -56,7 +56,19 @@ export interface OpenIDConnectConfig {
 }
 
 export interface LambdaConfig {
+  /**
+   * Function name, excluding optional `-{env}` suffix.
+   */
   lambdaFunction: string;
+
+  /**
+   * Actually lambda Arn, if provided this will circumvent the arn construction when building the API auth mode config.
```
palpatim, line 65:
```suggestion
   * The ARN of an existing Lambda function. If provided, this will circumvent the ARN construction when building the API auth mode config. The ARN must refer to the same function
   * named in `lambdaFunction`.
```

packages/amplify-graphql-transformer-core/src/utils/authType.ts
```diff
@@ -76,6 +76,7 @@ export const adoptAuthMode = (
       return {
         authorizationType: authType,
         lambdaAuthorizerConfig: {
+          lambdaArn: entry.lambdaAuthorizerConfig!.lambdaArn,
```
phani-srikar, line 79: Is this `!` intended since this is an optional property?

packages/amplify-graphql-transformer-interfaces/src/graphql-api-provider.ts
phani-srikar: Do we need the functionName as well when customers provide the Arn?

packages/amplify-graphql-transformer-core/src/utils/authType.ts
phani-srikar: If this is invoked after you construct it here https://github.com/aws-amplify/amplify-category-api/pull/2007/files#diff-d7d5f095c801bacf0479e8cd421efbcb56d9280b9e3a7eb1d254999c6e472f80R60, it should be fine.

phani-srikar (approved):

packages/amplify-graphql-transformer-core/src/utils/authType.ts
alharris-at: Yeah, there's probably a way to get the types to winnow correctly, but since the authorizationType has matched this is expected to exist.

packages/amplify-graphql-transformer-interfaces/src/graphql-api-provider.ts
alharris-at: It's possible we could make it optional in the future, but that'll require more tracing through the codebase here to ensure it's not used elsewhere (I don't trust the type system to complain in these packages if we make something optional that's actually required).

palpatim (approved):

PR #2008, opened 2023-11-01
fix: treating id field as implicit primary key which we miss today

phani-srikar:
#### Description of changes

The primary key transformer does not recognize the default `id: ID!` field on a model as implicit primary key and hence does not generate any resolvers for that field. One of the issues arising this is that the `preAuth` slot for Sync Queries is not generated which causes the Sync to be slow as it scans the whole table even if the filter is specified on the `id` field by the customers. Other side effects include missing `pre-Auth` slots that set the primary key information for CRUD operation resolvers for the models.
Please see the snapshot changes for the full change set.
This could be a breaking change since we're generating several new resolver slots and their overall evaluation might break existing apps.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/codebuild_specs/e2e_workflow.yml
+++ b/codebuild_specs/e2e_workflow.yml
@@ -135,24 +135,24 @@ batch:
           CLI_REGION: us-east-1
       depend-on:
         - publish_to_local_registry
-    - identifier: api_1_resolvers_sync_query_datastore_api_6
+    - identifier: rds_pg_refersTo_api_1_resolvers_sync_query_datastore
       buildspec: codebuild_specs/run_e2e_tests.yml
       env:
         compute-type: BUILD_GENERAL1_MEDIUM
         variables:
           TEST_SUITE: >-
-            src/__tests__/api_1.test.ts|src/__tests__/resolvers.test.ts|src/__tests__/graphql-v2/sync_query_datastore.test.ts|src/__tests__/api_6.test.ts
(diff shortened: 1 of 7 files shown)


codebuild_specs/e2e_workflow.yml
```diff
@@ -135,24 +135,24 @@ batch:
           CLI_REGION: us-east-1
       depend-on:
         - publish_to_local_registry
-    - identifier: api_1_resolvers_sync_query_datastore_api_6
+    - identifier: rds_pg_refersTo_api_1_resolvers_sync_query_datastore
```
phani-srikar, line 138: This will go away once my other commit shows up on Github. I pushed it but its not there yet for some reason.

packages/amplify-graphql-index-transformer/src/__tests__/amplify-graphql-primary-key-transformer.test.ts
```diff
@@ -1006,6 +1006,37 @@ describe('RDS primary key transformer tests', () => {
...
   });
+
+  it('id is treated as implicit primaryKey and necessary resolver resources are generated', () => {
+    const validSchema = `
+      type Song @model {
+        id: ID!
+        name: String!
+        genre: String!
+      }
+    `;
```
alharris-at, line 1017: Can we add a suite of tests that validates the entire app generated for this schema is the same as what is generated if there's an explicit `@primaryKey` on `id`? That'd give me more confidence that we're doing the right thing across all transformers in concert.

PR #2012, opened 2023-11-02
test: E2E tests for SQL support in CDK construct

palpatim:
#### Description of changes

Test 1 for SQL support in the CDK construct. This tests the basic case of specifying connection params for a pre-existing SQL database, and letting the transformer provision Lambda, VPC endpoint, and IAM resources via CDK.

Other changes:
* Add a new env variable to allow local invocations to skip the 0-3 minute delay when setting up a new project folder
* Add a `create-env.sh` script to write an `.env` file based on AWS environment variables in the current shell session

#### Checklist

- [x] PR description included
- [X] `yarn test` passes
- [X] Tests are [changed or added](https://github.com/aws-amplify/amplify-cli/blob/master/CONTRIBUTING.md#tests)


packages/amplify-e2e-core/src/index.ts
```diff
@@ -99,10 +99,10 @@ export async function installAmplifyCLI(version: string = 'latest') {
   console.log('PATH SET:', process.env.AMPLIFY_PATH);
 }
 
-export async function createNewProjectDir(
+export const createNewProjectDir = async (
```
palpatim, line 102: Linting rules wanted const exports rather than function exports, so I did that a few places while I was troubleshooting my test setups.

packages/amplify-e2e-core/src/utils/rds.ts
```diff
@@ -76,6 +92,7 @@ export const createRDSInstance = async (config: RDSConfig): Promise<{ endpoint:
       endpoint: dbInstance.Endpoint.Address as string,
       port: dbInstance.Endpoint.Port as number,
       dbName: dbInstance.DBName as string,
+      dbInstance,
```
palpatim, line 95: Added dbInstance to support discovery of VPC configs elsewhere. Technically we could just return the dbInstance since endpoint, port, dbName are all contained therein, but that was more work than I wanted to take on since it would impact existing tests.

packages/amplify-graphql-api-construct-tests/src/__tests__/backends/sql-models/app.ts
```diff
@@ -0,0 +1,72 @@
...
+import 'source-map-support/register';
+import { App, Stack, Duration } from 'aws-cdk-lib';
+// @ts-ignore
+import {
+  AmplifyGraphqlApi,
+  AmplifyGraphqlDefinition,
+  SqlModelDataSourceBindingDbConnectionConfig,
+} from '@aws-amplify/graphql-api-construct';
+
+interface DBDetails {
```
palpatim, line 11: This is a copy/paste of the same interface from the test code. I played with a couple of ways to factor it into a standalone file, but couldn't figure out a way I liked to copy it into the app.ts workspace to make it available for both test-time and compile-time use. Open to suggestions.

palpatim (commented):
marker review

packages/amplify-graphql-api-construct-tests/src/__tests__/sql-models.test.ts
```diff
@@ -0,0 +1,214 @@
...
+      region,
+    });
+  });
+
+  afterAll(async () => {
+    await cleanupDatabase({ identifier: identifier, region, dbDetails });
+  });
+
+  beforeEach(async () => {
+    projFolderName = 'sqlmodels';
```
alharris-at, line 67: No need to store this in a separate variable, can just inline the string and get rid of the `let`.

packages/amplify-graphql-api-construct-tests/src/__tests__/sql-models.test.ts
```diff
@@ -0,0 +1,214 @@
...
+      const result = await cdkDestroy(projRoot, '--all');
+      console.log(`cdkDestroy result: ${JSON.stringify(result)}`);
+    } catch (_) {
+      /* No-op */
+    }
+
+    deleteProjectDir(projRoot);
+  });
+
+  ['2.80.0', 'latest'].forEach((cdkVersion) => {
```
alharris-at, line 82: Your choice if you want to run both of these versions. We just run the basic smoke test against both to ensure we don't have grosse type errors occuring, etc, but it DOES slow you down quite a bit when you're running these.

packages/amplify-graphql-api-construct-tests/src/__tests__/sql-models.test.ts
```diff
@@ -0,0 +1,214 @@
...
+
+      expect(result).toMatchSnapshot({
+        body: {
+          data: {
+            createTodo: {
+              id: expect.any(String),
+            },
+          },
+        },
+      });
```
alharris-at, line 111: I know I was using snapshots in the `cdk-base` tests, but I think it's probably better to avoid snapshots, and just validate the data explicitly (e.g. response data and response codes) in the test itself.

packages/amplify-graphql-api-construct-tests/src/__tests__/sql-models.test.ts
palpatim: Spectacular. I'll take out the explicitly versioned one.

packages/amplify-graphql-api-construct-tests/src/__tests__/sql-models.test.ts
palpatim: Fair enough. I'll make sure the expect assertions cover the data appropriately (which they should already)

alharris-at (approved):

packages/amplify-graphql-api-construct-tests/src/__tests__/sql-models.test.ts
```diff
@@ -0,0 +1,194 @@
...
+      /* No-op */
+    }
+
+    deleteProjectDir(projRoot);
+  });
+
+  it('creates a GraphQL API from SQL-based models', async () => {
+    const templatePath = path.resolve(path.join(__dirname, 'backends', 'sql-models'));
+    const name = await initCDKProject(projRoot, templatePath, 'latest');
+    writeDbDetails(dbDetails, projRoot);
```
phani-srikar, line 84: Was hoping there would be a pass these to the App via context or app environment, but my brief web search did not point me to anything useful.

palpatim:
Test stacks aren't being deleted after the test. Investigating.

phani-srikar (approved):

palpatim:
Changes since last approval: 
* 735094b39 - Rebased against `feature/rds-support-preview2` & addressed merge conflicts
* 033626c94 - Fixed error destroying CDK project
* 4a1c64a92 - Added construct test reports to .prettierignore

palpatim (commented):
Marker review

phani-srikar (approved):

PR #2013, opened 2023-11-03
test(api): pg e2e auth tests

sundersc:
Add E2E auth tests for postgres database.
- IAM Auth
- Api Key
- Lambda Authorizer


packages/amplify-e2e-core/src/utils/nexpect.ts
```diff
@@ -683,9 +683,11 @@ function createExpectationError(expected: string | RegExp, actual: string) {
 
 export function nspawn(command: string | string[], params: string[] = [], options: SpawnOptions = {}) {
   if (Array.isArray(command)) {
+    console.log(`Executing command: ${command.join(' ')} with params: ${params.join(' ')}`);
```
palpatim, line 686: I'm a bit nervous about logging the command parameters, in case we spawn something with a secret value. Since we can't know ahead of time what the parameters are, we can either add an optional log config to the SpawnOptions, or just suppress parameter logging altogether.

packages/amplify-e2e-tests/src/__tests__/rds-pg-auth-apikey-lambda.test.ts
```diff
@@ -0,0 +1,447 @@
...
+  };
+
+  const constructUserHelper = (client): GQLQueryHelper => {
+    const createSelectionSet = /* GraphQL */ `
+      id
+      name
+    `;
+    const updateSelectionSet = createSelectionSet;
+    const deleteSelectionSet = createSelectionSet;
+    const getSelectionSet = /* GraphQL */ `
```
palpatim, line 359: nit: this is the whole query, not a selection set :)

packages/amplify-e2e-tests/src/__tests__/rds-pg-auth-apikey-lambda.test.ts
```diff
@@ -0,0 +1,447 @@
...
+          id
+          name
+          profile {
+            id
+            details
+          }
+        }
+      }
+    `;
+    const listSelectionSet = /* GraphQL */ `
```
palpatim, line 371: naming nit: not a selection set

phani-srikar (approved):
I've noted this previously as well. Since these tests are same across engines with minor setup changes, abstracting these into common tests folder would've improved the maintainability.

packages/amplify-e2e-core/src/utils/nexpect.ts
sundersc: This helps to understand the status of the E2E tests when we are waiting for them to complete. Currently we do not pass any secrets inline as part of the command but I agree that it could change in the future. Are you okay to disable it later if we decide to pass any secrets inline?

sundersc: > just suppress parameter logging altogether

The command is always just 'amplify'. Example of parameters are 'status -v' / 'push -y' / 'api gql-compile --force').

sundersc:
> I've noted this previously as well. Since these tests are same across engines with minor setup changes, abstracting these into common tests folder would've improved the maintainability.

Refactored. Moved the tests to common module.

packages/amplify-e2e-core/src/utils/nexpect.ts
phani-srikar: Since this change is at the root level where ANY amplify command is executed, I would bias towards not logging anything here.

palpatim: It's not always just `amplify`: we use it to spawn `npx` processes in the CDK e2e tests. The point being, we don't have control over the inputs to this method, so logging them is dangerous. Unless you want to write a masker to redact anything not beginning with a dash character, I'd say suppress param logging.

phani-srikar (commented):
Thanks for making those abstractions. Did you verify that these tests pass locally?

sundersc:
> Thanks for making those abstractions. Did you verify that these tests pass locally?

I ran all the new tests for Postgres. Since it is untouched for MySQL, I believe it should pass.

phani-srikar (approved):

palpatim (approved):

PR #2033, opened 2023-11-07
test: add unauth sql generate e2e tests

dpilch:
#### Description of changes

Add E2E tests for unauth sql generate schema

N/A

#### Issue #, if available

N/A

Run the new tests

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-e2e-core/src/categories/api.ts
```diff
@@ -1102,6 +1102,37 @@ export const importRDSDatabase = (cwd: string, opts: ImportApiOptions & { apiExi
...
+    }
+
+    generateCommand.run((err: Error) => {
+      if (!err) {
+        resolve();
+      } else {
+        reject(err);
+      }
+    });
+  });
```
alharris-at, line 1133: You should be able to lose the `new Promise...` and just use `return generateCommand.runAsync()` at the end, right?

alharris-at (approved):
Not blocking, but you can cleanup the command a bit.

packages/amplify-e2e-core/src/categories/api.ts
dpilch: I was going to say we don't have this functionality, because it is not available in the codegen repo. It is available in this repo though 🤔

palpatim (dismissed):

alharris-at (approved):

PR #2038, opened 2023-11-07
add new @aws-amplify/data-construct alias for the @aws-amplify/graphql-api-construct

alharris-at:
#### Description of changes
Create new cdk construct which will be published using the `data` moniker that is a clone of the graphql-api-construct.

**Note: I'm leaving the new construct package private until we're closed on details.**

Nothing internally, but externally we're publishing a new set of constructs.

#### Issue #, if available
N/A

Unit + E2E Test (will be adding a smoke test for this new construct).

#### Checklist
- [x] PR description included
- [x] `yarn test` passes


packages/amplify-data-construct/remove-module-type-from-package-json.ts
```diff
@@ -0,0 +1,19 @@
+import * as fs from 'fs';
+import * as path from 'path';
```
alharris-at, line 2: this file shouldn't be here.

packages/amplify-data-construct/package.json
```diff
@@ -0,0 +1,167 @@
+{
+  "name": "@aws-amplify/data-construct",
+  "version": "1.3.0",
```
dpilch, line 3: Have you found a way to keep the versions in sync?

packages/amplify-data-construct/API.md
```diff
@@ -0,0 +1,20 @@
+## API Report File for "@aws-amplify/data-construct"
```
dpilch, line 1: Aren't there other types that should be exported from the this package as well?

packages/amplify-data-construct/package.json
alharris-at: Nothing explicit, since lerna manages our versions.

packages/amplify-data-construct/API.md
alharris-at: Yeah, working through that now. I probably won't have a good way to guarantee they're in sync, but it should be pretty do-able.

dpilch (dismissed):

packages/amplify-data-construct/API.md
alharris-at: Added!

PR #2055, opened 2023-11-10
feat: sql lambda provisioned concurrency

dpilch:
#### Description of changes

API

Add `sqlLambdaProvisionedConcurrencyConfig` to `SQLLambdaModelDataSourceStrategy`. The provisioned concurrency will be configured separately for each data source.

```typescript
export interface SQLLambdaModelDataSourceStrategy {

  ...existing params
  
  /**
   * The configuration for the provisioned concurrency of the Lambda.
   */
  readonly sqlLambdaProvisionedConcurrencyConfig?: ProvisionedConcurrencyConfig;
}
```

Set `ProvisionedConcurrencyConfig` to an object to allow for additional params in the future.

```typescript
export interface ProvisionedConcurrencyConfig {
  /** The amount of provisioned concurrency to allocate. **/
  readonly provisionedConcurrentExecutions: number;
}
```

Usage

```typescript
const api = new AmplifyGraphqlApi(stack, 'MyAPI', {                                                                        
  definition: AmplifyGraphqlDefinition.fromString(schema, {                                                                   
    name: 'MySQLDefinition',                                                                                                         
    dbType: 'MYSQL',                                                                                                                 
    dbConnectionConfig,                                                                                                              
    sqlLambdaProvisionedConcurrencyConfig: {                                                                                         
      provisionedConcurrentExecutions: 5,                                                                                            
    },                                                                                                                               
  }),                                                                                                                                
}); 
```

Future Update: Add Auto scaling

https://docs.aws.amazon.com/lambda/latest/dg/provisioned-concurrency.html#managing-provisioned-concurency

```typescript
export interface ProvisionedConcurrencyConfig {
  /** The maximum value to use to scale during a scaling activity. Application auto scaling is enabled when targetValue, maxCapacity, and
   * minCapacity are set. **/
  readonly maxCapacity?: number;

  /** The minimum value to use to scale during a scaling activity. Application auto scaling is enabled when targetValue, maxCapacity, and
   * minCapacity are set. **/
  readonly minCapacity?: number;

  /** Target percentage of provisionedConcurrentExecutions when scaling. Application auto scaling is enabled when targetValue, maxCapacity,
   * and minCapacity are set. **/
  readonly targetValue?: number;

  /** The amount of time, in seconds, after a scale-in activity completes before another scale-in activity can start. Only used wehn
   * application aut scaling is set. **/
  readonly scaleInCooldown?: Duration;

  /** The amount of time, in seconds, to wait for a previous scale-out activity to take effect. Only used when application auto scaling is
   * set.  **/
  readonly scaleOutCooldown?: Duration;

  /** The amount of provisioned concurrency to allocate. **/
  readonly provisionedConcurrentExecutions: number;
}
```

#### Issue #, if available

N/A

* Manual testing
* Unit testing

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-graphql-api-construct-tests/src/__tests__/backends/sql-models/app.ts
```diff
@@ -58,6 +58,9 @@ new AmplifyGraphqlApi(stack, 'SqlBoundApi', {
       dbConnectionConfig: {
         ...dbDetails.ssmPaths,
       },
+      sqlLambdaProvisionedConcurrencyConfig: {
```
dpilch, line 61: I can't get the information I need in order to assert on the provisioned concurrency. I would need to add an additonal CFN output for the SQL Lambda function name. Then I could use the AWS SDK to query for the provisioned concurrency status. I'm not sure adding that is worth it though.

Adding `sqlLambdaProvisionedConcurrencyConfig` here does assert that the API doesn't break when using this setting.

alharris-at: Can't you `synth` and inspect the rendered template?

alharris-at: Also, since you own the test app, you can add your own `CfnExport` for `api.resources.functions.[WhateverTheFunctionNameIs].functionName` if you want to go that route.

packages/amplify-graphql-api-construct/src/internal/construct-exports.ts
```diff
@@ -56,9 +57,14 @@ export const getGeneratedResources = (scope: Construct): AmplifyGraphqlApiResour
...
-    // Retrieve reference name for indexed resources, and bail if none is found.
+    // Retrieve reference name for indexed resources, store resources in fallback if name not found
     const resourceName = getResourceName(currentScope);
-    if (!resourceName) return;
+    if (!resourceName) {
+      if (currentScope instanceof CfnResource) {
+        additionalCfnResourcesWithoutResourceName.push(currentScope);
+      }
+      return;
+    }
```
alharris-at, line 67: Is this just added for testing? I'd prefer we avoid doing this unless we really need it for some reason. If a customer wants to walk the full construct tree they can always do something similar to what we have in this utility.

packages/amplify-graphql-api-construct/src/index.ts
```diff
@@ -1,3 +1,7 @@
+/* ATTENTION
+ *
+ * If you modifiy this file your must also modify packages/amplify-data-construct/src/index.ts to have the same exports
```
alharris-at, line 3: modifiy

alharris-at (approved):

alharris-at (dismissed):

palpatim (commented):
As discussed, we need to ensure that simply adding an alias with provisioned concurrency is enough to wire the AppSync DataSource to use that alias. If it's still pointing to the function URL, not the alias URL, then AppSync queries won't be using the provisioned execution environments.

alharris-at (approved):

palpatim (approved):

PR #2059, opened 2023-11-11
fix: address managed table QA feedbacks

AaronZyLee:
#### Description of changes
Addressed some QA findings

- Used new resource regex for IAM to avoid table deletion failure when type is removed from schema
- Added `DeletionProtectionEnabled` in create table input
- Added `manyToMany` joint table in model datasource strategy map
- Table provision throughput will be the default for newly added GSI without provision defined when billing mode is `PROVISIONED`
- Added E2E tests for schemas by which customers fail the iterative deployment 

Unit & E2E tests(newly added passed locally)
Local CDK app

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes


packages/amplify-graphql-api-construct-tests/src/__tests__/amplify-table-1.test.ts
```diff
@@ -4,12 +4,12 @@ import { cdkDestroy, initCDKProject, cdkDeploy, updateCDKAppWithTemplate } from
 
 jest.setTimeout(1000 * 60 * 60 /* 1 hour */);
 
-describe('CDK amplify table', () => {
+describe('CDK amplify table 1', () => {
```
alharris-at, line 7: Is there a difference between 1 and 2 that'd make more sense to the reader? It's not clear how these test suites or apps are functionally different.

alharris-at (commented):
Overall changes make sense to me. Only comment is on the test naming.

packages/amplify-graphql-api-construct-tests/src/__tests__/amplify-table-1.test.ts
AaronZyLee: The theme is same for both tests. The split is to avoid the test exceeding the 1 hour which requires a credential refresh. The number in description is to stay the same as file name. I use this pattern as we have e2e tests like `api_#`, `schema-iterative-#`

alharris-at (dismissed):

packages/amplify-graphql-api-construct-tests/src/__tests__/amplify-table-1.test.ts
alharris-at: Okay, I'm fine w/ it for now in that case.

PR #2067, opened 2023-11-14
feat: transformer behavior of replacing table upon gsi updates

AaronZyLee:
#### Description of changes
Added feature of replacing table upon GSI changes when both `replaceTableUponGsiUpdate` and `allowDestructiveGraphqlSchemaUpdates` are `true`. This will wipe out all the existing data in the table but cost much less time when doing multiple GSI updates comparing the iterative deployment method.

Unit & e2e tests(newly added passed locally)
Local CDK app
#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes


alharris-at:
I haven't looked at the change in detail yet, but I don't think we should couple this tightly to the existing `sandboxModeEnabled` flag, which is poorly named, but applies to auth behavior specifically. I think a new parameter called `allowDestructiveActionsOnGsiUpdates` or something like that.

alharris-at (commented):
Reviewed inner changes, and it generally makes sense to me. Will approve after making the wiring and flag changes we discussed.

alharris-at (dismissed):

alharris-at:
LGTM - I think you just need to refresh the jsii assembly.

alharris-at (approved):

packages/amplify-graphql-model-transformer/src/resources/amplify-dynamodb-table/amplify-table-manager-lambda/amplify-table-manager-handler.ts
```diff
@@ -74,40 +74,27 @@ export const onEvent = async (event: AWSCDKAsyncCustomResource.OnEventRequest):
...
         } else {
           throw new Error(
             "Editing key schema of table requires replacement which will cause ALL EXISITING DATA TO BE LOST. If this is intended, set the 'allowDestructiveGraphqlSchemaUpdates' to true and deploy again.",
           );
         }
       }
+      // when both sandbox and destructive updates are enabled, gsi update will trigger a table replacement
+      if (tableDef.replaceTableUponGsiUpdate && tableDef.allowDestructiveGraphqlSchemaUpdates) {
+        const nextUpdate = getNextGSIUpdate(describeTableResult.Table, tableDef);
+        if (nextUpdate !== undefined) {
```
phani-srikar, line 91: why do we replace the table only when we have a next update? I don't see where we're using this computed `nextUpdate`?

alharris-at:
Looks like `build-tests` failed. Could you verify the `API.md` file is updated as well?

phani-srikar (approved):

packages/amplify-graphql-model-transformer/src/resources/amplify-dynamodb-table/amplify-table-manager-lambda/amplify-table-manager-handler.ts
AaronZyLee: The `nextUpdate` is a condition here, which indicates the GSI update detected. This is for a special use case for the initial schema design phase when you could do multiple GSI changes and do not care about data entries. The GSIs can be created much faster by using a create table command instead of update table command.

alharris-at (approved):

alharris-at (approved):

PR #2069, opened 2023-11-14
fix: Change 'rds' to 'sql' in public-facing symbols

palpatim:
#### Description of changes

This change leaves most of the internal symbols unchanged, but should pick up the public-facing ones. Notably:

* Imported schema file name
* Stack name
* SQL related CDK resource IDs
* Some error & warning messages

This PR does NOT override ‘RDS’ in resolver template output, since that would bloat the size of the PR and those files aren't generally reviewed by customers anyway.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are [changed or added](https://github.com/aws-amplify/amplify-cli/blob/master/CONTRIBUTING.md#tests)

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #2072, opened 2023-11-15
fix: Rename VPC Endpoint CDK prefix

palpatim:
#### Description of changes

Renames the VPC Endpoint prefix from RDS to SQL. This is not a breaking change, since these endpoints are created and consumed by Amplify. 

Additional changes:
- Fixes an e2e test that was looking for an incorrect Lambda Layer CDK resource name after the RDS -> SQL renaming in #2069.
- Add a transient dynamodb simulator file to the prettier ignore, since it is sometimes not removed if a local test is interrupted before cleaning up
- Remove deprecated `@types/knex` dependency per its deprecation message:
    > npm WARN deprecated @types/knex@0.16.1: This is a stub types
    > definition. knex provides its own type definitions, so you do not need
    > this installed.

#### Checklist

- [x] PR description included
- [x] Modified E2e test passes locally with latest codebase

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -9,6 +9,7 @@ CHANGELOG.md
 client-test-apps
 packages/amplify-category-api/resources/
 packages/amplify-category-api/src/__tests__/commands/api/mock-data/invalid_schema.graphql
+packages/amplify-dynamodb-simulator/emulator/dynamodb-local-metadata.json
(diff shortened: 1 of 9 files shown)


palpatim (commented):
marker review

alharris-at (approved):

phani-srikar (approved):

PR #2079, opened 2023-11-16
fix: regionalize lambda layer patching SNS topics

palpatim:
Other fixes:

- Extract region layer map to JSON file for easier programmatic updates
- Moved several hardcoded resource name strings to ResourceConstants

#### Description of changes

This is step 1 of N in a more comprehensive regionalization strategy for the patching infrastructure. Next step is to investigate the Right Way™ to ensure we have the correct layer at deployment. The CLI flow accomplishes this by downloading the most recent layer map at runtime, but that's not an option for the CDK.

e2e tests running

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


palpatim (commented):
marker review

alharris-at:
Confirming API updates are only destructive to `@experimental` props.

packages/amplify-graphql-model-transformer/src/resources/rds-model-resource-generator.ts
```diff
@@ -31,16 +28,33 @@ export class RdsModelResourceGenerator extends ModelResourceGenerator {
...
+      // resolve the latest layer version. The AwsCustomResource does not work with the CLI custom synth functionality, so we fork the
+      // behavior at this point.
+      let layerVersionArn: string;
+      if (context.rdsLayerMapping) {
+        setRDSLayerMappings(lambdaScope, context.rdsLayerMapping);
+        layerVersionArn = Fn.findInMap(ResourceConstants.RESOURCES.SQLLayerMappingID, Fn.ref('AWS::Region'), 'layerRegion');
+      } else {
+        const layerVersionCustomResource = createLayerVersionCustomResource(lambdaScope);
+        layerVersionArn = layerVersionCustomResource.getResponseField('Body');
+      }
```
alharris-at, line 56: N.p. but I'd prefer to extract the modal behavior, and let/secondary assignment into a separate function, that'll also give you a natural place for this comment.

alharris-at (dismissed):
Comment on structure, but non-blocking.

packages/amplify-category-api/src/graphql-transformer/transform-graphql-schema-v2.ts
```diff
@@ -273,17 +273,13 @@ export const getUserOverridenSlots = (userDefinedSlots: Record<string, UserDefin
...
-      return result as RDSLayerMapping;
-    }
-  } catch (err) {
-    // Ignore the error and return default layer mapping
+  const response = await fetch(LAYER_MAPPING_URL);
+  if (response.status === 200) {
+    const result = await response.json();
+    return result as RDSLayerMapping;
+  } else {
+    throw new Error(`Unable to retrieve layer mapping from ${LAYER_MAPPING_URL} with status code ${response.status}.`);
```
phani-srikar, line 281: We could add a re-try after a pause here to filter out any transient network connection issues. Can be a backlog item.

packages/amplify-graphql-model-transformer/src/resolvers/rds/resolver.ts
```diff
@@ -195,6 +110,41 @@ export const createRdsLambda = (
...
 
+export const createLayerVersionCustomResource = (scope: Construct): AwsCustomResource => {
+  const {
+    SQLLayerVersionCustomResourceID,
+    SQLLayerVersionManifestBucket,
+    SQLLayerVersionManifestBucketRegion,
+    SQLLayerVersionManifestKeyPrefix,
+  } = ResourceConstants.RESOURCES;
+
+  const key = Fn.join('', [SQLLayerVersionManifestKeyPrefix, Fn.ref('AWS::Region')]);
```
phani-srikar, line 121: What happens if there is no entry in the manifest for the customer's region? Do we have a check to fail the deployment early-on?

palpatim: It does fail in the deployment, but not terribly early. If there's a good way to resolve the current region at build time I'm open to adding that validation.

packages/amplify-graphql-model-transformer/src/resources/rds-model-resource-generator.ts
palpatim: Will refactor in the likely followup to fix e2e tests.

alharris-at (dismissed):

alharris-at (dismissed):

packages/amplify-graphql-model-transformer/src/resolvers/rds/resolver.ts
phani-srikar: yeah I don't think there is way to resolve the region in CDK.

phani-srikar (dismissed):

phani-srikar (approved):

packages/amplify-graphql-model-transformer/src/resources/rds-model-resource-generator.ts
```diff
@@ -130,3 +140,25 @@ export class RdsModelResourceGenerator extends ModelResourceGenerator {
...
+const resolveLayerVersion = (scope: Construct, context: TransformerContextProvider): string => {
+  let layerVersionArn: string;
+  if (context.rdsLayerMapping) {
+    setRDSLayerMappings(scope, context.rdsLayerMapping);
+    layerVersionArn = Fn.findInMap(ResourceConstants.RESOURCES.SQLLayerMappingID, Fn.ref('AWS::Region'), 'layerRegion');
+  } else {
+    const layerVersionCustomResource = createLayerVersionCustomResource(scope);
+    layerVersionArn = layerVersionCustomResource.getResponseField('Body');
+  }
+  return layerVersionArn;
```
alharris-at, line 163: super np, but I'd prefer to avoid the `let` entirely, and just return the right type based on 
```suggestion
  if (context.rdsLayerMapping) {
    setRDSLayerMappings(scope, context.rdsLayerMapping);
    return Fn.findInMap(ResourceConstants.RESOURCES.SQLLayerMappingID, Fn.ref('AWS::Region'), 'layerRegion');
  }
  const layerVersionCustomResource = createLayerVersionCustomResource(scope);
  return layerVersionCustomResource.getResponseField('Body');
};
```

alharris-at (approved):

packages/amplify-graphql-model-transformer/src/resources/rds-model-resource-generator.ts
palpatim: Ah, I follow you. I'll add that to a future change

PR #2092, opened 2023-11-27
chore: Refactor transformer internals to use ModelDataSourceStrategyDbType instead of DBType

palpatim:
#### Description of changes

Additional changes:
- Refactor graphql-many-to-many transformer to assign a default DynamoDB strategy to the join table created by that transformer. This is the only place in the internals of the transformer chain where we use a default data source strategy. Everywhere else expects the data source to be explicitly mapped by the transformer invocation.
- Refactor internal database names. All internal database references are now one of the constants:
  - `DDB_DB_TYPE`
  - `MYSQL_DB_TYPE`
  - `POSTGRES_DB_TYPE`
  However, there are still cases where database types are hardcoded, primarily in the Gen1 CLI SQL import flow, where we specify the engine as one of 'MySQL' or 'Postgres', and in the engine passed to the SQL Lambda, where we specify the engine as 'mysql' or 'postgres'.
- Add constants to define DataSourceTypes and ModelDataSourceStrategies for DynamoDB default and DynamoDB Amplify-managed tables. 
  As above, there are still cases where those are hardcoded, primarily in the amplify-graphql-api-construct tests where the customer is specifying the interface values rather than the internal values. We don't want to change those to use constants, so that our tests can continue to assert that the customer-facing interface works even as we change internal handling.
- Make a couple of opportunistic renames from RDS->SQL for internal symbols
- Added a default DynamoDB database strategy to the TestTransform context creator, so that tests don't have to specify a modelToDatasourceMap. 

- Unit tests
- E2E tests running

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/graphql-transformer-core/API.md
```diff
@@ -128,7 +128,7 @@ export interface DataSourceType {
 }
 
 // @public (undocumented)
-export type DBType = 'DDB' | 'MySQL' | 'Postgres';
+export type DBType = 'DYNAMODB' | 'MYSQL' | 'POSTGRES';
```
palpatim, line 131: Updating this for compatibility but it's not in use for CDK or Gen2

palpatim (commented):
marker review

phani-srikar (approved):

alharris-at (approved):

AaronZyLee (approved):

PR #2098, opened 2023-11-29
fix: correctly set pk/indexes when table name changes

dpilch:
#### Description of changes

* correctly set pk/indexes when table name changes
* add test for schema query drift
* Add DISTINCT to pg and mysql schema queries (reduces the size of the schema csv)
* recreate all test schemas

N/A

#### Issue #, if available

N/A

Unit testing

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-graphql-schema-generator/src/__tests__/__snapshots__/generate.test.ts.snap
```diff
@@ -80,6 +113,38 @@ type Story @refersTo(name: \\"stories\\") @model {
...
 
+exports[`generate graphqlSchemaFromSQLSchema creates graphql schema from "postgres" "newsNameChange" schema 1`] = `
+"input AMPLIFY {
+ engine: String = \\"postgres\\"  
+}
+
+
+type Article @refersTo(name: \\"articles\\") @model {
+  pubId: Int! @refersTo(name: \\"pub_id\\") @primaryKey(sortKeyFields: [\\"pubType\\"])
+  pubType: String! @refersTo(name: \\"pub_type\\") @default(value: \\"'A'::bpchar\\")
```
phani-srikar, line 124: The default values generated here don't look right. We would need to remove the `::<type>` part.

packages/amplify-graphql-schema-generator/src/datasource-adapter/pg-string-datasource-adapter.ts
```diff
@@ -118,7 +118,9 @@ export class PostgresStringDataSourceAdapter extends StringDataSourceAdapter {
   }
 
   public getPrimaryKey(tableName: string): Index | null {
-    const key = this.indexes.find((index) => index.tableName === tableName && index.indexName === `${tableName}_pkey`);
+    const key = this.indexes.find(
+      (index) => index.tableName === tableName && index.indexName.match(/.*_pkey/) && index.columns.includes(index.columnName),
```
phani-srikar, line 122: This is still using name matching on index to determine PK. If we have indices named as `my_index_pk` we would would mistake them to be primaryKey constraints?

packages/amplify-graphql-schema-generator/src/__tests__/__snapshots__/generate.test.ts.snap
dpilch: Looks like this an existing bug. I'll fix this in a separate PR.

packages/amplify-graphql-schema-generator/src/datasource-adapter/pg-string-datasource-adapter.ts
dpilch: I believe Postgres would throw an error here when trying to create the table. In our `news` test schema if I change `CONSTRAINT publications_superkey UNIQUE (pub_id, pub_type)` to `CONSTRAINT publications_pkey UNIQUE (pub_id, pub_type)` the query fails with: `ERROR:  relation "publications_pkey" already exists`.

phani-srikar (approved):

palpatim (approved):

PR #2102, opened 2023-11-30
test(api): use beta layer for e2e tests

sundersc:
Use the published beta layer versions to run all the E2E tests. This allows us to publish the layer version to PROD after full E2E testing.

### DIFF
--- a/packages/amplify-category-api/src/graphql-transformer/transform-graphql-schema-v2.ts
+++ b/packages/amplify-category-api/src/graphql-transformer/transform-graphql-schema-v2.ts
@@ -50,7 +50,6 @@ const PARAMETERS_FILENAME = 'parameters.json';
 const SCHEMA_FILENAME = 'schema.graphql';
 const SCHEMA_DIR_NAME = 'schema';
 const PROVIDER_NAME = 'awscloudformation';
-const LAYER_MAPPING_S3_BUCKET = 'amplify-rds-layer-resources';
 const USE_BETA_SQL_LAYER = 'use-beta-sql-layer';
 
 /**
@@ -396,9 +395,9 @@ export const getUserOverridenSlots = (userDefinedSlots: Record<string, UserDefin
     .filter((slotName) => slotName !== undefined);
 
 const getRDSLayerMapping = async (context: $TSContext, useBetaSqlLayer = false): Promise<RDSLayerMapping> => {
-  const bucket = `${LAYER_MAPPING_S3_BUCKET}${useBetaSqlLayer ? '-beta' : ''}`;
+  const bucket = `${ResourceConstants.RESOURCES.SQLLayerVersionManifestBucket}${useBetaSqlLayer ? '-beta' : ''}`;
   const region = context.amplify.getProjectMeta().providers.awscloudformation.Region;
-  const url = `https://${bucket}.s3.amazonaws.com/sql-layer-versions/${region}`;
+  const url = `https://${bucket}.s3.amazonaws.com/${ResourceConstants.RESOURCES.SQLLayerVersionManifestKeyPrefix}${region}`;
   const response = await fetch(url);
   if (response.status === 200) {
     const result = await response.text();
--- a/packages/amplify-category-api/src/graphql-transformer/transform-graphql-schema-v2.ts
+++ b/packages/amplify-category-api/src/graphql-transformer/transform-graphql-schema-v2.ts
@@ -406,7 +406,7 @@ const getRDSLayerMapping = async (context: $TSContext, useBetaSqlLayer = false):
       [region]: {
         layerRegion: result,
       },
-    }
+    };
     return mapping as RDSLayerMapping;
   } else {
     throw new Error(`Unable to retrieve layer mapping from ${url} with status code ${response.status}.`);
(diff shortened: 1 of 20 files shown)


packages/amplify-category-api/src/graphql-transformer/transform-graphql-schema-v2.ts
```diff
@@ -43,6 +43,8 @@ const SCHEMA_FILENAME = 'schema.graphql';
 const SCHEMA_DIR_NAME = 'schema';
 const PROVIDER_NAME = 'awscloudformation';
 const LAYER_MAPPING_URL = 'https://amplify-rds-layer-resources.s3.amazonaws.com/rds-layer-mapping.json';
+const LAYER_MAPPING_URL_BETA = 'https://amplify-rds-layer-resources-beta.s3.amazonaws.com/rds-layer-mapping.json';
```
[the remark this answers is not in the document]
sundersc: That isn't accurate. Gen1 CLI is still using this layer mapping file. The regionalized S3 layer version files are currently used by the CDK construct only.

I tried to use the custom resource in Gen 1 CLI but the custom resource is looking for S3 deployment bucket and root key as parameters in child stack instead of using the root stack parameters.

In long term, we need to migrate all our E2E tests to use the CDK construct.

dpilch (approved):

dpilch (approved):

palpatim:
The only failing e2e on this is `relationships_sql_models`, which is fixed on #2115

packages/amplify-category-api/src/graphql-transformer/transform-graphql-schema-v2.ts
```diff
@@ -394,13 +395,21 @@ export const getUserOverridenSlots = (userDefinedSlots: Record<string, UserDefin
     .flat()
     .filter((slotName) => slotName !== undefined);
 
-const getRDSLayerMapping = async (): Promise<RDSLayerMapping> => {
-  const response = await fetch(LAYER_MAPPING_URL);
+const getRDSLayerMapping = async (context: $TSContext, useBetaSqlLayer = false): Promise<RDSLayerMapping> => {
+  const bucket = `${LAYER_MAPPING_S3_BUCKET}${useBetaSqlLayer ? '-beta' : ''}`;
```
palpatim, line 399: Let's add this to `ResourceConstants` with the other patching constants.

packages/amplify-category-api/src/graphql-transformer/transform-graphql-schema-v2.ts
```diff
@@ -394,13 +395,21 @@ export const getUserOverridenSlots = (userDefinedSlots: Record<string, UserDefin
...
+  const region = context.amplify.getProjectMeta().providers.awscloudformation.Region;
+  const url = `https://${bucket}.s3.amazonaws.com/sql-layer-versions/${region}`;
```
palpatim, line 401: get the path name from ResourceConstants.

packages/amplify-category-api/src/graphql-transformer/transform-graphql-schema-v2.ts
```diff
@@ -50,7 +50,8 @@ const PARAMETERS_FILENAME = 'parameters.json';
 const SCHEMA_FILENAME = 'schema.graphql';
 const SCHEMA_DIR_NAME = 'schema';
 const PROVIDER_NAME = 'awscloudformation';
-const LAYER_MAPPING_URL = 'https://amplify-rds-layer-resources.s3.amazonaws.com/rds-layer-mapping.json';
+const LAYER_MAPPING_S3_BUCKET = 'amplify-rds-layer-resources';
+const USE_BETA_SQL_LAYER = 'use-beta-sql-layer';
```
palpatim, line 54: Here and throughout: Put these in ResourceConstants alongside the other patching lambda constants. That way we centralize the value and can access it easily from tests that don't need to assert the actual constant values

packages/amplify-e2e-core/src/init/amplifyPush.ts
```diff
@@ -49,7 +50,15 @@ export function amplifyPush(
         }
       });
     // Test amplify push
-    const pushCommands = spawn(getCLIPath(testingWithLatestCodebase), ['push'], { cwd, stripColors: true, noOutputTimeout: pushTimeoutMS })
+    const pushOptions = ['push'];
+    if (settings?.useBetaSqlLayer) {
+      pushOptions.push('--use-beta-sql-layer');
```
palpatim, line 55: I assume this flag is hidden from all customer-facing usages (e.g., help screens, documentation, manifests, etc), so that we can use it for our E2E tests without exposing the feature to customers?

packages/amplify-e2e-tests/src/rds-v2-tests-common/sql-e2e-config.ts
```diff
@@ -0,0 +1,3 @@
+// Setting SQL_TESTS_USE_BETA flag to 'true' withh test the CLI Gen1 tests with RDS Beta Layers.
```
palpatim, line 1:
```suggestion
// Setting SQL_TESTS_USE_BETA flag to 'true' with test the CLI Gen1 tests with RDS Beta Layers.
```

palpatim (commented):
If I read this correctly, the new flag gives us the opportunity to run E2Es in our layer pipeline against the beta layer, as well as prod validations against the prod layer, is that right?

packages/amplify-e2e-core/src/init/amplifyPush.ts
sundersc: Yes, that's correct. It is not exposed to customers.

packages/amplify-category-api/src/graphql-transformer/transform-graphql-schema-v2.ts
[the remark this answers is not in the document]
sundersc: Updated the PR to use the regionalized S3 manifests. Now Gen1 and CDK construct uses the regionalized S3 manifests.

packages/amplify-category-api/src/graphql-transformer/transform-graphql-schema-v2.ts
sundersc: I've reused the existing constant for the SQL Layer S3 manifest bucket. I left the CLI Gen1 option in the same file as this is not related to the transformers.

palpatim (approved):

dpilch (approved):

PR #2113, opened 2023-12-04
chore(api): improve overall test coverage report accuracy

sundersc:
Current jest coverage provider (babel) doesn't provide correct test coverage report. This PR changes the coverage provider to 'v8' which provides more accurate report.

**How did I verify `coverageProvider: v8` is accurate?**
 Default coverage provider (babel) showed some obvious straight forward blocks (such as vtl generation helper) as not covered. However, adding logs and breakpoints confirmed that those lines are covered. Changing the coverage provider to v8 produced more accurate report. I took few uncovered lines from v8 report and confirmed that tests are actually missing for those blocks.

### DIFF
--- a/packages/amplify-graphql-transformer-interfaces/package.json
+++ b/packages/amplify-graphql-transformer-interfaces/package.json
@@ -48,6 +48,7 @@
       "node"
     ],
     "collectCoverage": true,
+    "coverageProvider": "v8",
     "coverageReporters": [
       "clover",
       "text"
--- a/packages/amplify-util-mock/package.json
+++ b/packages/amplify-util-mock/package.json
@@ -94,6 +94,14 @@
       "^.+\\.tsx?$": "ts-jest"
     },
     "collectCoverage": true,
+    "coverageProvider": "v8",
+    "coverageThreshold": {
+      "global": {
+        "branches": 62,
+        "functions": 37,
+        "lines": 16
+      }
+    },
     "collectCoverageFrom": [
       "src/**/*.ts",
       "!**/node_modules/**",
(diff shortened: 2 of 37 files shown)


dpilch (approved):

dpilch (approved):

phani-srikar (approved):

packages/amplify-graphql-transformer-interfaces/package.json
```diff
@@ -48,6 +48,7 @@
       "node"
     ],
     "collectCoverage": true,
+    "coverageProvider": "v8",
     "coverageReporters": [
       "clover",
       "text"
```
palpatim, line 54: Understanding that this is an interface package and has no executable code, should we defensively code a threshold limit here to catch any future changes, or is there a global setting somewhere that applies in the absence of a specific configuration for this package?

packages/amplify-util-mock/package.json
```diff
@@ -94,6 +94,7 @@
       "^.+\\.tsx?$": "ts-jest"
     },
     "collectCoverage": true,
+    "coverageProvider": "v8",
```
palpatim, line 97: This package does have executable code. What limits apply?

packages/amplify-graphql-transformer-interfaces/package.json
sundersc: No, we do not have any global threshold. Since most of these types (or constants) are used in other projects, we do not have any threshold check for this package.

packages/graphql-transformers-e2e-tests/package.json
```diff
@@ -67,6 +67,7 @@
     },
     "testEnvironment": "node",
     "collectCoverage": true,
+    "coverageProvider": "v8",
```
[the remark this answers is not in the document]
sundersc: Since our current jest configuration is set to 'collectCoverage' to 'true', I'm changing the coverage provider to 'v8'. However we do not enforce any threshold for this package.

PR #2129, opened 2023-12-08
Schema validations

bzsurbhi:
#### Description of changes
Added schema validation for 
- manyToMany directive should have a relation name
- relationship name should not be inverse of relationName
- hasMany must only be used with with lists
- hasOne must not be used with lists
- object must be annotated with @model directive

NA

- added tests

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-schema-validator/src/validators/hasMany-must-be-used-with-lists.ts
```diff
@@ -0,0 +1,30 @@
...
+  ) as ObjectTypeDefinitionNode[];
+  objectTypeDefinitions.forEach((objectTypeDefinition) => {
+    const directiveFields = objectTypeDefinition.fields?.filter((objectField) =>
+      objectField.directives?.find((directive) => directive.name.value === 'hasMany'),
+    );
+
+    directiveFields?.forEach((directiveField) => {
+      const listType = isListType(directiveField.type);
+      if (!listType) {
+        errors.push(new InvalidDirectiveError('@hasMany must be used with a list. Use @hasOne for non-list types.'));
```
phani-srikar, line 25: Is this error something we surface to the customers? Adding which field and model is being referred to in the error message would make it easier to debug.

packages/amplify-schema-validator/src/validators/hasOne-cannot-be-used-with-lists.ts
```diff
@@ -0,0 +1,30 @@
...
+  ) as ObjectTypeDefinitionNode[];
+  objectTypeDefinitions.forEach((objectTypeDefinition) => {
+    const directiveFields = objectTypeDefinition.fields?.filter((objectField) =>
+      objectField.directives?.find((directive) => directive.name.value === 'hasOne'),
+    );
+
+    directiveFields?.forEach((directiveField) => {
+      const listType = isListType(directiveField.type);
+      if (listType) {
+        errors.push(new InvalidDirectiveError('@hasOne cannot be used with lists. Use @hasMany instead'));
```
phani-srikar, line 25: same as above. Adding which field and model is being referred to in the error message would make it easier to debug.

bzsurbhi: fixed

packages/amplify-schema-validator/src/validators/hasMany-must-be-used-with-lists.ts
bzsurbhi: fixed

phani-srikar (approved):

dpilch (approved):

PR #2130, opened 2023-12-09
fix(api): validate imported enum values

sundersc:
#### Description of changes

Add a validation step to make sure that the imported ENUM values are valid GraphQL enum values. Otherwise, throw an error.

No.

#### Issue #, if available
NA

- Manual test
- Unit test

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-graphql-schema-generator/src/schema-generator/generate-schema.ts
+++ b/packages/amplify-graphql-schema-generator/src/schema-generator/generate-schema.ts
@@ -189,6 +189,11 @@ const constructObjectType = (model: Model) => {
 };
 
 const constructEnumType = (type: EnumType): EnumWrapper => {
+  // Check if the enum values imported from the database contain any invalid character.
+  // Enum can contain A-Z, a-z, 0-9 or underscore(_) and must begin with an alphabet or an underscore(_).
+  if (!validateEnumValues(type.values)) {
+    throw new Error(`Enum "${type.name}" (values: ${type.values.join(',')}) contains one or more invalid values. Enum values can contain A-Z, a-z, 0-9 or underscore(_) and must begin with an alphabet or an underscore(_).`);
+  }
   const enumValues = type.values.map((t) => {
     return {
       kind: Kind.ENUM_VALUE_DEFINITION,
@@ -208,6 +213,12 @@ const constructEnumType = (type: EnumType): EnumWrapper => {
   });
   return enumType;
 };
+
+const validateEnumValues = (values: string[]): boolean => {
+ const regex = new RegExp(/^[_A-Za-z][_0-9A-Za-z]*$/);
+ const containsValidValues = values.every((value) => regex.test(value));
+ return containsValidValues;
+}
 
 const addIndexes = (type: ObjectDefinitionWrapper, indexes: Index[]): void => {
   indexes.forEach((index) => {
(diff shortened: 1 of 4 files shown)


dpilch (approved):

packages/amplify-graphql-transformer-core/package.json
```diff
@@ -69,7 +69,7 @@
     "coverageThreshold": {
       "global": {
         "branches": 69,
-        "functions": 32,
+        "functions": 31,
```
sundersc, line 72: This package hasn't been modified in this PR. Looks this is the current coverage in 'main' branch.

dpilch (approved):

packages/amplify-graphql-schema-generator/src/schema-generator/generate-schema.ts
```diff
@@ -189,6 +189,15 @@ const constructObjectType = (model: Model) => {
 };
 
 const constructEnumType = (type: EnumType): EnumWrapper => {
+  // Check if the enum values imported from the database contain any invalid character.
+  // Enum can contain A-Z, a-z, 0-9 or underscore(_) and must begin with an alphabet or an underscore(_).
+  if (!validateEnumValues(type.values)) {
+    throw new Error(
+      `Enum "${type.name}" (values: ${type.values.join(
+        ',',
+      )}) contains one or more invalid values. Enum values can contain A-Z, a-z, 0-9 or underscore(_) and must begin with an alphabet or an underscore(_).`,
```
palpatim, line 198: Let's get confirmation on the error message. The prose representation of the rules is probably _less_ understandable to devs than something like "...must match the regex `[A-Za-z_][A-Za-z0-9_]+`"

packages/amplify-graphql-schema-generator/src/schema-generator/generate-schema.ts
```diff
@@ -209,6 +218,12 @@ const constructEnumType = (type: EnumType): EnumWrapper => {
   return enumType;
 };
 
+const validateEnumValues = (values: string[]): boolean => {
+  const regex = new RegExp(/^[_A-Za-z][_0-9A-Za-z]*$/);
```
palpatim, line 222: I believe the spec requires there to be at least one `NameContinue` after `NameStart`, so the regex quantifier should be `+` rather than `*`:

```suggestion
  const regex = new RegExp(/^[_A-Za-z][_0-9A-Za-z]+$/);
```

We should add a test for single-character enum values to validate.

packages/amplify-graphql-schema-generator/src/schema-generator/generate-schema.ts
sundersc: Makes sense. Updated the error message to regex.

packages/amplify-graphql-schema-generator/src/schema-generator/generate-schema.ts
sundersc: I don't see an issue with the regex expression. There are two character classes in this regex expression. [_A-Za-z] verifies the first character (NameStart) and [_0-9A-Za-z]* confirms the rest of the string value (NameContinue). Added a test with single character enum value.

I have confirmed with the GraphQL 2018 spec (that's what AppSync supports) as well.
https://spec.graphql.org/June2018/#Name

phani-srikar (approved):

dpilch (approved):

PR #2166, opened 2023-12-29
test: test permutations of auth strategies & operations

palpatim:
#### Description of changes

Adds tests for iterating over single combinations of of auth strategies & operations, to cover some cases I couldn't find elsewhere. Also adds field-level auth rule combination tests. Adds support for generating snapshots as a developer tool for validating changes that impact resolver or stack resource generation.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-graphql-auth-transformer/src/__tests__/auth-operations.test.ts
```diff
@@ -351,3 +366,157 @@ test('can update on model without delete', () => {
...
+          const schema = schemaTemplate.replace('<AUTH_RULE>', resolvedAuthRule);
+          const resolverConfig: ResolverConfig | undefined = operation === 'sync' ? { project: syncConfig } : undefined;
+
+          const out = testTransform({
+            schema,
+            authConfig: { defaultAuthentication: rule.auth, additionalAuthenticationProviders: [] },
+            transformers,
+            resolverConfig,
+            dataSourceStrategies: constructDataSourceStrategies(schema, strategy),
+          });
```
phani-srikar, line 515: `sync` and `search` are not supported for SQL datasource strategies. But the fact that the transformation succeeds here probably means we're missing those validations.

packages/amplify-graphql-auth-transformer/src/utils/validations.ts
```diff
@@ -94,13 +100,45 @@ found '${rule.provider}' assigned.`,
   }
 };
 
-export const validateRules = (rules: AuthRule[], configuredAuthProviders: ConfiguredAuthProviders, typeName: string) => {
+export const validateRules = (
+  rules: AuthRule[],
+  configuredAuthProviders: ConfiguredAuthProviders,
+  typeName: string,
+  context: TransformerSchemaVisitStepContextProvider,
```
palpatim: I'm a bit uncomfortable plumbing the transformer DataSourceProvider/transformer context through to this level, since we've avoided coupling up until now. But the validation needs to be able to tell if these are SQL strategies, and it's cleaner to centralize validations in a single file rather than spreading them through the transformer code, so this seemed like a compromise I could live with.

phani-srikar (dismissed):

phani-srikar (dismissed):

phani-srikar (dismissed):

sundersc (approved):

AaronZyLee (approved):

PR #2194, opened 2024-01-11
test(api): rds e2e field auth tests for iam provider

sundersc:
Add E2E tests for RDS MySQL and Postgres field level auth rules for IAM provider.

Changes are tested by running the new E2E tests locally and codebuild.

### DIFF
--- a/packages/amplify-e2e-tests/src/rds-v2-tests-common/rds-field-auth-iam.ts
+++ b/packages/amplify-e2e-tests/src/rds-v2-tests-common/rds-field-auth-iam.ts
@@ -1,0 +1,636 @@
+import {
+  addApiWithAllAuthModes,
+  amplifyPush,
+  apiGqlCompile,
+  createNewProjectDir,
+  deleteDBInstance,
+  deleteProject,
+  deleteProjectDir,
+  enableUserPoolUnauthenticatedAccess,
+  getAppSyncApi,
+  getProjectMeta,
+  importRDSDatabase,
+  initJSProjectWithProfile,
+  setupRDSInstanceAndData,
+  sleep,
+} from 'amplify-category-api-e2e-core';
+import { existsSync, writeFileSync } from 'fs-extra';
+import generator from 'generate-password';
+import path from 'path';
+import { GQLQueryHelper } from '../query-utils/gql-helper';
+import { configureAmplify, getConfiguredAppsyncClientIAMAuth, getUserPoolId, setupUser, signInUser } from '../schema-api-directives';
+import { ImportedRDSType } from '@aws-amplify/graphql-transformer-core';
+import { SQL_TESTS_USE_BETA } from './sql-e2e-config';
+import { getDefaultDatabasePort } from '../rds-v2-test-utils';
+import { API, Auth } from 'aws-amplify';
+import gql from 'graphql-tag';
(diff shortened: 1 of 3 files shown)


packages/amplify-e2e-tests/src/rds-v2-tests-common/rds-field-auth-iam.ts
```diff
@@ -0,0 +1,636 @@
...
+      const dbConfig = {
+        identifier,
+        engine,
+        dbname: database,
+        username,
+        password,
+        region,
+      };
+
+      console.log(JSON.stringify(dbConfig, null, 4));
```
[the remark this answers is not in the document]
sundersc: This has been removed.

packages/amplify-e2e-tests/src/rds-v2-tests-common/rds-field-auth-iam.ts
```diff
@@ -0,0 +1,634 @@
...
+      expect(graphqlApi).toBeDefined();
+      expect(graphqlApi.apiId).toEqual(GraphQLAPIIdOutput);
+
+      await createAppSyncClients(apiEndPoint, appRegion);
+    });
+
+    const createAppSyncClients = async (apiEndPoint, appRegion): Promise<void> => {
+      configureAmplify(projRoot);
+      const unAuthCredentials = await Auth.currentCredentials();
+      const [cognito_username, cognito_password] = ['test@test.com', 'Password123!'];
```
palpatim, line 97: very nit, but why snake case instead of camel case?

packages/amplify-e2e-tests/src/rds-v2-tests-common/rds-field-auth-iam.ts
```diff
@@ -0,0 +1,634 @@
...
+        input AMPLIFY {
+          engine: String = "${engineName}"
+        }
+
+        # IAM Public can't delete a record
+        # IAM Public can't update SSN field
+        # IAM Public can't read SSN field 
+        # IAM Public can create a record with all fields, but can't read SSN response
+        # IAM Public subscription can't read SSN field
+        type Person1 @model @auth(rules: [{allow: private, provider: iam}, {allow: public, provider: iam}]) {
```
palpatim, line 176: For all Person* type tests: it looks like we're not testing that the private client has full access while the public client is appropriately restricted. Is the private client behavior covered elsewhere? We've seen odd behaviors with combinations in the past.

phani-srikar (approved):

packages/amplify-e2e-tests/src/rds-v2-tests-common/rds-field-auth-iam.ts
sundersc: Though it is not explicitly mentioned, the test also covers 'get', 'list' and 'create' operation on IAM private client. I will add test cases for 'update' and 'delete' for IAM::private.

packages/amplify-e2e-tests/src/rds-v2-tests-common/rds-field-auth-iam.ts
sundersc: It carried forward from our existing tests, I can correct it on the follow-up PR.

palpatim (approved):

PR #2204, opened 2024-01-18
fix: support 'read' operation alias for SQL data sources

palpatim:
#### Description of changes

Support 'read' operation alias for SQL data sources:

```graphql
type Todo @model @auth(rules: [{allow: owner, operations: [read]}]) { ... }
```

This change is enabled by adding an `isSqlDataSource` option to the `getAuthDirectiveRules` utility function. That function is exported from the V2 transformer and so changes the API.md, but does not change the customer-facing API.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-graphql-auth-transformer/src/utils/index.ts
+++ b/packages/amplify-graphql-auth-transformer/src/utils/index.ts
@@ -47,9 +47,11 @@ export const getAuthDirectiveRules = (authDir: DirectiveWrapper, options?: GetAu
       operations.splice(indexOfRead, 1);
       operations.push('get');
       operations.push('list');
-      operations.push('search');
-      operations.push('sync');
       operations.push('listen');
+      if (!options.isSqlDataSource) {
+        operations.push('search');
+        operations.push('sync');
+      }
       // eslint-disable-next-line no-param-reassign
       rule.operations = operations as ModelOperation[];
     }
--- a/packages/amplify-graphql-auth-transformer/src/utils/index.ts
+++ b/packages/amplify-graphql-auth-transformer/src/utils/index.ts
@@ -48,7 +48,7 @@ export const getAuthDirectiveRules = (authDir: DirectiveWrapper, options?: GetAu
       operations.push('get');
       operations.push('list');
       operations.push('listen');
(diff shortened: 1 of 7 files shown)


sundersc (approved):

packages/amplify-graphql-auth-transformer/src/utils/index.ts
```diff
@@ -47,9 +47,11 @@ export const getAuthDirectiveRules = (authDir: DirectiveWrapper, options?: GetAu
       operations.splice(indexOfRead, 1);
       operations.push('get');
       operations.push('list');
-      operations.push('search');
-      operations.push('sync');
       operations.push('listen');
+      if (!options.isSqlDataSource) {
```
phani-srikar, line 51: Since `options` is an optional input, this will fail if it's not passed. Would recommend `if (!options?.isSqlDataSource) {`

AaronZyLee (approved):
LGTM. Could you double check on the ACM for a sql model by running `amplify-dev status api --acm ${modelName}`? The expected matrix should be `true` for `get, list, listen` and `false` for `sync, search`.

palpatim:
> LGTM. Could you double check on the ACM for a sql model by running `amplify-dev status api --acm ${modelName}`? The expected matrix should be `true` for `get, list, listen` and `false` for `sync, search`.

ACM support isn't enabled yet. The command line ACM doesn't have any context for knowing the data source, since it operates on the bare schema. We'll have to think about if/how we enable support for specifying a data source in the command line ACM.

packages/amplify-graphql-auth-transformer/src/utils/index.ts
palpatim: Thanks for catching that! I wonder why the ACM tests are passing, since one of the `getAuthDirectiveRules` invocations inside of `printAcm` doesn't pass args. (And for that matter, I wonder why `tsc` didn't complain about this?)

AaronZyLee (approved):

sundersc (approved):

svidgen (approved):

PR #2206, opened 2024-01-18
chore(api): add cleanup script to delete rds instances

sundersc:
Add cleanup script to delete the RDS instances from the E2E account.
For now, we will run this manually when required.

### DIFF
--- a/packages/amplify-e2e-tests/src/cleanup-e2e-rds-instances.sh
+++ b/packages/amplify-e2e-tests/src/cleanup-e2e-rds-instances.sh
@@ -1,0 +1,16 @@
+declare -a regions=( 'us-east-1' 'us-east-2' 'us-west-2' 'eu-west-2' 'eu-central-1' 'ap-northeast-1' 'ap-southeast-1' 'ap-southeast-2') ## now loop through the above array
+for region in "${regions[@]}"
+do
+   echo "Deleting all the RDS instances in $region"
+   aws rds describe-db-instances --region $region \
+    --query 'DBInstances[*].[DBInstanceIdentifier]' \
+    --output text | xargs -I {} \
+      aws rds delete-db-instance \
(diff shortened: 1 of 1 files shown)


packages/amplify-e2e-tests/src/cleanup-e2e-rds-instances.sh
```diff
@@ -0,0 +1,16 @@
...
+    --query 'DBInstances[*].[DBInstanceIdentifier]' \
+    --output text | xargs -I {} \
+      aws rds delete-db-instance \
+         --region $region \
+         --db-instance-identifier {} \
+         --skip-final-snapshot \
+         --query 'DBInstance.DBInstanceIdentifier' \
+         --delete-automated-backups \
+         --output text | xargs -I {} \
+      echo "--- Deleting DB instance" {}
```
phani-srikar, line 15: nit: "Deleted" since this runs at the end?

phani-srikar (approved):

packages/amplify-e2e-tests/src/cleanup-e2e-rds-instances.sh
sundersc: We only initiate the delete here, we do not await until the deletion is complete.

packages/amplify-e2e-tests/src/cleanup-e2e-rds-instances.sh
```diff
@@ -0,0 +1,16 @@
+declare -a regions=( 'us-east-1' 'us-east-2' 'us-west-2' 'eu-west-2' 'eu-central-1' 'ap-northeast-1' 'ap-southeast-1' 'ap-southeast-2') ## now loop through the above array
```
dpilch, line 1: nit: the file does not have executable permissions set

Issue #1978, opened 2023-10-20
@aws-amplify/graphql-api-construct   call stack size exceeded in Lodash.js


jimjiminyjimjim:
### How did you install the Amplify CLI?

yarn add @aws-amplify/graphql-api-construct

### If applicable, what version of Node.js are you using?

v18.14.2

### Amplify CLI Version

12.2.5

### What operating system are you using?

Mac 14.0

### Did you make any manual changes to the cloud resources managed by Amplify? Please describe the changes made.

No manual changes made

### Describe the bug

I get the following error while building my schema with the new cdk construct:

`/Users/jimjiminyjimjim/web_projects/amplifyCDK/node_modules/@aws-amplify/graphql-api-construct/node_modules/lodash/lodash.js:6598
      return nativeObjectToString.call(value);
                                  ^
RangeError: Maximum call stack size exceeded
    at Token.toString (<anonymous>)
    at objectToString (/Users/jimjiminyjimjim/web_projects/amplifyCDK/node_modules/@aws-amplify/graphql-api-construct/node_modules/lodash/lodash.js:6598:35)
    at baseGetTag (/Users/jimjiminyjimjim/web_projects/amplifyCDK/node_modules/@aws-amplify/graphql-api-construct/node_modules/lodash/lodash.js:3111:11)
    at baseClone (/Users/jimjiminyjimjim/web_projects/amplifyCDK/node_modules/@aws-amplify/graphql-api-construct/node_modules/lodash/lodash.js:2684:19)
    at /Users/jimjiminyjimjim/web_projects/amplifyCDK/node_modules/@aws-amplify/graphql-api-construct/node_modules/lodash/lodash.js:2733:34
    at arrayEach (/Users/jimjiminyjimjim/web_projects/amplifyCDK/node_modules/@aws-amplify/graphql-api-construct/node_modules/lodash/lodash.js:530:11)
    at baseClone (/Users/jimjiminyjimjim/web_projects/amplifyCDK/node_modules/@aws-amplify/graphql-api-construct/node_modules/lodash/lodash.js:2727:7)
    at /Users/jimjiminyjimjim/web_projects/amplifyCDK/node_modules/@aws-amplify/graphql-api-construct/node_modules/lodash/lodash.js:2733:34
    at arrayEach (/Users/jimjiminyjimjim/web_projects/amplifyCDK/node_modules/@aws-amplify/graphql-api-construct/node_modules/lodash/lodash.js:530:11)
[...]

PR #2212, opened 2024-01-23
fix: deep merge default behavior

phani-srikar:
#### Description of changes

This PR optimizes the merging of default and user provided directive arguments by avoiding deep cloning of default values in cases where:
- No user provided args exist. In this case, we essentially just return the default values as is.
- No common properties exist. In this case, we include properties from both inputs and no need for recursively merging them.

This fixes the https://github.com/aws-amplify/amplify-category-api/issues/1978 without changing the default behavior of the construct. The stack size error is being thrown by lodash's deep clone utility due to some circular references in the nested object which includes GraphQL AST nodes with multiple sub-nodes within them. These don't imply any relational circular dependencies between models in the schema.

Added unit tests which also include the [examples from docs](https://docs.amplify.aws/react/build-a-backend/graphqlapi/data-modeling/#rename-generated-queries-mutations-and-subscriptions) that were missing previously.
[E2Es are Green](https://us-east-1.console.aws.amazon.com/codesuite/codebuild/594813022831/projects/amplify-category-api-e2e-workflow/batch/amplify-category-api-e2e-workflow:7e2943aa-c553-4cf2-a544-a43319ac829f?region=us-east-1#).

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-graphql-transformer-core/src/utils/directive-wrapper.ts
+++ b/packages/amplify-graphql-transformer-core/src/utils/directive-wrapper.ts
@@ -54,8 +54,9 @@ export class DirectiveWrapper {
       }),
       {},
     );
-    if (options?.deepMergeArguments) {
-      return _.merge(_.cloneDeep(defaultValue), argValues);
+    if (options?.deepMergeArguments && !_.isEmpty(argValues)) {
+      const clonedDefaultArgValues = _.cloneDeep(defaultValue);
+      return _.merge(clonedDefaultArgValues, argValues);
     }
     return Object.assign(defaultValue, argValues);
   };
--- a/packages/amplify-graphql-transformer-core/src/utils/directive-wrapper.ts
+++ b/packages/amplify-graphql-transformer-core/src/utils/directive-wrapper.ts
@@ -54,9 +54,8 @@ export class DirectiveWrapper {
       }),
       {},
     );
-    if (options?.deepMergeArguments && !_.isEmpty(argValues)) {
-      const clonedDefaultArgValues = _.cloneDeep(defaultValue);
-      return _.merge(clonedDefaultArgValues, argValues);
+    if (options?.deepMergeArguments && needsDeepMerge(defaultValue, argValues)) {
+      return _.merge(_.cloneDeep(defaultValue), argValues);
     }
     return Object.assign(defaultValue, argValues);
   };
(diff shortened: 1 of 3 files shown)


packages/amplify-graphql-transformer-core/src/utils/directive-wrapper.ts
```diff
@@ -54,7 +54,7 @@ export class DirectiveWrapper {
       }),
       {},
     );
-    if (options?.deepMergeArguments) {
+    if (options?.deepMergeArguments && needsDeepMerge(defaultValue, argValues)) {
       return _.merge(_.cloneDeep(defaultValue), argValues);
```
sundersc, line 58: Do we need 'lodash' here? Can't we do this with object destructuring?

sundersc:
Can you add some details on why recursive call resulting in call stack overflow error. Is there a circular reference in the object? or Is there an error on our code?

packages/amplify-graphql-transformer-core/src/utils/directive-wrapper.ts
phani-srikar: There are more than 2 levels in the `defaultValue` that is set by some of the transformers (for example [this](https://github.com/aws-amplify/amplify-category-api/blob/3564898d75c777f6fdf26b8842eed10a179bdf0f/packages/amplify-graphql-index-transformer/src/graphql-index-transformer.ts#L51)) where we set the object and field def GraphQL AST nodes as defaultValues. The `...` spread operator would copy the nested items by reference. This might be okay in majority of cases, but since this is used in multiple transformations, keeping the original deep cloning here. I am not sure how object destructuring is useful here.

AaronZyLee (approved):
The change LGTM. Just have another question for the issue. Looks like it is only reported in the CDK construct rather than using Amplify CLI gen1, whereas the problem seem to be related to transformer which should affect Gen1 as well. Did you see the same issue when using the Gen1?

phani-srikar:
> The change LGTM. Just have another question for the issue. Looks like it is only reported in the CDK construct rather than using Amplify CLI gen1, whereas the problem seem to be related to transformer which should affect Gen1 as well. Did you see the same issue when using the Gen1?

The issue occurs with Gen-1 CLI too when the feature flag `shouldDeepMergeDirectiveConfigDefaults` is set to `true`.

packages/amplify-graphql-transformer-core/src/utils/directive-wrapper.ts
```diff
@@ -64,3 +64,20 @@ export class DirectiveWrapper {
...
+ * These include cases when the user provided arguments are empty or when there are no common keys between the default and user provided arguments.
+ * @param defaultValue the default properties set for the directive
+ * @param argValues the user provided arguments for the directive
+ * @returns if deep cloning and merging of arguments is needed
+ */
+export const needsDeepMerge = <T>(defaultValue: Required<T>, argValues: { [x: string]: any }): boolean => {
+  if (_.isEmpty(argValues)) {
+    return false;
+  }
+  if (typeof defaultValue === 'object') {
```
sundersc, line 79: Won't this be always true?

phani-srikar: Since `defaultValue` is a generic, this will filter out if any scalars like strings are passed to it.

sundersc: Can we say `needsDeepMerge = <T extends object>(defaultValue: Required<T>, argValues: { [x: string]: any })`?

sundersc (approved):

PR #2233, opened 2024-01-31
feat: add credential retry logic to sql lambda

dpilch:
#### Description of changes

Add credential refresh and retry logic to SQL lambda. 

This is merging into a feature branch.

N/A

#### Issue #, if available

N/A

Manual testing. E2E tests will come in subsequent PR.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-graphql-model-transformer/rds-lambda/handler.ts
+++ b/packages/amplify-graphql-model-transformer/rds-lambda/handler.ts
@@ -47,10 +47,10 @@ const retryWithRefreshedCredentials = async (event, debugMode): Promise<any> => 
 
 const isRetryableError = (error): boolean => {
   // https://www.postgresql.org/docs/current/errcodes-appendix.html
-  const postgresRetryableError = error.code === '28P01' && error.routine === 'auth_failed';
+  const postgresRetryableError = error.code === '28P01';
 
   // https://dev.mysql.com/doc/mysql-errors/8.0/en/server-error-reference.html
-  const mysqlRetryableError = error.errno === '1045' && error.code === 'ER_ACCESS_DENIED_ERROR';
+  const mysqlRetryableError = error.errno === '1045';
 
   return postgresRetryableError || mysqlRetryableError;
 }
--- a/packages/amplify-graphql-model-transformer/rds-lambda/handler.ts
+++ b/packages/amplify-graphql-model-transformer/rds-lambda/handler.ts
@@ -46,7 +46,10 @@ const retryWithRefreshedCredentials = async (event, debugMode): Promise<any> => 
 };
 
 const isRetryableError = (error): boolean => {
+  // https://www.postgresql.org/docs/current/errcodes-appendix.html
   const postgresRetryableError = error.code === '28P01' && error.routine === 'auth_failed';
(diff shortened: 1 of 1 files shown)


packages/amplify-graphql-model-transformer/rds-lambda/handler.ts
```diff
@@ -24,10 +24,34 @@ export const run = async (event): Promise<any> => {
...
+    adapter = null;
+    throw err;
+  }
 };
 
+const isRetryableError = (error): boolean => {
+  const postgresRetryableError = error.code === '28P01' && error.routine === 'auth_failed';
+  const mysqlRetruableError = error.errno === '1045' && error.code === 'ER_ACCESS_DENIED_ERROR';
+
+  return postgresRetryableError || mysqlRetruableError;
```
palpatim, line 52: nit 

```suggestion
  const mysqlRetryableError = error.errno === '1045' && error.code === 'ER_ACCESS_DENIED_ERROR';

  return postgresRetryableError || mysqlRetryableError;
```

palpatim: Could you also add links to the sources of each of these error numbers & codes, so if we need to extend them in future we have a good starting point?

palpatim (dismissed):

packages/amplify-graphql-model-transformer/rds-lambda/handler.ts
```diff
@@ -24,10 +24,37 @@ export const run = async (event): Promise<any> => {
...
+    return await adapter.executeRequest(event, debugMode);
+  } catch (err) {
+    adapter = null;
+    throw err;
+  }
+};
+
+const isRetryableError = (error): boolean => {
+  // https://www.postgresql.org/docs/current/errcodes-appendix.html
+  const postgresRetryableError = error.code === '28P01' && error.routine === 'auth_failed';
```
atierian, line 50: From looking at the PostgreSQL docs, `routine` isn't part of the public interface for errors. It appears to be the name of the [function that reports the error](https://github.com/postgres/postgres/blob/master/src/backend/libpq/auth.c#L248). 

Are we comfortable relying on this to determine the error type? Is the `error.code` check maybe enough here?

dpilch: Good point. Looking at the docs I think using only `code` for Postgres and only `errno` for MySQL should be sufficient.

atierian (approved):
This handler doesn't know which DB system it's hitting, right?

Since a MySQL error has `code` property, I have a slight concern around future collisions -- basically if there ever was a MySQL error code `28P01`, we'd incorrectly interpret that as a retryable PostgreSQL error. Those appear to be disjoint sets today though and considering this is a single retry, nothing terrible would happen in the unlikely event that changes. We could mitigate by checking the existence of a known PostgreSQL property on the error, but I'm approving because the concern is probably unfounded.

dpilch:
The handler has the engine stored as an environment variable. So we could add that to the expression.

palpatim (approved):

PR #2248, opened 2024-02-05
BREAKING CHANGE: expose table representative & access refactor for amplify managed table in api construct

AaronZyLee:
#### Description of changes

- Export amplify table representatives as `ITable` in the generated resources `api.resources.tables`. As a result, the table can be referenced for other resources when it uses the amplify managed table strategy.
- Move amplify table L1 wrapper under the `cfnTables`. The new access pattern is `api.resources.cfnResources.amplifyDynamoDBTables`
- Have the `BREAKING CHANGE` commit as the access pattern is changed. Will follow up with the doc update PR

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes


packages/amplify-graphql-api-construct/src/internal/construct-exports.ts
```diff
@@ -33,7 +41,7 @@ export const getGeneratedResources = (scope: Construct): AmplifyGraphqlApiResour
   const cfnResolvers: Record<string, CfnResolver> = {};
   const cfnFunctionConfigurations: Record<string, CfnFunctionConfiguration> = {};
   const cfnDataSources: Record<string, CfnDataSource> = {};
-  const tables: Record<string, Table> = {};
+  const tables: Record<string, ITable> = {};
```
AaronZyLee, line 44: This does not change the final output type for tables, which is already defined as `ITable` [here ](https://github.com/aws-amplify/amplify-category-api/blob/cd23fb9a9ad982198c10ebfba07d285871e5d62a/packages/amplify-graphql-api-construct/src/types.ts#L747-L750)

palpatim:
Please confirm that this does not actually generate a reference to the table in the CDK stack -- the physical DDB table is still not a member of any stack.

packages/amplify-graphql-api-construct/src/internal/construct-exports.ts
```diff
@@ -17,6 +17,14 @@ import { AmplifyGraphqlApiResources, FunctionSlot } from '../types';
 import { AmplifyDynamoDbTableWrapper } from '../amplify-dynamodb-table-wrapper';
 import { walkAndProcessNodes } from './construct-tree';
 
+/**
+ * Check if a resource is implementing table interface
+ * @param table table resource
+ * @returns whether the resource is a ITable or not
+ */
+function isITable(table: any): table is ITable {
```
palpatim, line 25: Does the table in `tableFromAttr` below actually conform to ITable, or does it only contain the `tableArn`? If the former, this type predicate is inaccurate, since it is both missing required fields and requiring `tableArn` which is an optional field. ([Ref](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_dynamodb.ITable.html)) If the latter, we shouldn't be putting it in a structure which is declaring `ITable` since it doesn't actually conform to that interface.

AaronZyLee:
> Please confirm that this does not actually generate a reference to the table in the CDK stack -- the physical DDB table is still not a member of any stack.

Yes it does not exist in any CDK stacks

packages/amplify-graphql-api-construct/src/internal/construct-exports.ts
AaronZyLee: The `tableFromAttr` is defined as `ITable` [here](https://github.com/aws-amplify/amplify-category-api/blob/main/packages/amplify-graphql-model-transformer/src/resources/amplify-dynamodb-table/amplify-dynamodb-table-construct/index.ts#L37). It is created via the CDK Table construct call `Table.fromTableAttributes` ([source code](https://github.com/aws-amplify/amplify-category-api/blob/main/packages/amplify-graphql-model-transformer/src/resources/amplify-dynamodb-table/amplify-dynamodb-table-construct/index.ts#L106-L109)).

AaronZyLee: I've added all required field of `ITable` in the check

packages/amplify-graphql-api-construct/src/__tests__/amplify-dynamodb-table-wrapper.test.ts
```diff
@@ -292,5 +292,16 @@ describe('AmplifyDynamoDbTable', () => {
         });
       });
     });
+
+    describe('applyRemovalPolicy', () => {
+      it('overrides the removal policy', () => {
+        testResource.applyRemovalPolicy(RemovalPolicy.DESTROY);
```
palpatim, line 298: 1. Do we have tests to assert default values, or is that a relevant concept here?
2. Is there a removal policy getter on the wrapper? If so, does it store a separate property or proxy remoal policy of the underlying resource?

phani-srikar (approved):

phani-srikar:
I would rename the title to include `BREAKING CHANGE` so it's more apparent.

packages/amplify-graphql-api-construct/src/__tests__/amplify-dynamodb-table-wrapper.test.ts
AaronZyLee: 1. We don't have the test. The default values are set in the resource generator of model transformer which is `DESTROY` be default. I can add the default value check in my follow import table PR as its default value is `RETAIN`. This unit test is for the setter only.
2. There is no getter for the wrapper. Technically, the removal policy is not a resource property but a resource option as in the CFN template they are defined at the same level as `Properties`

palpatim:
> I would rename the title to include `BREAKING CHANGE` so it's more apparent.

nit: Let's use [conventional commit format](https://www.conventionalcommits.org/en/v1.0.0/#commit-message-with-both--and-breaking-change-footer):

In your squash & merge commit, the commit title should be something like:
`feat!: expose table representative for managed table in api construct`

And the commit footer/message should be:

BREAKING CHANGE: <description>

palpatim (approved):

PR #2261, opened 2024-02-07
chore(ci-cd): expand e2e testing regions to all Amplify supported regions

phani-srikar:
#### Description of changes

- Adds support for running E2Es in all [Amplify supported regions](https://docs.aws.amazon.com/general/latest/gr/amplify.html) including opt-in regions `me-south-1, eu-south-1, ap-east-1`.
- For the opt-in regions, we check if one or more of the child E2E accounts have them enabled and choose the runner accordingly.
- Add select DDB and CDK construct base case E2Es to be run in all regions.
- Add bootstrap step before we deploy CDK apps. This does not add any significant overhead with the advantage of preventing test failures due to unrelated CI setup scenarios.

[Full E2E run](https://tiny.amazon.com/1dd8ijyww/IsenLink) and [admin-role test run](https://tiny.amazon.com/17p8sddhu/IsenLink)
Only the `sql_models` CDK test is failing which will be fixed on `main`. This test fails for non-opt-in regions too.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


scripts/e2e-test-regions.json
```diff
@@ -0,0 +1,21 @@
+[
+  "ap-northeast-2",
```
sundersc, line 2: 'ap-northeast-3' is missing.

packages/amplify-e2e-tests/src/cleanup-e2e-resources.ts
```diff
@@ -292,7 +300,7 @@ const getS3Buckets = async (account: AWSAccountInfo): Promise<S3BucketInfo[]> =>
         });
       }
     } catch (e) {
-      if (e.code !== 'NoSuchTagSet' && e.code !== 'NoSuchBucket') {
+      if (e.code !== 'NoSuchTagSet' && e.code !== 'NoSuchBucket' && e.code !== 'IllegalLocationConstraintException') {
```
palpatim, line 303: Please see https://github.com/aws-amplify/amplify-category-api/pull/2269

The `IllegalLocationConstraintException` appears to be related to buckets created in opt-in regions. The proper fix for that appears to be to use a regionalized S3 client to get the bucket tagging. However, as noted in that PR, there are some buckets that were created in opt-in regions other than the one to which the account itself is opted-in! We don't quite understand how that happened (maybe it was previously opted into that region?) but those buckets can't be operated on.

palpatim (changes requested):
See comment on bucket validation

packages/amplify-graphql-api-construct-tests/src/commands.ts
```diff
@@ -84,6 +84,14 @@ export type CdkDeployProps = {
  * @returns the generated outputs file as a JSON object
  */
 export const cdkDeploy = async (cwd: string, option: string, props?: CdkDeployProps): Promise<any> => {
+  await spawn(getNpxPath(), ['cdk', 'bootstrap'], {
```
palpatim, line 87: We should not do this for all regions -- it's not clear to me what the behavior of a "bootstrap" is on a disabled optin region, but I have found instances of cdk deployment buckets for opt-in regions in accounts that do not have that region enabled.

scripts/e2e-test-regions.json
phani-srikar: That is not an amplify supported region.

scripts/split-e2e-tests.ts
```diff
@@ -118,14 +110,34 @@ const RUN_SOLO: (string | RegExp)[] = [
...
+  'src/__tests__/api_canary.test.ts',
+  // CDK tests
+  'src/__tests__/base-cdk.test.ts',
+];
+
+const RUN_IN_NON_OPT_IN_REGIONS: (string | RegExp)[] = [
+  // SQL tests
   /src\/__tests__\/rds-.*\.test\.ts/,
+  // Searchable tests
+  /src\/__tests__\/.*searchable.*\.test\.ts/,
```
phani-srikar, line 140: We have items in our backlog like [Manifest for the regionalized SNS topics](https://app.asana.com/0/1206048977824382/1206213458097061/f), [Add me-south-1 to beta stage](https://app.asana.com/0/1206048977824382/1206648775264573/f) and [Searchable support for eu-south-1](https://app.asana.com/0/1206048977824382/1206648775264575/f) that need to be addressed before these can be supported.

scripts/e2e-test-regions.json
sundersc: Looks, that's right. But I don't understand why CLI use this region with their E2E tests.
https://github.com/aws-amplify/amplify-cli/blob/d3261769506403c3a3222a5af2eb213b0f3b7844/packages/amplify-e2e-core/src/configure/index.ts#L26-L46

sundersc (approved):

scripts/e2e-test-regions.json
palpatim, line 2: nit: Please alphabetize this list so it's easier to scan for missing/new regions in future.

Slightly bigger nit: If we're already maintaining a list of supported regions somewhere else in this codebase, let's reuse that rather than maintaining another source of truth for the supported regions.

scripts/split-e2e-tests.ts
```diff
@@ -3,6 +3,11 @@ import * as glob from 'glob';
 import * as fs from 'fs-extra';
 import * as yaml from 'js-yaml';
 
+const REPO_ROOT = join(__dirname, '..');
+const SUPPORTED_REGIONS_PATH = join(REPO_ROOT, 'scripts', 'e2e-test-regions.json');
+const AMPLIFY_SUPPORTED_REGIONS: string[] = JSON.parse(fs.readFileSync(SUPPORTED_REGIONS_PATH, 'utf-8'));
+const AWS_REGIONS_TO_RUN_TESTS = AMPLIFY_SUPPORTED_REGIONS;
+const SUPPORTED_OPT_IN_REGIONS = ['ap-east-1', 'eu-south-1', 'me-south-1'];
```
palpatim, line 10: Let's include the opt in status in the region list, rather than re-typing a list of regions. Basically, we should reuse the structure of the [`Account.ListRegions` API](https://docs.aws.amazon.com/accounts/latest/reference/API_ListRegions.html) `Regions` response element.

shared-scripts.sh
```diff
@@ -276,10 +276,27 @@ function _unassumeTestAccountCredentials {
 function useChildAccountCredentials {
     if [ -z "$USE_PARENT_ACCOUNT" ]; then
         export AWS_PAGER=""
+        export AWS_MAX_ATTEMPTS=5
+        export AWS_STS_REGIONAL_ENDPOINTS=regional
         parent_acct=$(aws sts get-caller-identity | jq -cr '.Account')
         child_accts=$(aws organizations list-accounts | jq -c "[.Accounts[].Id | select(. != \"$parent_acct\")]")
         org_size=$(echo $child_accts | jq 'length')
-        pick_acct=$(echo $child_accts | jq -cr ".[$RANDOM % $org_size]")
+        opt_in_regions="eu-south-1 ap-east-1 me-south-1"
```
palpatim, line 284: After changing as noted above, pull this from the JSON file.

shared-scripts.sh
```diff
@@ -276,10 +276,27 @@ function _unassumeTestAccountCredentials {
...
+        if echo "$opt_in_regions" | grep -qw "$CLI_REGION"; then
+            child_accts=$(echo $child_accts | jq -c '.[]')
+            for child_acct in $child_accts; do
+                child_acct=$(echo $child_acct | tr -d '"')
```
palpatim, line 288: can you get rid of this by using the `-r` flag in the `jq` on L286?

packages/amplify-e2e-tests/tsconfig.json
```diff
@@ -10,7 +10,8 @@
     "outDir": "lib",
     "lib": ["es2015", "es2016.array.include", "esnext.asynciterable", "dom"],
     "declaration": true,
-    "typeRoots": ["../../node_modules/@types", "node_modules/@types", "./typings"]
+    "typeRoots": ["../../node_modules/@types", "node_modules/@types", "./typings"],
+    "resolveJsonModule": true
```
AaronZyLee, line 14: Is this config needed? I see it is required if you import directly from a JSON file [here](https://www.typescriptlang.org/tsconfig#resolveJsonModule). In the changes for cleanup script, I see they are parsed by a readFileSync and JSON.parse

scripts/split-e2e-tests.ts
```diff
@@ -154,10 +168,12 @@ const createJob = (os: OSType, jobIdx: number, runSolo = false): CandidateJob =>
...
 });
 
-const getTestNameFromPath = (testSuitePath: string): string => {
+const getTestNameFromPath = (testSuitePath: string, region?: string): string => {
   const startIndex = testSuitePath.lastIndexOf('/') + 1;
   const endIndex = testSuitePath.lastIndexOf('.test');
-  return testSuitePath.substring(startIndex, endIndex).split('.e2e').join('').split('.').join('-');
+  const regionSuffix =
+    RUN_IN_ALL_REGIONS.find((allRegions) => testSuitePath === allRegions || testSuitePath.match(allRegions)) && region ? region : '';
+  return testSuitePath.substring(startIndex, endIndex).split('.e2e').join('').split('.').join('-').concat(regionSuffix);
```
AaronZyLee, line 176: nit: The delimiter `-` can be added before the regionSuffix for a better reading

packages/amplify-e2e-tests/tsconfig.json
phani-srikar: Removed, not needed

scripts/e2e-test-regions.json
phani-srikar: Changed the format to include the region optionality info and sorted by region name. We previously had 2 lists in cleanup, split-e2e-tests scripts but with this change I have consolidated them into a single file.

AaronZyLee (approved):

PR #2263, opened 2024-02-09
test(api): add e2e for iam private field auth

sundersc:
Add E2E test cases for validating IAM provider field level auth for update and delete operations.


AaronZyLee (approved):

atierian (approved):
LGTM.

What do you think about adding (in a follow up) a matrix that exhaustively covers possible combinations or model level / field level auth?

PR #2268, opened 2024-02-13
test(api): add create api canary codebuild buildspec

sundersc:
Add codebuild spec to run a canary test in all supported regions. The region list is taken from the SQL layer pipeline.

### DIFF
--- a/codebuild_specs/createapi_canary_workflow.yml
+++ b/codebuild_specs/createapi_canary_workflow.yml
@@ -1,0 +1,267 @@
+version: 0.2
+env:
+  shell: bash
+  compute-type: BUILD_GENERAL1_MEDIUM
+batch:
+  fast-fail: false
+  build-graph:
+    - identifier: build_linux
+      buildspec: codebuild_specs/build_linux.yml
+      env:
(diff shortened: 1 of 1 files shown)


palpatim (approved):

codebuild_specs/createapi_canary_workflow.yml
```diff
@@ -0,0 +1,267 @@
...
+  shell: bash
+  compute-type: BUILD_GENERAL1_MEDIUM
+batch:
+  fast-fail: false
+  build-graph:
+    - identifier: build_linux
+      buildspec: codebuild_specs/build_linux.yml
+      env:
+        compute-type: BUILD_GENERAL1_LARGE
+    - identifier: build_windows
```
AaronZyLee, line 12: Nit: The test dependencies involved are only `build_linux` and `publish_to_local_registry`. The other tests such as `verify_yarn_lock`, `mock_e2e_tests`, can be removed if the test mainly focuses on the region e2e tests, since those are already included in the PR workflow

AaronZyLee (approved):

PR #2269, opened 2024-02-13
test: improve E2E test reliability

palpatim:
#### Description of changes

- Fixes deployment errors due to unsupported Node14 Lambda runtime in custom resource provider in CDK v2.80.0
- Increases "no output timeout" for a few tests to account for codegen asset BucketDeployment
- Suppresses errors getting bucket tags, so non-erroring buckets can be processed

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [ ] [E2E tests](https://us-east-1.console.aws.amazon.com/codesuite/codebuild/594813022831/projects/amplify-category-api-e2e-workflow/batch/amplify-category-api-e2e-workflow%3A977e7bf8-a103-4352-a84b-c46cb6df7ca4?region=us-east-1) pass

### DIFF
--- a/packages/amplify-e2e-core/src/init/initProjectHelper.ts
+++ b/packages/amplify-e2e-core/src/init/initProjectHelper.ts
@@ -49,7 +49,13 @@ export function initJSProjectWithProfile(cwd: string, settings?: Partial<typeof 
   if (s?.name?.length > 20) console.warn('Project names should not be longer than 20 characters. This may cause tests to break.');
 
   return new Promise((resolve, reject) => {
-    const chain = spawn(getCLIPath(), cliArgs, { cwd, stripColors: true, env, disableCIDetection: s.disableCIDetection })
+    const chain = spawn(getCLIPath(), cliArgs, {
+      cwd,
+      stripColors: true,
+      env,
+      disableCIDetection: s.disableCIDetection,
+      noOutputTimeout: 10 * 60 * 1000,
+    })
       .wait('Enter a name for the project')
       .sendLine(s.name)
       .wait('Initialize the project with the above configuration?')
--- a/packages/amplify-e2e-tests/src/cleanup-e2e-resources.ts
+++ b/packages/amplify-e2e-tests/src/cleanup-e2e-resources.ts
@@ -302,9 +302,9 @@ const getS3Buckets = async (account: AWSAccountInfo): Promise<S3BucketInfo[]> =>
   const buckets = await s3Client.listBuckets().promise();
   const result: S3BucketInfo[] = [];
   for (const bucket of buckets.Buckets) {
+    let region: string | undefined;
     try {
-      const region = await getBucketRegion(account, bucket.Name);
-
+      region = await getBucketRegion(account, bucket.Name);
       // Operations on buckets created in opt-in regions appear to require region-specific clients
       const regionalizedClient = new aws.S3({
         region,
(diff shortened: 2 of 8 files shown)


packages/amplify-e2e-tests/src/cleanup-e2e-resources.ts
```diff
@@ -277,27 +288,47 @@ const getJobCodeBuildDetails = async (jobIds: string[]): Promise<CodeBuild.Build
...
           name: bucket.Name,
           jobId,
+          region,
         });
       }
     } catch (e) {
       if (e.code !== 'NoSuchTagSet' && e.code !== 'NoSuchBucket') {
-        throw e;
+        console.error(`Skipping processing ${account.accountId}, bucket ${bucket.Name}`, e);
+        // throw e;
```
atierian, line 325: Do we still want to throw here or should this line be removed?

packages/amplify-graphql-api-construct-tests/src/commands.ts
```diff
@@ -52,7 +52,7 @@ export type InitCDKProjectProps = {
  * @returns a promise which resolves to the stack name
  */
 export const initCDKProject = async (cwd: string, templatePath: string, props?: InitCDKProjectProps): Promise<string> => {
-  const { cdkVersion = '2.80.0', additionalDependencies = [] } = props ?? {};
+  const { cdkVersion = '2.97.0', additionalDependencies = [] } = props ?? {};
```
atierian, line 55: I see you noted in the description that a bump is necessary due to the unsupported Node14 Lambda runtime. Why only bump to `2.97.0` when the current version is [2.128.0](https://github.com/aws/aws-cdk/releases/tag/v2.128.0)?

palpatim: Another excellent question. We have an ongoing conversation with the CLI team about what version we should actually be testing:
- The CDK version officially supported by Gen1 CLI is 2.80.0 (which fails as noted)
- The earliest CDK version we can safely use in our CDK constructs is 2.97.0
- The version officially supported by the Gen2 CLI was [2.110.0](https://github.com/aws-amplify/amplify-backend/blame/a01f6b9777f4fe7394b05de115f37e70c10d9af8/packages/backend/package.json) up until 2 days ago, when the team bumped it to [2.127.0](https://github.com/aws-amplify/amplify-backend/blob/main/packages/backend/package.json#L38)

For now, we'll maintain the earliest working version so we introduce the fewest changes. We'll synchronize the supported versions between the CDK construct, the Gen2 CLI, and the associated tests, which we're tracking in a backlog item elsewhere.

atierian: Thanks for the context.

packages/amplify-e2e-core/src/init/initProjectHelper.ts
```diff
@@ -49,7 +49,13 @@ export function initJSProjectWithProfile(cwd: string, settings?: Partial<typeof
   if (s?.name?.length > 20) console.warn('Project names should not be longer than 20 characters. This may cause tests to break.');
 
   return new Promise((resolve, reject) => {
-    const chain = spawn(getCLIPath(), cliArgs, { cwd, stripColors: true, env, disableCIDetection: s.disableCIDetection })
+    const chain = spawn(getCLIPath(), cliArgs, {
+      cwd,
+      stripColors: true,
+      env,
+      disableCIDetection: s.disableCIDetection,
+      noOutputTimeout: 10 * 60 * 1000,
```
sundersc, line 57: Does `init` take 10 minutes? Searchable tests may need longer timeouts too.

sundersc (approved):

packages/amplify-e2e-tests/src/cleanup-e2e-resources.ts
palpatim: Updated

packages/amplify-e2e-core/src/init/initProjectHelper.ts
palpatim: Let's keep an eye on that -- I'm not seeing failures in searchable tests, so it may be that I am barking up the wrong tree entirely.

AaronZyLee: Searchable tests usually takes longer time during `amplify push` which was resolved previously. The init process time should stay the same as the others

AaronZyLee (approved):

PR #2272, opened 2024-02-14
test(api): remove unwanted steps from canary build spec

sundersc:
Remove the below steps which are not required for canary.
- Windows build.
- Yarn lock verify.
- License validation.
- Verify CDK version.
- Verify API extract.

Modified the test to a simple usecase so that the canary is faster.

### DIFF
--- a/packages/amplify-e2e-tests/src/__tests__/api_canary.ts
+++ b/packages/amplify-e2e-tests/src/__tests__/api_canary.ts
@@ -1,0 +1,63 @@
+import path from 'path';
+import { existsSync } from 'fs';
+import {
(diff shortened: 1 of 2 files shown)


palpatim (approved):

packages/amplify-e2e-tests/src/__tests__/api_canary.ts
AaronZyLee, line 1: TODO: this test file might be excluded from the `split-e2e-tests` scripts if it is for the canary only. This can be addressed separately.

AaronZyLee (approved):

PR #2275, opened 2024-02-15
test: add basic secrets manager tests

dpilch:
#### Description of changes

Add E2E tests for using secrets manager as credential store using default and custom encryption keys.

This change is merging into a feature branch.

N/A

#### Issue #, if available

N/A

E2E tests: https://us-east-1.console.aws.amazon.com/codesuite/codebuild/594813022831/projects/amplify-category-api-e2e-workflow/batch/amplify-category-api-e2e-workflow:46ce09a7-d708-4ead-8ade-dd8e75e3fd7b?region=us-east-1#
#### Checklist

- [x] PR description included


packages/amplify-graphql-api-construct-tests/src/__tests__/sql-models.test.ts
```diff
@@ -199,7 +300,23 @@ const cleanupDatabase = async (options: { identifier: string; region: string; db
  * @param dbDetails the details object
  * @param projRoot the destination directory to write the `db-details.json` file to
  */
-const writeDbDetails = (dbDetails: DBDetails, projRoot: string): void => {
+const writeDbDetails = (
+  dbDetails: {
```
palpatim, line 304: Suggest we reuse the `DBDetails` type instead of manually specifying the shape

```ts
dbDetails: DBDetails & Omit<DBDetails, 'connectionConfigs'> & {dbConnectionConfig: SqlModelDataSourceDbConnectionConfig}
```

Alternately we could declare a base shape & extend it for individual DB connections vs. the aggregate connection.

packages/amplify-graphql-api-construct-tests/src/__tests__/sql-models.test.ts
```diff
@@ -83,10 +85,22 @@ describe('CDK GraphQL Transformer', () => {
     deleteProjectDir(projRoot);
   });
 
-  it('creates a GraphQL API from SQL-based models', async () => {
+  test('creates a GraphQL API from SQL-based models with Secrets Manager Credential Store default encryption key', async () => {
```
palpatim, line 88: I note that these tests won't exercise the "secrets created by RDS" flow, only the "secrets manually created" flow, correct? Given that's an officially supported use case I'd be happier if we had a test to exercise that flow. I think it would require a different `backend` though, since the one for the sql-models tests explicitly passes username/password to the RDS instance construct. If you agree, please add in a separate PR rather than increasing the line count on this one :)

AaronZyLee (approved):
LGTM. It might be a possible TODO task for the ops to clean the secrets generated from the failed tests.

palpatim (approved):

PR #2279, opened 2024-02-16
test: increase no output timeout for sql-models test

palpatim:
#### Description of changes

Between the SQL Layer custom resource, Codegen asset auto delete custom resource, Codegen asset bucket deployment, and Lambda layer provisioned concurrency, this test doesn't produce output frequently enough to keep nexpect happy. The test itself appears to be stable, but this change increases the timeout to account for slower deployment times in CICD.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [ ] [E2E tests pass](https://us-east-1.console.aws.amazon.com/codesuite/codebuild/594813022831/projects/amplify-category-api-e2e-workflow/batch/amplify-category-api-e2e-workflow:195bf1ce-9b91-47e5-bd53-d3ce4c98d73f?region=us-east-1)

### DIFF
--- a/packages/amplify-graphql-api-construct-tests/src/__tests__/sql-models.test.ts
+++ b/packages/amplify-graphql-api-construct-tests/src/__tests__/sql-models.test.ts
@@ -87,7 +87,11 @@ describe('CDK GraphQL Transformer', () => {
     const templatePath = path.resolve(path.join(__dirname, 'backends', 'sql-models'));
     const name = await initCDKProject(projRoot, templatePath);
     writeDbDetails(dbDetails, projRoot);
-    const outputs = await cdkDeploy(projRoot, '--all');
+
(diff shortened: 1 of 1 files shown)


palpatim:
Closing this PR -- although there is no output from the CDK deploy, this does not appear to be fundamentally a prompt timeout. This test passes reliably locally, so it's either a race condition (and thus sensitive to environmental problems) or a configuration problem that manifests in this account/region combination. Taking a different troubleshooting tack.

palpatim:
Error during deployment of SQL layer custom resource. This is just a call to S3, so we'll want to make sure that layer manifest is present:

```
CloudFormation did not receive a response from your Custom Resource. Please check your logs for requestId [7b1e969a-e919-4727-97cf-fb6144c2a26e]. If you are using the Python cfn-response module, you may need to update your Lambda function code so that CloudFormation can attach the updated version.
```

palpatim:
This does not occur with CDK v 2.127.0

PR #2283, opened 2024-02-20
chore(ci): emit canary metrics on each region test

sundersc:
Currently the metrics for failed canaries are not emitted to cloudwatch. We expect codebuild to send '0' for failed canary runs.

This is fundamentally due to how codebuild works, codebuild sets the environment variable `CODEBUILD_BUILD_SUCCEEDING` to 1 if the last step of the build step is succeeding (based on error code). Since we have emit metrics command in 'build' step, this emits `1` when all the tests are passing but skipped if any of the prior steps fail.

In this PR, we are changing this behavior. We will emit metrics for each test. The alarm would consolidate all these values and work as expected because we consider only the AVERAGE to trigger it.

### DIFF
--- a/codebuild_specs/run_canary_e2e_tests.yml
+++ b/codebuild_specs/run_canary_e2e_tests.yml
@@ -10,6 +10,16 @@ env:
 phases:
   build:
     commands:
+      - codebuild-breakpoint
+      - source ./shared-scripts.sh && _setupE2ETestsLinux
+      - source ./shared-scripts.sh && _runE2ETestsLinux
+  post_build:
+    commands:
       - source ./shared-scripts.sh && _unassumeTestAccountCredentials
       - aws sts get-caller-identity
-      - source ./shared-scripts.sh && _emitCreateApiCanaryMetric
+      - source ./shared-scripts.sh && _scanArtifacts && _emitCreateApiCanaryMetric
+
+artifacts:
+  files:
+    - '**/*'
+  base-directory: $CODEBUILD_SRC_DIR/packages/amplify-e2e-tests/amplify-e2e-reports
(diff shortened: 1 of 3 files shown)


codebuild_specs/run_canary_e2e_tests.yml
```diff
@@ -9,7 +9,17 @@ env:
...
+    commands:
+      - codebuild-breakpoint
+      - source ./shared-scripts.sh && _setupE2ETestsLinux
+      - source ./shared-scripts.sh && _runE2ETestsLinux
+  post_build:
     commands:
       - source ./shared-scripts.sh && _unassumeTestAccountCredentials
       - aws sts get-caller-identity
-      - source ./shared-scripts.sh && _emitCreateApiCanaryMetric
+      - source ./shared-scripts.sh && _scanArtifacts && _emitCreateApiCanaryMetric
```
palpatim, line 20: Can you confirm that `_emitCreateApiCanaryMetric` includes the region as a metric dimension?

palpatim (approved):
Approved contingent upon emitting region in the metric

codebuild_specs/run_canary_e2e_tests.yml
sundersc: I've added region to metric dimension.

palpatim (approved):

shared-scripts.sh
```diff
@@ -445,6 +445,6 @@ function _emitCreateApiCanaryMetric {
     --namespace amplify-category-api-e2e-tests \
     --unit Count \
     --value $CODEBUILD_BUILD_SUCCEEDING \
-    --dimensions branch=main \
+    --dimensions branch=main,region=$CLI_REGION \
```
phani-srikar, line 448: Is this canary intended to run on `main` or `release`?

phani-srikar (approved):

PR #2298, opened 2024-02-23
feat: replace conflictResolution with dataStoreConfiguration

dpilch:
#### Description of changes

* replace `conflictResolution` with `dataStoreConfiguration`
* mark `conflictResolution` as deprecated.
  * `conflictResolution` can still be used for backwards compatibility.
* throw error if `conflictResolution` and `dataStoreConfiguration` are used at the same time

See description of changes

#### Issue #, if available

N/A

* Unit tests
* E2E tests: https://us-east-1.console.aws.amazon.com/codebuild/home?region=us-east-1#/builds/build-batch/amplify-category-api-e2e-workflow:d1c2489e-f8bb-4b14-b004-75551ac2602f/view/new

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-graphql-api-construct/API.md
```diff
@@ -125,6 +125,7 @@ export interface AmplifyGraphqlApiProps {
     readonly apiName?: string;
     readonly authorizationModes: AuthorizationModes;
     readonly conflictResolution?: ConflictResolution;
```
palpatim, line 127: Shouldn't this be deprecated now? Maybe API.md doesn't support the deprecation flag though

packages/amplify-graphql-api-construct/src/amplify-graphql-api.ts
```diff
@@ -152,8 +152,17 @@ export class AmplifyGraphqlApi extends Construct {
       translationBehavior,
       functionNameMap,
       outputStorageStrategy,
+      dataStoreConfiguration,
     } = props;
 
+    if (conflictResolution && dataStoreConfiguration) {
+      throw new Error(
+        `conflictResolution is deprecrated. conflictResolution and dataStoreConfiguration cannot be used together. Please use dataStoreConfiguration.`,
```
palpatim, line 160:
```suggestion
        `conflictResolution is deprecated. conflictResolution and dataStoreConfiguration cannot be used together. Please use dataStoreConfiguration.`,
```

packages/amplify-graphql-api-construct/src/amplify-graphql-api.ts
```diff
@@ -217,7 +226,7 @@ export class AmplifyGraphqlApi extends Construct {
     this.codegenAssets = new CodegenAssets(this, 'AmplifyCodegenAssets', { modelSchema: definition.schema });
 
     this.resources = getGeneratedResources(this);
-    this.conflictResolution = conflictResolution;
+    this.dataStoreConfiguration = dataStoreConflictResolution;
```
palpatim, line 229: Now that we're doing some magic with this, I'd prefer to move this higher up in the constructor--like around L164. You could then bypass the whole `dataStoreConflictResolution` intermediate variable entirely and just operate on `this.dataStoreConfiguration` throughout the rest of the ctor.

packages/amplify-graphql-api-construct/.jsii
```diff
@@ -4879,19 +4879,38 @@
...
           "locationInModule": {
             "filename": "src/types.ts",
-            "line": 640
+            "line": 647
           },
           "name": "conflictResolution",
           "optional": true,
           "type": {
-            "fqn": "@aws-amplify/graphql-api-construct.ConflictResolution"
+            "fqn": "@aws-amplify/graphql-api-construct.DataStoreConfiguration"
```
palpatim, line 4895: This worries me a little bit because JSII now declares that the `conflictResolution` field has a new type, as opposed to declaring the existing type or describing the existing shape.. I don't think it'll be a problem for TS, but when we start supporting more strictly typed languages we might need to keep an eye on this kind of change. 

In the meantime, can you confirm that you can build a standalone project using the packaged Data construct, using both `dataStoreConfiguration` and `conflictResolution` fields? I'd like this to be totally separate from the E2E tests, so we're not relying on any inside project configs. I don't think it's necessary to commit it, but I do want to validate that things work for existing customers without anything more dramatic than a deprecation message.

packages/amplify-graphql-api-construct/API.md
dpilch: It looks like this is because I included a comma after the `@deprecated`. There are a few other instances of this for unrelated properties. I'll create a separate pr to fix this.

dpilch: I've fixed two `@deprecated` tags added in this PR.

I looked into fixing all other issues using `jsdoc/check-tag-names` eslint rule. There are many more occurrences than I thought. I've created a backlog item to fix this later.

packages/amplify-graphql-api-construct/.jsii
dpilch: Looks like it could be an issue: https://aws.github.io/jsii/user-guides/lib-author/typescript-restrictions/#type-aliases

I've thought of another option that would give equivalent types.

```typescript
export interface ConflictResolution extends DataStoreConfiguration {}
```

I'll test it out an report back.

packages/amplify-graphql-api-construct/src/__tests__/__functional__/props.test.ts
```diff
@@ -69,52 +74,216 @@ describe('supports different props configurations', () => {
...
+      });
+
+      expect(executeTransformSpy.mock.calls[0][0].resolverConfig).toEqual(executeTransformSpy.mock.calls[1][0].resolverConfig);
+      expect(dataStoreConfigurationTemplate.toJSON().Metadata).toEqual(conflictResolutionTemplate.toJSON().Metadata);
+    });
   });
 
-  it('supports conflict resolution with lambda handling', () => {
+  it('supports a definition with a SQLModelDataSourceBinding with a VPC configuration', () => {
     verifySynth((stack) => {
```
AaronZyLee, line 286: Are line 286~303 related to the test description  in line 285?

codebuild_specs/e2e_workflow.yml
AaronZyLee, line 1: Are these changes expected? It seems no e2e tests being added/removed in the PR

packages/amplify-graphql-api-construct/src/__tests__/__functional__/props.test.ts
```diff
@@ -69,52 +74,216 @@ describe('supports different props configurations', () => {
...
+            project: {
+              handlerType: 'AUTOMERGE',
+              detectionType: 'VERSION',
+            },
+          },
+        });
+      });
+
+      expect(executeTransformSpy.mock.calls[0][0].resolverConfig).toEqual(executeTransformSpy.mock.calls[1][0].resolverConfig);
+      expect(dataStoreConfigurationTemplate.toJSON().Metadata).toEqual(conflictResolutionTemplate.toJSON().Metadata);
```
AaronZyLee, line 281: Is the datastore related config being compared in the `Metadata` section of CFN tempalte?

packages/amplify-graphql-api-construct/.jsii
dpilch: Works fine for all cases.

```typescript
export class BackendStack extends cdk.Stack {
  constructor(scope: Construct, id: string, props?: cdk.StackProps) {
    super(scope, id, props);

    // also works if type is DataStoreConfiguration
    const conflictResolution: ConflictResolution = 
      {
        models: {
          Blog: {
            handlerType: 'AUTOMERGE',
            detectionType: 'VERSION',
          }
        }
      };

    const amplifyApi = new AmplifyGraphqlApi(this, "MyNewApi", {
      definition: AmplifyGraphqlDefinition.fromFiles(path.join(__dirname, "schema.graphql")),
      authorizationModes: {
        defaultAuthorizationMode: 'API_KEY',
        apiKeyConfig: {
          expires: cdk.Duration.days(30)
        }
      },
      
      // works with either ConflictResolution or DataStoreConfiguration type
      // works when dataStoreConfiguration keys is used with either type
      conflictResolution: conflictResolution,
    })
  }
}

```

packages/amplify-graphql-api-construct/src/__tests__/__functional__/props.test.ts
dpilch: Oops, accidentally combined two tests.

packages/amplify-graphql-api-construct/src/__tests__/__functional__/props.test.ts
dpilch: This was supposed to assert that using `dataStoreConfiguration` and `conflictResolution` would result in the same template. But I forgot to update one of the props.

palpatim (approved):

AaronZyLee (approved):

PR #2320, opened 2024-03-04
feat: deprecate iam auth provider and rename it to identityPool

sobolk:
#### Description of changes

This pull request renames `provider: iam` to `provider: identityPool`. In such a way that:
1. Old alias - `iam` - is still supported without any changes in handling logic
2. New alias - `identityPool` - is now supported and follows the same logic old alias was using
3. There is a warning emitted if schema contains `iam` provider asking to migrate to `identityPool`

Out of scope:
1. V1 auth transformer
2. V1 -> V2 migrators - since we still support `iam` then V1 -> V2 with iam -> V2 with identityPool works.

Changes:
1. Duplicated test coverage for `iam` to cover new `identityPool` and assure they both work the same way.
2. Changes to logic to treat `iam` and `identityPool` provider the same way
3. Added deprecation warning.

E2E tests: https://tiny.amazon.com/1ds3s74vj/IsenLink

`searchable_datastore` is failing, but it's known problem not related to this PR.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-graphql-auth-transformer/src/__tests__/resolvers/__snapshots__/subscriptions.test.ts.snap
```diff
@@ -60,6 +60,52 @@ $util.toJson({\\"version\\":\\"2018-05-29\\",\\"payload\\":{}})
 ## [End] Authorization Steps. **"
 `;
 
+exports[`subscriptions generateAuthExpressionForSubscriptions identityPool renders for identityPool auth with no admin roles 1`] = `
+"## [Start] Authorization Steps. **
+$util.qr($ctx.stash.put(\\"hasAuth\\", true))
+#set( $isAuthorized = false )
+#if( $util.authType() == \\"IAM Authorization\\" )
+  #if( !$isAuthorized )
+    #if( $ctx.identity.userArn == $ctx.stash.undefined )
```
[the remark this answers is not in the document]
sobolk: @phani-srikar 
There's 10 hits on `main` for the same file , see https://github.com/search?q=repo%3Aaws-amplify%2Famplify-category-api%20%24ctx.stash.undefined&type=code .

Perhaps it's test/mocking problem.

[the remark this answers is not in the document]
sobolk: It was testing problem, fixed in https://github.com/aws-amplify/amplify-category-api/pull/2320/commits/902280b5065e151996919b9faca7c642572d87f7 .

phani-srikar (approved):

sundersc:
> * Old alias - `iam` - is still supported without any changes in handling logic

- If IAM auth is going to rely on policies, what are the differences between allow:public and allow:private?
- If we are deprecating IAM auth rule, do we expect IAM policy based auth to work even if there are no auth rules explicitly mentioned on the model?

packages/amplify-graphql-auth-transformer/src/__tests__/searchable-auth.test.ts
```diff
@@ -110,6 +110,57 @@ test('auth logic is enabled for iam/apiKey auth rules', () => {
...
+            id: ID!
+            content: String
+            secret: String @auth(rules: [{ allow: private, provider: identityPool }]) # only auth role can do crud on this
+        }
+    `;
+  const authConfig: AppSyncAuthConfiguration = {
+    defaultAuthentication: {
+      authenticationType: 'AMAZON_COGNITO_USER_POOLS',
+    },
+    additionalAuthenticationProviders: [
```
sundersc, line 131: What if the model contains only API KEY rule, do we still expect IAM policy based auth to work with those models?

sobolk:
> > * Old alias - `iam` - is still supported without any changes in handling logic
> 
> * If IAM auth is going to rely on policies, what are the differences between allow:public and allow:private?
> * If we are deprecating IAM auth rule, do we expect IAM policy based auth to work even if there are no auth rules explicitly mentioned on the model?

1. This PR does not introduce "new extra IAM behavior"
2. When new IAM behavior is introduced the IAM auth with Cognito Identity Pool (auth/unauth roles) is going to work the same (for backwards compatibility), regardless of new IAM opt-in flag.

This diagram should give an idea.
image

packages/amplify-graphql-auth-transformer/src/__tests__/searchable-auth.test.ts
sobolk: Only if new IAM behavior is enabled. (It's going to be boolean flag at data construct level).

sobolk: See also https://app.asana.com/0/1206048977824382/1206486113291633/f .

sundersc: In that case, please add a test with API_KEY as default authentication and IAM as additional authentication provider. In such cases, we need to add AppSync aws_iam directive to all models (may be on all fields too - IDR exact implementation).

sobolk: @sundersc The tests in this file seem to already assert `aws_iam` presence. So I just added extra `authConfig` permutations to iam/identityPool tests here https://github.com/aws-amplify/amplify-category-api/pull/2320/commits/b8fe34179b020852701f36872148141148445924 .

phani-srikar (approved):

packages/amplify-graphql-auth-transformer/src/__tests__/searchable-auth.test.ts
sundersc: > Only if new IAM behavior is enabled.
Is that in scope of this PR?

sobolk: > Only if new IAM behavior is enabled.

No this will be separate change. This PR is focused only on aliasing existing functionality.

sundersc (approved):
Changes LGTM. Let's not merge to main until we confirm with PM whether we want to add the new auth rule provider or not.

packages/graphql-transformers-e2e-tests/src/__tests__/AuthV2ExhaustiveT2A.test.ts
```diff
@@ -10,8 +10,8 @@ import {
 
 const strategyProviders: Record<AuthStrategy, AuthProvider[]> = {
   owner: [],
-  public: ['apiKey', 'iam'],
```
palpatim, line 13: Aren't tests in the "graphql-transformers-e2e-tests" package V1 tests? Should we be asserting `identityPool` support for V1 transformer?

sobolk: > Aren't tests in the "graphql-transformers-e2e-tests" package V1 tests?

They're testing both. See https://github.com/aws-amplify/amplify-category-api/blob/main/packages/graphql-transformers-e2e-tests/package.json#L36-L58 . The tests I modified say `AuthV2` - presumably v2 auth transformers.

> Should we be asserting identityPool support for V1 transformer?

No. This is out of scope (see PR description and comments here https://app.asana.com/0/0/1206486113291629/1206699612098204/f )

palpatim (approved):

PR #2330, opened 2024-03-07
refactor(graphql-relational-transformer): split out datasource specific directive logic

atierian:
#### Description of changes
Refactor of `amplify-graphql-relational-transformer` splitting out datasource specific directive logic. The goal of refactor is to allow additional logical branches for data sources & directive parameter combinations (e.g. DynamoDB references) to be added in a maintainable way.

There are no functional changes included in this PR.

**Changes:**
- Adds `DataSourceBasedDirectiveTransformer` interface to represent the methods called in the existing relational transformers that operate on individual `<RelationType>DirectiveConfiguration` types. [[11efed6](https://github.com/aws-amplify/amplify-category-api/pull/2330/commits/11efed6a5f54c87153c773046ea0f043e635a34f)]
- Adds `DataSourceBasedDirectiveTransformer` implementing types for:
  - `belongsTo` [[2d4b7fe](https://github.com/aws-amplify/amplify-category-api/pull/2330/commits/2d4b7fe90365cdb1e9679a61852f81ebb86fd6cf)]
  - `hasMany`: [[b87f784](https://github.com/aws-amplify/amplify-category-api/pull/2330/commits/b87f78414f05d27982a7226aa5bb509229108bd6)]
  - `hasOne`: [[4ea48e6](https://github.com/aws-amplify/amplify-category-api/pull/2330/commits/4ea48e6f809ff993bab5c5300efd97f31843390e)]
as well as applicable factory methods, for the currently supported data sources: SQL w/ references and DynamoDB w/ fields. `manyToMany` is explicitly not included in this change because it's currently only supported for a single data source (DynamoDB w/ fields parameter).
- Moves all data source specific directive logic out of the relational transformer methods into the `DataSourceBasedDirectiveTransformer` implementing types.

Note: I'd like to move the existing relational directive transformer files into the directive based directories created in this PR. But let's do that in a follow up to keep the scope of this diff as focused as possible.

N/A

#### Issue #, if available
N/A

E2E Tests: https://us-east-1.console.aws.amazon.com/codesuite/codebuild/594813022831/projects/amplify-category-api-e2e-workflow/batch/amplify-category-api-e2e-workflow%3A683530a8-acba-4a46-9d8b-0b809c5ca988?region=us-east-1

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-graphql-relational-transformer/src/belongs-to/belongs-to-directive-ddb-fields-transformer.ts
+++ b/packages/amplify-graphql-relational-transformer/src/belongs-to/belongs-to-directive-ddb-fields-transformer.ts
@@ -7,6 +7,12 @@ import { BelongsToDirectiveConfiguration } from '../types';
 import { getRelatedTypeIndex, ensureFieldsArray, getFieldsNodes, registerHasOneForeignKeyMappings } from '../utils';
 import { DataSourceBasedDirectiveTransformer } from '../data-source-based-directive-transformer';
 
+/**
+ * BelongsToDirectiveDDBFieldsTransformer executes transformations based on `@belongsTo(fields: [String!])` configurations
+ * and surrounding TransformerContextProviders for DynamoDB data sources.
+ *
+ * This should not be used for `@belongsTo(references: [String!])` definitions.
+ */
 export class BelongsToDirectiveDDBFieldsTransformer implements DataSourceBasedDirectiveTransformer<BelongsToDirectiveConfiguration> {
   dbType: 'DYNAMODB';
   constructor(dbType: 'DYNAMODB') {
@@ -31,7 +37,8 @@ export class BelongsToDirectiveDDBFieldsTransformer implements DataSourceBasedDi
     config.relatedTypeIndex = getRelatedTypeIndex(config, context as TransformerContextProvider);
   };
 
-  generateResolvers = (context: TransformerContextProvider, config: BelongsToDirectiveConfiguration): void => {
+  /** no-op */
+  generateResolvers = (_context: TransformerContextProvider, _config: BelongsToDirectiveConfiguration): void => {
     return;
   };
 
--- a/packages/amplify-graphql-relational-transformer/src/belongs-to/belongs-to-directive-ddb-fields-transformer.ts
+++ b/packages/amplify-graphql-relational-transformer/src/belongs-to/belongs-to-directive-ddb-fields-transformer.ts
@@ -1,0 +1,42 @@
+import {
+  TransformerContextProvider,
+  TransformerPrepareStepContextProvider,
+  TransformerTransformSchemaStepContextProvider,
+} from '@aws-amplify/graphql-transformer-interfaces';
+import { BelongsToDirectiveConfiguration } from '../types';
+import { getRelatedTypeIndex, ensureFieldsArray, getFieldsNodes, registerHasOneForeignKeyMappings } from '../utils';
+import { DataSourceBasedDirectiveTransformer } from '../data-source-based-directive-transformer';
+
+export class BelongsToDirectiveDDBFieldsTransformer implements DataSourceBasedDirectiveTransformer<BelongsToDirectiveConfiguration> {
+  dbType: 'DYNAMODB';
+  constructor(dbType: 'DYNAMODB') {
+    this.dbType = dbType;
+  }
+
+  prepare = (context: TransformerPrepareStepContextProvider, config: BelongsToDirectiveConfiguration): void => {
+    if (config.relationType !== 'hasOne') {
+      return;
+    }
+    const modelName = config.object.name.value;
+    registerHasOneForeignKeyMappings({
+      transformParameters: context.transformParameters,
+      resourceHelper: context.resourceHelper,
+      thisTypeName: modelName,
+      thisFieldName: config.field.name.value,
+      relatedType: config.relatedType,
+    });
+  };
+
+  transformSchema = (context: TransformerTransformSchemaStepContextProvider, config: BelongsToDirectiveConfiguration): void => {
+    config.relatedTypeIndex = getRelatedTypeIndex(config, context as TransformerContextProvider);
+  };
+
+  generateResolvers = (context: TransformerContextProvider, config: BelongsToDirectiveConfiguration): void => {
+    return;
+  };
+
+  validate = (context: TransformerContextProvider, config: BelongsToDirectiveConfiguration): void => {
+    ensureFieldsArray(config);
+    config.fieldNodes = getFieldsNodes(config, context);
+  };
+}
--- a/packages/amplify-graphql-relational-transformer/src/belongs-to/belongs-to-directive-sql-transformer.ts
+++ b/packages/amplify-graphql-relational-transformer/src/belongs-to/belongs-to-directive-sql-transformer.ts
@@ -9,6 +9,10 @@ import { BelongsToDirectiveConfiguration } from '../types';
 import { ensureReferencesArray, validateChildReferencesFields, getBelongsToReferencesNodes } from '../utils';
 import { DataSourceBasedDirectiveTransformer } from '../data-source-based-directive-transformer';
 
+/**
+ * BelongsToDirectiveSQLTransformer executes transformations based on `@belongsTo(references: [String!])` configurations
+ * and surrounding TransformerContextProviders for SQL data sources.
+ */
 export class BelongsToDirectiveSQLTransformer implements DataSourceBasedDirectiveTransformer<BelongsToDirectiveConfiguration> {
   dbType: ModelDataSourceStrategySqlDbType;
   constructor(dbType: ModelDataSourceStrategySqlDbType) {
@@ -24,7 +28,8 @@ export class BelongsToDirectiveSQLTransformer implements DataSourceBasedDirectiv
     validateChildReferencesFields(config, context as TransformerContextProvider);
   };
 
-  generateResolvers = (context: TransformerContextProvider, config: BelongsToDirectiveConfiguration): void => {
+  /** no-op */
+  generateResolvers = (_context: TransformerContextProvider, _config: BelongsToDirectiveConfiguration): void => {
     return;
   };
 
--- a/packages/amplify-graphql-relational-transformer/src/belongs-to/belongs-to-directive-sql-transformer.ts
+++ b/packages/amplify-graphql-relational-transformer/src/belongs-to/belongs-to-directive-sql-transformer.ts
@@ -1,0 +1,34 @@
+import {
+  TransformerContextProvider,
+  TransformerPrepareStepContextProvider,
+  TransformerTransformSchemaStepContextProvider,
+} from '@aws-amplify/graphql-transformer-interfaces';
+import { setFieldMappingResolverReference } from '../resolvers';
+import { BelongsToDirectiveConfiguration } from '../types';
+import { ensureReferencesArray, validateChildReferencesFields, getBelongsToReferencesNodes } from '../utils';
+import { DataSourceBasedDirectiveTransformer } from '../data-source-based-directive-transformer';
+
+export class BelongsToDirectiveSQLTransformer implements DataSourceBasedDirectiveTransformer<BelongsToDirectiveConfiguration> {
+  dbType: 'MYSQL' | 'POSTGRES';
+  constructor(dbType: 'MYSQL' | 'POSTGRES') {
+    this.dbType = dbType;
+  }
+
+  prepare = (context: TransformerPrepareStepContextProvider, config: BelongsToDirectiveConfiguration): void => {
+    const modelName = config.object.name.value;
+    setFieldMappingResolverReference(context, config.relatedType?.name?.value, modelName, config.field.name.value);
+  };
+
+  transformSchema = (context: TransformerTransformSchemaStepContextProvider, config: BelongsToDirectiveConfiguration): void => {
+    validateChildReferencesFields(config, context as TransformerContextProvider);
+  };
+
+  generateResolvers = (context: TransformerContextProvider, config: BelongsToDirectiveConfiguration): void => {
+    return;
+  };
+
+  validate = (context: TransformerContextProvider, config: BelongsToDirectiveConfiguration): void => {
+    ensureReferencesArray(config);
+    getBelongsToReferencesNodes(config, context);
+  };
+}
--- a/packages/amplify-graphql-relational-transformer/src/belongs-to/belongs-to-directive-sql-transformer.ts
+++ b/packages/amplify-graphql-relational-transformer/src/belongs-to/belongs-to-directive-sql-transformer.ts
@@ -1,4 +1,5 @@
 import {
+  ModelDataSourceStrategySqlDbType,
   TransformerContextProvider,
   TransformerPrepareStepContextProvider,
   TransformerTransformSchemaStepContextProvider,
@@ -9,8 +10,8 @@ import { ensureReferencesArray, validateChildReferencesFields, getBelongsToRefer
 import { DataSourceBasedDirectiveTransformer } from '../data-source-based-directive-transformer';
 
 export class BelongsToDirectiveSQLTransformer implements DataSourceBasedDirectiveTransformer<BelongsToDirectiveConfiguration> {
-  dbType: 'MYSQL' | 'POSTGRES';
-  constructor(dbType: 'MYSQL' | 'POSTGRES') {
+  dbType: ModelDataSourceStrategySqlDbType;
+  constructor(dbType: ModelDataSourceStrategySqlDbType) {
     this.dbType = dbType;
   }
 
--- a/packages/amplify-graphql-relational-transformer/src/belongs-to/belongs-to-directive-transformer-factory.ts
+++ b/packages/amplify-graphql-relational-transformer/src/belongs-to/belongs-to-directive-transformer-factory.ts
@@ -1,0 +1,17 @@
+import { ModelDataSourceStrategyDbType } from '@aws-amplify/graphql-transformer-interfaces';
+import { BelongsToDirectiveConfiguration } from '../types';
+import { DataSourceBasedDirectiveTransformer } from '../data-source-based-directive-transformer';
+import { BelongsToDirectiveSQLTransformer } from './belongs-to-directive-sql-transformer';
+import { BelongsToDirectiveDDBFieldsTransformer } from './belongs-to-directive-ddb-fields-transformer';
+
+export const getBelongsToDirectiveTransformer = (
+  dbType: ModelDataSourceStrategyDbType,
+): DataSourceBasedDirectiveTransformer<BelongsToDirectiveConfiguration> => {
+  switch (dbType) {
+    case 'MYSQL':
+    case 'POSTGRES':
+      return new BelongsToDirectiveSQLTransformer(dbType);
+    case 'DYNAMODB':
+      return new BelongsToDirectiveDDBFieldsTransformer(dbType);
+  }
+};
--- a/packages/amplify-graphql-relational-transformer/src/data-source-based-directive-transformer.ts
+++ b/packages/amplify-graphql-relational-transformer/src/data-source-based-directive-transformer.ts
@@ -1,0 +1,21 @@
+import {
+  TransformerContextProvider,
+  TransformerPrepareStepContextProvider,
+  ModelDataSourceStrategyDbType,
+  TransformerTransformSchemaStepContextProvider,
+} from '@aws-amplify/graphql-transformer-interfaces';
+import { BelongsToDirectiveConfiguration, HasManyDirectiveConfiguration, HasOneDirectiveConfiguration } from './types';
+
+export type DirectiveConfiguration = HasOneDirectiveConfiguration | HasManyDirectiveConfiguration | BelongsToDirectiveConfiguration;
+
+/**
+ * Represents a subset of transformer methods based on a specific
+ * data source (currently SQL / DDB Fields).
+*/
+export interface DataSourceBasedDirectiveTransformer<Config extends DirectiveConfiguration> {
+  dbType: ModelDataSourceStrategyDbType;
+  prepare: (context: TransformerPrepareStepContextProvider, config: Config) => void;
+  transformSchema: (context: TransformerTransformSchemaStepContextProvider, config: Config) => void;
+  generateResolvers: (context: TransformerContextProvider, config: Config) => void;
+  validate: (context: TransformerContextProvider, config: Config) => void;
+}
--- a/packages/amplify-graphql-relational-transformer/src/data-source-based-directive-transformer.ts
+++ b/packages/amplify-graphql-relational-transformer/src/data-source-based-directive-transformer.ts
@@ -11,8 +11,12 @@ export type DirectiveConfiguration = HasOneDirectiveConfiguration | HasManyDirec
 /**
  * Represents a subset of transformer methods based on a specific
  * data source (currently SQL / DDB Fields).
+ *
+ * Each method is to be invoked by the applicable relational transformer when iterating through
+ * its directive configurations the applicable transformer step.
  */
 export interface DataSourceBasedDirectiveTransformer<Config extends DirectiveConfiguration> {
+  // Constrains the DataSourceBasedDirectiveTransformer to a specific database type
   dbType: ModelDataSourceStrategyDbType;
   prepare: (context: TransformerPrepareStepContextProvider, config: Config) => void;
   transformSchema: (context: TransformerTransformSchemaStepContextProvider, config: Config) => void;
--- a/packages/amplify-graphql-relational-transformer/src/data-source-based-directive-transformer.ts
+++ b/packages/amplify-graphql-relational-transformer/src/data-source-based-directive-transformer.ts
@@ -11,7 +11,7 @@ export type DirectiveConfiguration = HasOneDirectiveConfiguration | HasManyDirec
 /**
  * Represents a subset of transformer methods based on a specific
  * data source (currently SQL / DDB Fields).
-*/
+ */
 export interface DataSourceBasedDirectiveTransformer<Config extends DirectiveConfiguration> {
   dbType: ModelDataSourceStrategyDbType;
   prepare: (context: TransformerPrepareStepContextProvider, config: Config) => void;
--- a/packages/amplify-graphql-relational-transformer/src/data-source-based-directive-transformer.ts
+++ b/packages/amplify-graphql-relational-transformer/src/data-source-based-directive-transformer.ts
@@ -6,7 +6,10 @@ import {
 } from '@aws-amplify/graphql-transformer-interfaces';
 import { BelongsToDirectiveConfiguration, HasManyDirectiveConfiguration, HasOneDirectiveConfiguration } from './types';
 
-export type DirectiveConfiguration = HasOneDirectiveConfiguration | HasManyDirectiveConfiguration | BelongsToDirectiveConfiguration;
+export type RelationalDirectiveConfiguration =
+  | HasOneDirectiveConfiguration
+  | HasManyDirectiveConfiguration
+  | BelongsToDirectiveConfiguration;
 
 /**
  * Represents a subset of transformer methods based on a specific
@@ -15,7 +18,7 @@ export type DirectiveConfiguration = HasOneDirectiveConfiguration | HasManyDirec
  * Each method is to be invoked by the applicable relational transformer when iterating through
  * its directive configurations the applicable transformer step.
  */
-export interface DataSourceBasedDirectiveTransformer<Config extends DirectiveConfiguration> {
+export interface DataSourceBasedDirectiveTransformer<Config extends RelationalDirectiveConfiguration> {
   // Constrains the DataSourceBasedDirectiveTransformer to a specific database type
   dbType: ModelDataSourceStrategyDbType;
   prepare: (context: TransformerPrepareStepContextProvider, config: Config) => void;
--- a/packages/amplify-graphql-relational-transformer/src/graphql-has-many-transformer.ts
+++ b/packages/amplify-graphql-relational-transformer/src/graphql-has-many-transformer.ts
@@ -131,7 +131,8 @@ export class HasManyTransformer extends TransformerPluginBase {
     this.directiveList.forEach((config) => {
       const modelName = config.object.name.value;
       const dbType = getStrategyDbTypeFromModel(context as TransformerContextProvider, modelName);
-      getHasManyDirectiveTransformer(dbType).prepare(context, config);
+      const dataSourceBasedTransformer = getHasManyDirectiveTransformer(dbType);
+      dataSourceBasedTransformer.prepare(context, config);
     });
   };
 
@@ -140,7 +141,8 @@ export class HasManyTransformer extends TransformerPluginBase {
 
     for (const config of this.directiveList) {
       const dbType = getStrategyDbTypeFromTypeNode(config.field.type, context);
-      getHasManyDirectiveTransformer(dbType).transformSchema(ctx, config);
+      const dataSourceBasedTransformer = getHasManyDirectiveTransformer(dbType);
+      dataSourceBasedTransformer.transformSchema(ctx, config);
       ensureHasManyConnectionField(config, context);
       extendTypeWithConnection(config, context);
     }
@@ -151,7 +153,8 @@ export class HasManyTransformer extends TransformerPluginBase {
 
     for (const config of this.directiveList) {
       const dbType = getStrategyDbTypeFromTypeNode(config.field.type, context);
-      getHasManyDirectiveTransformer(dbType).generateResolvers(ctx, config);
+      const dataSourceBasedTransformer = getHasManyDirectiveTransformer(dbType);
+      dataSourceBasedTransformer.generateResolvers(ctx, config);
       const generator = getGenerator(dbType);
       generator.makeHasManyGetItemsConnectionWithKeyResolver(config, context);
     }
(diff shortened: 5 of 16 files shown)


packages/amplify-graphql-transformer-core/src/utils/model-datasource-strategy-utils.ts
```diff
@@ -150,6 +150,19 @@ export const isSqlStrategy = (strategy: ModelDataSourceStrategy): strategy is SQ
...
+ * Provides the data source strategy for a given model
+ * @param ctx Transformer Context
+ * @param typename Model name
+ * @returns ModelDataSourceStrategyDbType
+ */
+export const getStrategyDbTypeFromModel = (ctx: DataSourceStrategiesProvider, typename: string): ModelDataSourceStrategyDbType => {
+  if (isBuiltInGraphqlType(typename)) {
+    return 'DYNAMODB';
+  }
+  return getModelDataSourceStrategy(ctx, typename).dbType;
```
atierian, line 163: This logic is based on the existing [`isSqlModel` function](https://github.com/aws-amplify/amplify-category-api/blob/main/packages/amplify-graphql-transformer-core/src/utils/model-datasource-strategy-utils.ts#L136-L142).

packages/amplify-graphql-relational-transformer/src/data-source-based-directive-transformer.ts
```diff
@@ -0,0 +1,21 @@
...
+ * Represents a subset of transformer methods based on a specific
+ * data source (currently SQL / DDB Fields).
+ */
+export interface DataSourceBasedDirectiveTransformer<Config extends DirectiveConfiguration> {
+  dbType: ModelDataSourceStrategyDbType;
+  prepare: (context: TransformerPrepareStepContextProvider, config: Config) => void;
+  transformSchema: (context: TransformerTransformSchemaStepContextProvider, config: Config) => void;
+  generateResolvers: (context: TransformerContextProvider, config: Config) => void;
+  validate: (context: TransformerContextProvider, config: Config) => void;
+}
```
atierian, line 21: One of the goals with this interface and its implementing types is to have the compiler prevent passing an inapplicable directive configuration. If you see a simpler/better way to do this, please let me know!

palpatim: Is it your goal that the interface provides type safety to prevent incorrect directive configs, or that it gives a framework for runtime validation?

packages/amplify-graphql-relational-transformer/src/graphql-has-many-transformer.ts
```diff
@@ -177,39 +151,20 @@ export class HasManyTransformer extends TransformerPluginBase {
 
     for (const config of this.directiveList) {
       const dbType = getStrategyDbTypeFromTypeNode(config.field.type, context);
-      if (dbType === DDB_DB_TYPE) {
-        updateTableForConnection(config, context);
-      }
-      makeQueryResolver(config, context, dbType);
+      getHasManyDirectiveTransformer(dbType).generateResolvers(ctx, config);
```
palpatim, line 154: We don't need to make the change now--in fact I would recommend against it in the interest of making no functional changes--but let's remember that we'll eventually have to migrate this to strategy instead of db type, since our resource generation will be different for imported gen1 DDB tables than it is for greenfield DDB tables.

packages/amplify-graphql-relational-transformer/src/has-many/has-many-directive-ddb-fields-transformer.ts
```diff
@@ -0,0 +1,40 @@
...
+  TransformerPrepareStepContextProvider,
+  TransformerTransformSchemaStepContextProvider,
+} from '@aws-amplify/graphql-transformer-interfaces';
+import { updateTableForConnection } from '../resolvers';
+import { HasManyDirectiveConfiguration } from '../types';
+import { registerHasManyForeignKeyMappings, getRelatedTypeIndex, ensureFieldsArray, getFieldsNodes } from '../utils';
+import { DataSourceBasedDirectiveTransformer } from '../data-source-based-directive-transformer';
+
+export class HasManyDirectiveDDBFieldsTransformer implements DataSourceBasedDirectiveTransformer<HasManyDirectiveConfiguration> {
+  dbType: 'DYNAMODB';
```
palpatim, line 12: as elsewhere, please use static type for this

packages/amplify-graphql-transformer-core/src/utils/model-datasource-strategy-utils.ts
```diff
@@ -150,6 +150,19 @@ export const isSqlStrategy = (strategy: ModelDataSourceStrategy): strategy is SQ
...
 
+/**
+ * Provides the data source strategy for a given model
+ * @param ctx Transformer Context
+ * @param typename Model name
+ * @returns ModelDataSourceStrategyDbType
+ */
+export const getStrategyDbTypeFromModel = (ctx: DataSourceStrategiesProvider, typename: string): ModelDataSourceStrategyDbType => {
+  if (isBuiltInGraphqlType(typename)) {
+    return 'DYNAMODB';
```
palpatim, line 161: use constant

palpatim (commented):
Please ensure e2e tests pass as well

packages/amplify-graphql-relational-transformer/src/belongs-to/belongs-to-directive-sql-transformer.ts
```diff
@@ -0,0 +1,34 @@
...
+  TransformerPrepareStepContextProvider,
+  TransformerTransformSchemaStepContextProvider,
+} from '@aws-amplify/graphql-transformer-interfaces';
+import { setFieldMappingResolverReference } from '../resolvers';
+import { BelongsToDirectiveConfiguration } from '../types';
+import { ensureReferencesArray, validateChildReferencesFields, getBelongsToReferencesNodes } from '../utils';
+import { DataSourceBasedDirectiveTransformer } from '../data-source-based-directive-transformer';
+
+export class BelongsToDirectiveSQLTransformer implements DataSourceBasedDirectiveTransformer<BelongsToDirectiveConfiguration> {
+  dbType: 'MYSQL' | 'POSTGRES';
```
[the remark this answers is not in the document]
atierian: `ModelDataSourceStrategySqlDbType` works here, thanks for the suggestion!

packages/amplify-graphql-relational-transformer/src/belongs-to/belongs-to-directive-transformer-factory.ts
```diff
@@ -0,0 +1,17 @@
...
+import { BelongsToDirectiveDDBFieldsTransformer } from './belongs-to-directive-ddb-fields-transformer';
+
+export const getBelongsToDirectiveTransformer = (
+  dbType: ModelDataSourceStrategyDbType,
+): DataSourceBasedDirectiveTransformer<BelongsToDirectiveConfiguration> => {
+  switch (dbType) {
+    case 'MYSQL':
+    case 'POSTGRES':
+      return new BelongsToDirectiveSQLTransformer(dbType);
+    case 'DYNAMODB':
```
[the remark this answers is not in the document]
atierian: From what I can tell, using the runtime constants here means we lose exhaustiveness in the switch and would have to add a `default` statement. Do the benefits of using the defined runtime constants outweigh losing exhaustiveness? Or should exhaustiveness work and I'm just holding it wrong?

```ts
// ❌ Function lacks ending return statement and return type does not include 'undefined'.ts(2366)
export const getBelongsToDirectiveTransformer = (
  dbType: ModelDataSourceStrategyDbType,
): DataSourceBasedDirectiveTransformer<BelongsToDirectiveConfiguration> => {
  switch (dbType) {
    case MYSQL_DB_TYPE:
    case POSTGRES_DB_TYPE:
      return new BelongsToDirectiveSQLTransformer(dbType);
    case DDB_DB_TYPE:
      return new BelongsToDirectiveDDBFieldsTransformer(dbType);
  }
};
```

packages/amplify-graphql-relational-transformer/src/data-source-based-directive-transformer.ts
atierian: The former. I want this to be a compiler error:
```ts
const config: HasOneDirectiveConfiguration = ...
// ❌ Argument of type 'HasOneDirectiveConfiguration' is not assignable 
// to parameter of type 'BelongsToDirectiveConfiguration'.
getBelongsToDirectiveTransformer(dbType).prepare(context, config);
```
This PR accomplishes that. But I assume there's a typescripty-er way of doing this -- hence the ask.

packages/amplify-graphql-relational-transformer/src/graphql-has-many-transformer.ts
atierian: 💯. And yes, moving from db type to strategy, despite the fact that it should just work, was a bit more than I felt comfortable than I felt comfortable with for this PR.

packages/amplify-graphql-relational-transformer/src/has-many/has-many-directive-ddb-fields-transformer.ts
atierian: You're referring to `ModelDataSourceStrategyDbType`, correct? I'd prefer not to do that here because it removes an existing compiler-enforced invariant used when selecting the correct `DataSourceBasedDirectiveTransformer`.

Here's an example to demonstrate my concerns. 
```ts
export const getHasManyDirectiveTransformer = (
  dbType: ModelDataSourceStrategyDbType,
): DataSourceBasedDirectiveTransformer<HasManyDirectiveConfiguration> => {
  switch (dbType) {
    case 'MYSQL':
    case 'POSTGRES':
      return new HasManyDirectiveDDBFieldsTransformer(dbType); // ❌  Argument of type 'ModelDataSourceStrategySqlDbType' is not assignable to parameter of type '"DYNAMODB"'.
    case 'DYNAMODB':
      return new HasManyDirectiveDDBFieldsTransformer(dbType);
  }
};
```
This is currently an error, which is what we want; by moving to `ModelDataSourceStrategyDbType`, it's no longer an error.

Is there a best of both worlds scenario here that I'm missing?

packages/amplify-graphql-relational-transformer/src/belongs-to/belongs-to-directive-ddb-fields-transformer.ts
```diff
@@ -0,0 +1,42 @@
+import {
+  TransformerContextProvider,
+  TransformerPrepareStepContextProvider,
+  TransformerTransformSchemaStepContextProvider,
+} from '@aws-amplify/graphql-transformer-interfaces';
+import { BelongsToDirectiveConfiguration } from '../types';
+import { getRelatedTypeIndex, ensureFieldsArray, getFieldsNodes, registerHasOneForeignKeyMappings } from '../utils';
+import { DataSourceBasedDirectiveTransformer } from '../data-source-based-directive-transformer';
+
+export class BelongsToDirectiveDDBFieldsTransformer implements DataSourceBasedDirectiveTransformer<BelongsToDirectiveConfiguration> {
```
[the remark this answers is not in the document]
atierian: Yes, thanks for the feedback. Addressed in 3531cfc5faaa94c6f4ca975a395f12cc73f3461e.

packages/amplify-graphql-transformer-core/src/utils/model-datasource-strategy-utils.ts
atierian: Thanks, addressed in 265f598bd80242a244ab56a1181a2added827edf.

packages/amplify-graphql-relational-transformer/src/data-source-based-directive-transformer.ts
palpatim: Thanks for clarifying. I think the approach is reasonable--you know I love me some generics--but we'll want to refine the naming a bit. Even though they aren't publicly exported, we want to be more specific about the intended usages of these types. We could change `DirectiveConfiguration` to `RelationalDirectiveConfiguration` but that leaves open the question of why `ManyToManyDirectiveConfiguration` isn't included. Naming is hard, so I will leave this to you :P

packages/amplify-graphql-relational-transformer/src/has-many/has-many-directive-ddb-fields-transformer.ts
palpatim: I'm already a seeing consistency error, which manifests as a linting warning: 

image

which is fixed with a default. I'm unclear why TS isn't able to infer exhaustiveness, since the types are defined such that there are only three possible string values. 

We can chat with other folks on the team to see if they have any ideas--in particular, the runtime folks are probably good to check with. But if we aren't able to come up with a solution that resolves the warnings, I'd prefer to go with the constants rather than hardcoded strings for value comparisons.

palpatim: By the way, my comment above for the usage in [has-many-directive-ddb-fields-transformer.ts](https://github.com/aws-amplify/amplify-category-api/pull/2330/files/d579ce982146e3d3ec6039422ab56ac294266c05#diff-9e2644d489b13192d40d6d86918daa01673645d7440a1bf15b13feeeac389d26) is incorrect -- that's a type, not a value. A constant isn't appropriate for that, and we don't have a type defined for the Dynamo DB data type, so the bare string is the right way to enforce the restriction you're after.

(We could consider adding a `type ModelDataSourceStrategyDynamoDbType = 'DYNAMODB'` and refactoring `type ModelDataSourceStrategyDbType = ModelDataSourceStrategyDynamoDbType | ModelDataSourceStrategySqlDbType` but I'm not sure whether that actually worth it.)

palpatim: This was nagging at me, so for grins I put this into [a TS 4.5.5 playground](https://www.typescriptlang.org/play?target=5&module=1&ts=4.5.5#code/C4TwDgpgBAsg9gEwgGwCIENjoMpwK4BOAxhNsAZhAOYioBGAKuNALxQDkqAmgHICCMAPKoAQuygAfWIhQYsuQiTIVg1ENgCOaRswDcAKFCRpSNJhz5ipcpRqbtTY23Ywu2AIoAZcVPYAFQWwGAHEAJQBRbHYDIjgAOwBnYChUUQB9dIYuP3CALhNZcwUrZVtaHScObn4hUWj9WMTk1w9PDJE0rJz8+FM5C0VrFTV7ekdWDhavesakqACgsMj2zuy8grN5SyUbVTstMeYoZwWQiKiDBvi5hIB3AEtgIgALBggkz0eICmRjqAAKBAVCA9GSbAYlXZqQ6QACU+TieAAtnRvscAHxQADe+igUDujxeAKB41h2NxeKgRHQCWgnF4AmEYlyFMpUAIEGAhDiUAAjAY2dTaZM3NMWWy8RyuQQeQAmAWUoV005LKLiiVS7lQADMCoAvvoDbM4MgIAA6ZBwKj-dgEp6vd7AT6qH7sAA0+Ie9reHy+Pxt1UZdVhsMus2SdpePuAAGFrlg4sk2IDgaC+kVtkMyjCIPCoIiUWiWJicXjI89icCyaXFTToKkOpk1uq2ZqZXyFXilbBRW0m91WZS23LO1S6-NAmdlv28oPJZytbqKQajdcTebLdbbV6o464010In3Z7CQ6kvukofgAGGbUxCGDEA) and it works as expected with either strategy--constant or literal.

My best guess at this point is that we have a TSConfig option that is causing this to choke.

packages/amplify-graphql-relational-transformer/src/data-source-based-directive-transformer.ts
atierian: I like the `RelationalDirectiveConfiguration` naming suggestion -- addressed in 87fae81efe5d7ff4ed850dc32f598d1a8458e636.

For consistency's sake, we could add a `DataSourceBasedDirectiveTransformer<ManyToManyDirectiveConfiguration>` with a single implementation for all data sources. I'm inclined to leave it as is for now though.

packages/amplify-graphql-relational-transformer/src/has-many/has-many-directive-ddb-fields-transformer.ts
atierian: > (We could consider adding a type ModelDataSourceStrategyDynamoDbType = 'DYNAMODB' and refactoring type ModelDataSourceStrategyDbType = ModelDataSourceStrategyDynamoDbType | ModelDataSourceStrategySqlDbType but I'm not sure whether that actually worth it.)

I actually tried this and ended up hitting a blocker.
```ts
src/types/model-datasource-strategies.ts:7:14 - error TS2322: Type '"DYNAMODB"' is not assignable to type '"DNAMODB"'.

7 export const DDB_DB_TYPE: ModelDataSourcesStrategyDynamoDbType = 'DYNAMODB';
```

Here's the patch in case anything jumps out at you.
```diff
diff --git a/packages/amplify-graphql-relational-transformer/src/belongs-to/belongs-to-directive-ddb-fields-transformer.ts b/packages/amplify-graphql-relational-transformer/src/belongs-to/belongs-to-directive-ddb-fields-transformer.ts
index 80c7d9a67..fa6b2b932 100644
--- a/packages/amplify-graphql-relational-transformer/src/belongs-to/belongs-to-directive-ddb-fields-transformer.ts
+++ b/packages/amplify-graphql-relational-transformer/src/belongs-to/belongs-to-directive-ddb-fields-transformer.ts
@@ -1,4 +1,5 @@
 import {
+  ModelDataSourcesStrategyDynamoDbType,
   TransformerContextProvider,
   TransformerPrepareStepContextProvider,
   TransformerTransformSchemaStepContextProvider,
@@ -14,8 +15,8 @@ import { DataSourceBasedDirectiveTransformer } from '../data-source-based-direct
  * This should not be used for `@belongsTo(references: [String!])` definitions.
  */
 export class BelongsToDirectiveDDBFieldsTransformer implements DataSourceBasedDirectiveTransformer<BelongsToDirectiveConfiguration> {
-  dbType: 'DYNAMODB';
-  constructor(dbType: 'DYNAMODB') {
+  dbType: ModelDataSourcesStrategyDynamoDbType;
+  constructor(dbType: ModelDataSourcesStrategyDynamoDbType) {
     this.dbType = dbType;
   }
 
diff --git a/packages/amplify-graphql-relational-transformer/src/belongs-to/belongs-to-directive-transformer-factory.ts b/packages/amplify-graphql-relational-transformer/src/belongs-to/belongs-to-directive-transformer-factory.ts
index f1e47cb23..47493e203 100644
--- a/packages/amplify-graphql-relational-transformer/src/belongs-to/belongs-to-directive-transformer-factory.ts
+++ b/packages/amplify-graphql-relational-transformer/src/belongs-to/belongs-to-directive-transformer-factory.ts
@@ -3,15 +3,16 @@ import { BelongsToDirectiveConfiguration } from '../types';
 import { DataSourceBasedDirectiveTransformer } from '../data-source-based-directive-transformer';
 import { BelongsToDirectiveSQLTransformer } from './belongs-to-directive-sql-transformer';
 import { BelongsToDirectiveDDBFieldsTransformer } from './belongs-to-directive-ddb-fields-transformer';
+import { DDB_DB_TYPE, MYSQL_DB_TYPE, POSTGRES_DB_TYPE } from '@aws-amplify/graphql-transformer-core';
 
 export const getBelongsToDirectiveTransformer = (
   dbType: ModelDataSourceStrategyDbType,
 ): DataSourceBasedDirectiveTransformer<BelongsToDirectiveConfiguration> => {
   switch (dbType) {
-    case 'MYSQL':
-    case 'POSTGRES':
+    case MYSQL_DB_TYPE:
+    case POSTGRES_DB_TYPE:
       return new BelongsToDirectiveSQLTransformer(dbType);
-    case 'DYNAMODB':
+    case DDB_DB_TYPE:
       return new BelongsToDirectiveDDBFieldsTransformer(dbType);
   }
 };
diff --git a/packages/amplify-graphql-relational-transformer/src/has-many/has-many-directive-ddb-fields-transformer.ts b/packages/amplify-graphql-relational-transformer/src/has-many/has-many-directive-ddb-fields-transformer.ts
index 2055f797a..af82a8415 100644
--- a/packages/amplify-graphql-relational-transformer/src/has-many/has-many-directive-ddb-fields-transformer.ts
+++ b/packages/amplify-graphql-relational-transformer/src/has-many/has-many-directive-ddb-fields-transformer.ts
@@ -1,4 +1,5 @@
 import {
+  ModelDataSourcesStrategyDynamoDbType,
   TransformerContextProvider,
   TransformerPrepareStepContextProvider,
   TransformerTransformSchemaStepContextProvider,
@@ -15,8 +16,8 @@ import { DataSourceBasedDirectiveTransformer } from '../data-source-based-direct
  * This should not be used for `@hasMany(references: [String!])` definitions.
  */
 export class HasManyDirectiveDDBFieldsTransformer implements DataSourceBasedDirectiveTransformer<HasManyDirectiveConfiguration> {
-  dbType: 'DYNAMODB';
-  constructor(dbType: 'DYNAMODB') {
+  dbType: ModelDataSourcesStrategyDynamoDbType;
+  constructor(dbType: ModelDataSourcesStrategyDynamoDbType) {
     this.dbType = dbType;
   }
 
diff --git a/packages/amplify-graphql-relational-transformer/src/has-many/has-many-directive-transformer-factory.ts b/packages/amplify-graphql-relational-transformer/src/has-many/has-many-directive-transformer-factory.ts
index 328243cdb..7181e109b 100644
--- a/packages/amplify-graphql-relational-transformer/src/has-many/has-many-directive-transformer-factory.ts
+++ b/packages/amplify-graphql-relational-transformer/src/has-many/has-many-directive-transformer-factory.ts
@@ -3,15 +3,16 @@ import { HasManyDirectiveConfiguration } from '../types';
 import { DataSourceBasedDirectiveTransformer } from '../data-source-based-directive-transformer';
 import { HasManyDirectiveDDBFieldsTransformer } from './has-many-directive-ddb-fields-transformer';
 import { HasManyDirectiveSQLTransformer } from './has-many-directive-sql-transformer';
+import { DDB_DB_TYPE, MYSQL_DB_TYPE, POSTGRES_DB_TYPE } from '@aws-amplify/graphql-transformer-core';
 
 export const getHasManyDirectiveTransformer = (
   dbType: ModelDataSourceStrategyDbType,
 ): DataSourceBasedDirectiveTransformer<HasManyDirectiveConfiguration> => {
   switch (dbType) {
-    case 'MYSQL':
-    case 'POSTGRES':
+    case MYSQL_DB_TYPE:
+    case POSTGRES_DB_TYPE:
       return new HasManyDirectiveSQLTransformer(dbType);
-    case 'DYNAMODB':
-      return new HasManyDirectiveDDBFieldsTransformer(dbType);
+      case DDB_DB_TYPE:
+        return new HasManyDirectiveDDBFieldsTransformer(dbType);
   }
 };
diff --git a/packages/amplify-graphql-relational-transformer/src/has-one/has-one-directive-ddb-fields-transformer.ts b/packages/amplify-graphql-relational-transformer/src/has-one/has-one-directive-ddb-fields-transformer.ts
index e0436a52c..219d10a73 100644
--- a/packages/amplify-graphql-relational-transformer/src/has-one/has-one-directive-ddb-fields-transformer.ts
+++ b/packages/amplify-graphql-relational-transformer/src/has-one/has-one-directive-ddb-fields-transformer.ts
@@ -1,4 +1,5 @@
 import {
+  ModelDataSourcesStrategyDynamoDbType,
   TransformerContextProvider,
   TransformerPrepareStepContextProvider,
   TransformerTransformSchemaStepContextProvider,
@@ -14,8 +15,8 @@ import { DataSourceBasedDirectiveTransformer } from '../data-source-based-direct
  * This should not be used for `@hasOne(references: [String!])` definitions.
  */
 export class HasOneDirectiveDDBFieldsTransformer implements DataSourceBasedDirectiveTransformer<HasOneDirectiveConfiguration> {
-  dbType: 'DYNAMODB';
-  constructor(dbType: 'DYNAMODB') {
+  dbType: ModelDataSourcesStrategyDynamoDbType;
+  constructor(dbType: ModelDataSourcesStrategyDynamoDbType) {
     this.dbType = dbType;
   }
 
diff --git a/packages/amplify-graphql-relational-transformer/src/has-one/has-one-directive-transformer-factory.ts b/packages/amplify-graphql-relational-transformer/src/has-one/has-one-directive-transformer-factory.ts
index b83e17b57..7ac49ebd2 100644
--- a/packages/amplify-graphql-relational-transformer/src/has-one/has-one-directive-transformer-factory.ts
+++ b/packages/amplify-graphql-relational-transformer/src/has-one/has-one-directive-transformer-factory.ts
@@ -3,15 +3,16 @@ import { HasOneDirectiveConfiguration } from '../types';
 import { DataSourceBasedDirectiveTransformer } from '../data-source-based-directive-transformer';
 import { HasOneDirectiveDDBFieldsTransformer } from './has-one-directive-ddb-fields-transformer';
 import { HasOneDirectiveSQLTransformer } from './has-one-directive-sql-transformer';
+import { DDB_DB_TYPE, MYSQL_DB_TYPE, POSTGRES_DB_TYPE } from '@aws-amplify/graphql-transformer-core';
 
 export const getHasOneDirectiveTransformer = (
   dbType: ModelDataSourceStrategyDbType,
 ): DataSourceBasedDirectiveTransformer<HasOneDirectiveConfiguration> => {
   switch (dbType) {
-    case 'MYSQL':
-    case 'POSTGRES':
+    case MYSQL_DB_TYPE:
+    case POSTGRES_DB_TYPE:
       return new HasOneDirectiveSQLTransformer(dbType);
-    case 'DYNAMODB':
+    case DDB_DB_TYPE:
       return new HasOneDirectiveDDBFieldsTransformer(dbType);
   }
 };
diff --git a/packages/amplify-graphql-transformer-core/src/types/model-datasource-strategies.ts b/packages/amplify-graphql-transformer-core/src/types/model-datasource-strategies.ts
index 31d986120..591694b97 100644
--- a/packages/amplify-graphql-transformer-core/src/types/model-datasource-strategies.ts
+++ b/packages/amplify-graphql-transformer-core/src/types/model-datasource-strategies.ts
@@ -1,10 +1,10 @@
 import {
   ModelDataSourceStrategy,
-  ModelDataSourceStrategyDbType,
+  ModelDataSourcesStrategyDynamoDbType,
   ModelDataSourceStrategySqlDbType,
 } from '@aws-amplify/graphql-transformer-interfaces';
 
-export const DDB_DB_TYPE: ModelDataSourceStrategyDbType = 'DYNAMODB';
+export const DDB_DB_TYPE: ModelDataSourcesStrategyDynamoDbType = 'DYNAMODB';
 
 export const MYSQL_DB_TYPE: ModelDataSourceStrategySqlDbType = 'MYSQL';
 
diff --git a/packages/amplify-graphql-transformer-core/src/utils/model-datasource-strategy-utils.ts b/packages/amplify-graphql-transformer-core/src/utils/model-datasource-strategy-utils.ts
index 5e460a0b7..6bbe22e93 100644
--- a/packages/amplify-graphql-transformer-core/src/utils/model-datasource-strategy-utils.ts
+++ b/packages/amplify-graphql-transformer-core/src/utils/model-datasource-strategy-utils.ts
@@ -5,6 +5,7 @@ import {
   ModelDataSourceStrategy,
   ModelDataSourceStrategyDbType,
   ModelDataSourceStrategySqlDbType,
+  ModelDataSourcesStrategyDynamoDbType,
   SQLLambdaModelDataSourceStrategy,
 } from '@aws-amplify/graphql-transformer-interfaces';
 import { DDB_DB_TYPE, ImportedRDSType, MYSQL_DB_TYPE, POSTGRES_DB_TYPE } from '../types';
@@ -103,8 +104,8 @@ export const isDynamoDbModel = (ctx: DataSourceStrategiesProvider, typename: str
 /**
  * Type predicate that returns true if `dbType` is the DynamoDB database type
  */
-export const isDynamoDbType = (dbType: ModelDataSourceStrategyDbType): dbType is 'DYNAMODB' => {
-  return dbType === 'DYNAMODB';
+export const isDynamoDbType = (dbType: ModelDataSourceStrategyDbType): dbType is ModelDataSourcesStrategyDynamoDbType => {
+  return dbType === DDB_DB_TYPE;
 };
 
 /**
diff --git a/packages/amplify-graphql-transformer-interfaces/src/model-datasource/types.ts b/packages/amplify-graphql-transformer-interfaces/src/model-datasource/types.ts
index ef0a20019..aa5a8c5ec 100644
--- a/packages/amplify-graphql-transformer-interfaces/src/model-datasource/types.ts
+++ b/packages/amplify-graphql-transformer-interfaces/src/model-datasource/types.ts
@@ -18,7 +18,12 @@ export interface ModelDataSourceStrategyBase {
 /**
  * All supported database types that can be used to resolve models.
  */
-export type ModelDataSourceStrategyDbType = 'DYNAMODB' | ModelDataSourceStrategySqlDbType;
+export type ModelDataSourceStrategyDbType = ModelDataSourcesStrategyDynamoDbType | ModelDataSourceStrategySqlDbType;
+
+/**
+ * DynamoDB supported database type that can be used to resovle models.
+ */
+export type ModelDataSourcesStrategyDynamoDbType = 'DNAMODB'
 
 /**
  * All supported SQL database types that can be used to resolve models.
@@ -29,7 +34,7 @@ export type ModelDataSourceStrategySqlDbType = 'MYSQL' | 'POSTGRES';
  * Use default CloudFormation type 'AWS::DynamoDB::Table' to provision table.
  */
 export interface DefaultDynamoDbModelDataSourceStrategy extends ModelDataSourceStrategyBase {
-  readonly dbType: 'DYNAMODB';
+  readonly dbType: ModelDataSourcesStrategyDynamoDbType;
   readonly provisionStrategy: 'DEFAULT';
 }
 
@@ -37,7 +42,7 @@ export interface DefaultDynamoDbModelDataSourceStrategy extends ModelDataSourceS
  * Use custom resource type 'Custom::AmplifyDynamoDBTable' to provision table.
  */
 export interface AmplifyDynamoDbModelDataSourceStrategy extends ModelDataSourceStrategyBase {
-  readonly dbType: 'DYNAMODB';
+  readonly dbType: ModelDataSourcesStrategyDynamoDbType;
   readonly provisionStrategy: 'AMPLIFY_TABLE';
 }
```

palpatim (approved):

packages/amplify-graphql-relational-transformer/src/has-many/has-many-directive-sql-transformer.ts
```diff
@@ -0,0 +1,40 @@
...
+    const modelName = config.object.name.value;
+    setFieldMappingResolverReference(context, config.relatedType?.name?.value, modelName, config.field.name.value, true);
+  };
+
+  transformSchema = (context: TransformerTransformSchemaStepContextProvider, config: HasManyDirectiveConfiguration): void => {
+    validateParentReferencesFields(config, context as TransformerContextProvider);
+  };
+
+  /** no-op */
+  generateResolvers = (_context: TransformerContextProvider, _config: HasManyDirectiveConfiguration): void => {
```
AaronZyLee, line 32: nit: Is the `_` added in the param due to the no op of this function?

packages/amplify-graphql-transformer-core/src/utils/model-datasource-strategy-utils.ts
```diff
@@ -150,6 +150,19 @@ export const isSqlStrategy = (strategy: ModelDataSourceStrategy): strategy is SQ
...
 
+/**
+ * Provides the data source strategy for a given model
+ * @param ctx Transformer Context
+ * @param typename Model name
+ * @returns ModelDataSourceStrategyDbType
+ */
+export const getStrategyDbTypeFromModel = (ctx: DataSourceStrategiesProvider, typename: string): ModelDataSourceStrategyDbType => {
+  if (isBuiltInGraphqlType(typename)) {
+    return DDB_DB_TYPE;
```
AaronZyLee, line 161: I am a little bit confused by the `DDB` type returned here. It seems that `isBuiltInGraphqlType` checks for the type `Query`, `Mutation` and `Subscription`, which are the types not annotated with `@model` and should not have an associated DB type. I can see this check is used in `isSqlModel` and `isDynamoDBModel` method which returns false. Is this just a default value returned here?

In addition, I think there is a validation in the transformer for the relational directives should be attached to the `@model` types (both parent and related). Are you aware of those validations when returning the default DB type for `Query` type here?

packages/amplify-graphql-relational-transformer/src/has-many/has-many-directive-sql-transformer.ts
atierian: It is, yes. Is there a better way to do this?

packages/amplify-graphql-transformer-core/src/utils/model-datasource-strategy-utils.ts
atierian: Great question and thanks for calling this out!

Up front here's some context on the addition:
The relational transformers `graphql-has-many-transformer`, `graphql-has-one-transformer`, `graphql-belongs-to-transformer` were previously using [`isSqlModel`](https://github.com/aws-amplify/amplify-category-api/blob/main/packages/amplify-graphql-transformer-core/src/utils/model-datasource-strategy-utils.ts#L136-L142) in a few places:
- [HasManyTransformer.prepare](https://github.com/aws-amplify/amplify-category-api/pull/2330/files#diff-04f1e3a6a8ce5bd950c9ebc3bc3ec562a443601d914dee53eeb82393825ba5d0L146)
- [HasOneTransformer.prepare](https://github.com/aws-amplify/amplify-category-api/pull/2330/files#diff-dd1c54a3c47c334e725948adce8ef86c20941fa2361ec6f864dc6c0c17f36a28L170)
- [setFieldMappingsReference](https://github.com/aws-amplify/amplify-category-api/pull/2330/files#diff-0110e5f5bea6c7251347e3013b183a69c2202d6756882aa750258296f0d617f2L247)

It was being used to determine whether the model is backed by a SQL or DDB data source. In refactoring this code, it was simpler to consistently work with a `ModelDataSourceStrategyDbType` rather than a `bool` with if / else branches. As you can see in existing usages of `isSqlModel` linked above, `true` was handled as SQL and `false` was handled as DDB. 

Based on your statements and questions, this logic seems incorrect. I presume that there's some validation happening further upstream preventing a non model type from being passed here.

Because this is a refactor, I'd like to keep this change as 100% behavioral parity with the existing implementation, even if that includes this incorrect logic. Then follow up with a fix. Is that reasonable or do you feel strongly that this should be addressed in this PR?

AaronZyLee: Since it is the refactor, I agree the behavior parity consistence should be kept first. I am OK with it in current PR. The logic can be addressed separately

AaronZyLee (approved):

PR #2331, opened 2024-03-08
feat: generic iam authorization

sobolk:
#### Description of changes

This change introduces additional mode for IAM authorization.
1. Existing `IAMAuthorizationConfig` is moved to `IdentityPoolAuthorizationConfig` (except admin roles which are now deprecated)
2. New setting `IAMAuthorizationConfig.enableIamAuthorizationMode` is introduced and works as follows:
    1. If not enabled (default) existing IAM rules execute without change (i.e. only IAM roles attached to Cognito Identity Pool work)
    2. If setting is enabled
        1. IAM roles attached to Cognito Identity Pool work as previously
        2. Deprecated admin roles work as previously
        3. Any other IAM principal is assumed to have access via it's IAM Policies and thus auth directive rules are skipped.

1. 1st e2e run without adding tests to see if existing behaviors hold true https://tiny.amazon.com/fssp5yo5/IsenLink (unit tests are failing due to lack of coverage for now).

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes


packages/amplify-graphql-api-construct/src/types.ts
```diff
@@ -139,6 +179,12 @@ export interface AuthorizationModes {
    */
```
palpatim, line 179: L177 above now has an incomplete comment. We should work with @renebrandel for wording, but maybe something like "IAM Auth config, required if an 'iam' auth provider is specified in the API, or if specifying that all IAM authorization should be controlled by IAM policies rather than `@auth` directives."

packages/amplify-graphql-auth-transformer/src/vtl-generator/ddb/resolvers/helpers.ts
```diff
@@ -165,12 +166,31 @@ export const iamExpression = (
...
   }
+
   return iff(equals(ref('util.authType()'), str(IAM_AUTH_TYPE)), compoundExpression(expression));
 };
 
+/**
+ * Creates an expression that allows generic IAM access for principals not associated to CognitoIdentityPool
+ */
+export const genericIamAccessExpression = (): Expression => {
+  const doesNotHaveIdentityPoolExpression = and([
```
palpatim, line 184: Thinking through this--I think `and` is the right expression to use here. If either identityId or identityPoolId is missing, the identity pool evaluation should fail, but we don't want to automatically fall back to IAM authorization if it's in an inconsistent state like that. 

We should verify that there are no cases where one is present without the other, but I think this is right.

packages/amplify-graphql-auth-transformer/src/vtl-generator/ddb/resolvers/helpers.ts
```diff
@@ -165,12 +166,31 @@ export const iamExpression = (
         iff(not(ref(IS_AUTHORIZED_FLAG)), iamCheck(role.claim!, set(ref(IS_AUTHORIZED_FLAG), bool(true)), hasIdentityPoolId)),
       );
     });
-  } else {
+  }
+
+  if (genericIamAccessEnabled) {
```
palpatim, line 171: Is this the right place to evaluate genericIamAccess or do we need to do it before L159? Is `roles` populated for generic IAM cases? Should we only evaluate non-IAM cases if the IAM conditions don't apply, rather than evaluating every condition?

packages/amplify-graphql-auth-transformer/src/vtl-generator/ddb/resolvers/mutation.create.ts
```diff
@@ -85,9 +91,14 @@ const iamExpression = (roles: Array<RoleDefinition>, hasAdminRolesEnabled = fals
         );
       }
     });
-  } else {
+  }
+
+  if (genericIamAccessEnabled) {
```
palpatim, line 96: As above, is this the right order of evaluation, or should it happen before checking for Admin; and should we only evaluate either the "legacy" or "new" branch rather than both?

packages/amplify-graphql-auth-transformer/src/vtl-generator/ddb/resolvers/mutation.delete.ts
```diff
@@ -66,9 +72,14 @@ const iamExpression = (roles: Array<RoleDefinition>, hasAdminRolesEnabled = fals
     roles.forEach((role) => {
       expression.push(iamCheck(role.claim!, set(ref(IS_AUTHORIZED_FLAG), bool(true)), hasIdentityPoolId));
     });
-  } else {
+  }
+
+  if (genericIamAccessEnabled) {
```
palpatim, line 77: ditto

packages/amplify-graphql-auth-transformer/src/vtl-generator/ddb/resolvers/mutation.update.ts
```diff
@@ -112,9 +118,14 @@ const iamExpression = (roles: Array<RoleDefinition>, hasAdminRolesEnabled = fals
         );
       }
     });
-  } else {
+  }
+
+  if (genericIamAccessEnabled) {
```
palpatim, line 123: ditto

packages/amplify-graphql-auth-transformer/src/vtl-generator/ddb/resolvers/search.ts
```diff
@@ -101,6 +105,13 @@ const iamExpression = (roles: Array<RoleDefinition>, hasAdminRolesEnabled = fals
       expression.push(iff(not(ref(IS_AUTHORIZED_FLAG)), iamCheck(role.claim!, compoundExpression(exp), hasIdentityPoolId)));
     });
   }
+
+  if (genericIamAccessEnabled) {
```
palpatim, line 109: ditto

packages/amplify-graphql-auth-transformer/src/vtl-generator/ddb/resolvers/helpers.ts
sobolk: Sounds good. I'll add something to handle inconsistent state - likely a validation that fails request (unauthorized seems to be wrong error for this case).

packages/amplify-graphql-auth-transformer/src/vtl-generator/ddb/resolvers/helpers.ts
sobolk: I was reverse engineering this from test snapshots by looking at expressions there.
I also added a diagram with desired algorithm in PR description.

Whether this is right place or 159 - I intend to figure this out though new tests that I'm yet to add. For now this is the best effort.
One thing I checked so far is that when this is disabled all existing tests are passing.

packages/amplify-graphql-auth-transformer/src/vtl-generator/rds/resolvers/common.ts
```diff
@@ -268,3 +277,20 @@ export const generateFieldResolverForOwner = (entity: string): string => {
...
+export const generateIAMAccessCheck = (enableIamAccess: boolean, expression: Expression): Expression => {
+  if (!enableIamAccess) {
+    // No-op if generic IAM access is not enabled.
+    return expression;
+  }
+  const isNonCognitoIAMPrincipal = and([
+    equals(ref('util.authType()'), str(IAM_AUTH_TYPE)),
+    methodCall(ref('util.isNull'), ref('ctx.identity.cognitoIdentityPoolId')),
+    methodCall(ref('util.isNull'), ref('ctx.identity.cognitoIdentityId')),
+  ]);
```
sobolk, line 294: TODO: we need consistency check here as well.

packages/amplify-graphql-auth-transformer/src/vtl-generator/ddb/resolvers/helpers.ts
sobolk: I ended up reversing this check and making DDB generators look like RDS.
Not sure if consistency check is needed now.

packages/amplify-graphql-auth-transformer/src/vtl-generator/ddb/resolvers/helpers.ts
sobolk: I ended up refactoring this to look similar to RDS, i.e. just wrapping old rules with if/else.

packages/graphql-transformers-e2e-tests/src/__tests__/AuthV2ExhaustiveT3A.test.ts
```diff
@@ -0,0 +1,88 @@
+import { AuthProvider, AuthStrategy, ModelOperation } from '@aws-amplify/graphql-auth-transformer';
```
sobolk, line 1: `AuthV2ExhaustiveT3A.test.ts` are pretty much copy of `AuthV2ExhaustiveT2A.test.ts` but with new setting.

This is to assert that iam directive works as before while new setting is one.

(this applies to 3 similar tests)

packages/graphql-transformers-e2e-tests/src/__tests__/AuthV2TransformerIAM.test.ts
```diff
@@ -0,0 +1,952 @@
+import { PrimaryKeyTransformer } from '@aws-amplify/graphql-index-transformer';
```
sobolk, line 1: This file is new test that covers multiple combinations of cognito and non-cognito access patterns.

packages/amplify-graphql-api-construct/src/types.ts
sobolk: updated.
I didn't elaborate about `iam` provider since we're deprecating it and moving that config to `IdentityPoolAuthorizationConfig`

renebrandel: it's not a `rather` as the use cases are slightly different. I'd suggest this:
```suggestion
IAM Auth config, required to allow IAM-based access to this API. This applies to any IAM principal except Amazon Cognito identity pool's authenticated and unauthenticated roles. This behavior was has recently been improved. See https://docs.amplify.aws/cli/react/tools/cli/migration/iam-auth-updates-for-cdk-construct for details.
```

renebrandel:
Corresponding docs PR: https://github.com/aws-amplify/docs/pull/7056

packages/amplify-graphql-api-construct-tests/src/__tests__/backends/sql-models/app.ts
```diff
@@ -41,11 +43,29 @@ const stack = new Stack(app, packageJson.name.replace(/_/g, '-'), {
   env: { region: process.env.CLI_REGION || 'us-west-2' },
 });
 
+const auth = new AmplifyAuth(stack, 'Auth');
+
+const enableIamAuthorizationMode = process.env.ENABLE_IAM_AUTHORIZATION_MODE === 'true';
```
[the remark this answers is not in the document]
sobolk: The problem is that there's a child process spawn involved between test runner and this code. I.e. in test we call `cdk deploy` that runs this code in child process.

So options to are:
1. Use env variable
2. Use CDK context
3. Duplicate `sql-models/app.ts` to accommodate new behavior (not recommended given overlap and intent - to test same app with just one knob).

Let me know if which one you'd prefer.
FWIW, in gen 2 we went with this https://github.com/aws-amplify/amplify-backend/blob/591b08ee98d49467d07a713b63ed91c40ee2b285/packages/integration-tests/src/test-cdk-projects/auth-cdk/test-cdk-project-auth-cdk-stack.ts#L8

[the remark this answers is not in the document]
phani-srikar: If we go with using CDK contexts, can we set this explicitly per test in code? I would prefer that option if we can.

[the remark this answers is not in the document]
sobolk: I do not see a big difference between cdk context and env variables from testing code pov. They're both k-v string:string maps that are passed to child process. 

What we can do to somewhat aid test author is to make the `app.ts` throw if param is not provided explicitly - this should prevent error on test author side.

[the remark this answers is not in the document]
phani-srikar: I was expecting using the CDK context to be more readable but its not much according to your explanation, thanks.

> make the app.ts throw if param is not provided explicitly - this should prevent error on test author side.

That would be a nice to a have validation.

packages/amplify-graphql-model-transformer/src/graphql-model-transformer.ts
```diff
@@ -319,22 +319,47 @@ export class ModelTransformer extends TransformerModelBase implements Transforme
...
+          serviceDirectiveNames.add(API_KEY_DIRECTIVE);
+          if (defaultServiceDirectiveName) {
+            serviceDirectiveNames.add(defaultServiceDirectiveName);
+          }
+        } else if (ctx.synthParameters.enableIamAccess && ctx.authConfig.defaultAuthentication.authenticationType !== 'AWS_IAM') {
+          serviceDirectiveNames.add(AWS_IAM_DIRECTIVE);
+          if (defaultServiceDirectiveName) {
+            serviceDirectiveNames.add(defaultServiceDirectiveName);
+          }
+        }
```
sobolk, line 344: todo: remove default fallback.
if there's no `@auth` we should just check sandbox and iam check.

sobolk: this is fixed by https://github.com/aws-amplify/amplify-category-api/pull/2331/commits/5935803b0aea7f48194cd32aa4534f70c7a353d9

packages/amplify-graphql-api-construct-tests/src/__tests__/backends/sql-models/app.ts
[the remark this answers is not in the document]
sobolk: This is what I came up with https://github.com/aws-amplify/amplify-category-api/pull/2331/commits/260c8efb17feedf7c1458eb3472bc9b465346307 .

packages/amplify-graphql-api-construct/src/__tests__/__functional__/auth-modes.test.ts
```diff
@@ -161,4 +161,133 @@ describe('auth modes', () => {
...
+              identityPoolId: identityPool.logicalId,
+              authenticatedUserRole,
+              unauthenticatedUserRole,
+              enableIamAuthorizationMode: true,
+            },
+          },
+        });
+      });
+    }).toThrow(
+      "Cannot use deprecated 'authorizationModes.iamConfig' options with 'authorizationModes.identityPoolConfig' " +
```
sundersc, line 231: The error message is not clear. Can we change the wordings? Something like `Invalid authorization modes configuration provided. Deprecated IAM configuration cannot be used with identity pool configuration.`.

packages/amplify-graphql-api-construct/src/internal/authorization-modes.ts
```diff
@@ -5,19 +5,61 @@ import { IRole, ServicePrincipal } from 'aws-cdk-lib/aws-iam';
...
   ApiKeyAuthorizationConfig,
-  IAMAuthorizationConfig,
   LambdaAuthorizationConfig,
   OIDCAuthorizationConfig,
   UserPoolAuthorizationConfig,
 } from '../types';
 
 type AuthorizationConfigMode =
-  | (IAMAuthorizationConfig & { type: 'AWS_IAM' })
+  | { type: 'AWS_IAM' }
```
sundersc, line 14: Why are we removing the `IAMAuthorizationConfig` type?

packages/amplify-graphql-api-construct/src/internal/authorization-modes.ts
```diff
@@ -5,19 +5,61 @@ import { IRole, ServicePrincipal } from 'aws-cdk-lib/aws-iam';
...
+ *
+ * Rules:
+ * 1. Validates that deprecated settings ('iamConfig.authenticatedUserRole', 'iamConfig.unauthenticatedUserRole',
+ *    'iamConfig.identityPoolId', 'iamConfig.allowListedRoles' and 'adminRoles') are mutually exclusive with new settings that
+ *    replaced them ('iamConfig.enableIamAuthorizationMode' and any of 'authorizationModes.identityPoolConfig')
+ * 2. If deprecated identity pool settings are used ('iamConfig.authenticatedUserRole', 'iamConfig.unauthenticatedUserRole',
+ *    and 'iamConfig.identityPoolId') validate that all are provided.
+ */
+export const validateAuthorizationModes = (authorizationModes: AuthorizationModes): void => {
+  const hasDeprecatedIdentityPoolSettings =
```
sundersc, line 31: Should this variable be named `hasDeprecatedIAMSettings`?

packages/amplify-graphql-api-construct/src/internal/authorization-modes.ts
sobolk: Because the `type AuthorizationConfigMode` is only used locally in this file for type checking below and non of `IAMAuthorizationConfig` properties are used in these lines (which was confusing when I was reading this code first time).
https://github.com/aws-amplify/amplify-category-api/blob/11336ca59d035e5b57b62e11c583158291c9b8b4/packages/amplify-graphql-api-construct/src/internal/authorization-modes.ts#L36-L37

packages/amplify-graphql-api-construct/src/internal/authorization-modes.ts
sobolk: Not this one. This one is specifically targetting identity pool (authrole, unauthrole and identity pool) for the second rule I mention in the doc.

However, there are two variables below that should get 'Iam` word in the name. I'll add it there to make it clear.

packages/amplify-graphql-api-construct/src/__tests__/__functional__/auth-modes.test.ts
sobolk: Updated.

packages/amplify-graphql-auth-transformer/src/__tests__/resolvers/__snapshots__/subscriptions.test.ts.snap
```diff
@@ -14,18 +14,65 @@ $util.toJson({\\"version\\":\\"2018-05-29\\",\\"payload\\":{}})
...
+"## [Start] Authorization Steps. **
+$util.qr($ctx.stash.put(\\"hasAuth\\", true))
+#set( $isAuthorized = false )
+#if( $util.authType() == \\"IAM Authorization\\" )
+  #if( $util.isNull($ctx.identity.cognitoIdentityPoolId) && $util.isNull($ctx.identity.cognitoIdentityId) )
+    $util.qr($ctx.stash.put(\\"hasAuth\\", true))
+    #set( $isAuthorized = true )
+  #else
+    #if( !$isAuthorized )
+      #if( $ctx.identity.userArn == $ctx.stash.testClaim )
```
sundersc, line 46: Where are we setting this stash variable?

packages/amplify-graphql-auth-transformer/src/vtl-generator/ddb/resolvers/field.ts
```diff
@@ -204,9 +215,18 @@ export const setDeniedFieldFlag = (operation: string, subscriptionsEnabled: bool
...
+    exp = iff(notEquals(methodCall(ref('util.authType')), str(API_KEY_AUTH_TYPE)), exp);
+  }
+  if (genericIamAccessEnabled) {
+    const isNonCognitoIAMPrincipal = and([
+      equals(ref('util.authType()'), str(IAM_AUTH_TYPE)),
+      methodCall(ref('util.isNull'), ref('ctx.identity.cognitoIdentityPoolId')),
+      methodCall(ref('util.isNull'), ref('ctx.identity.cognitoIdentityId')),
+    ]);
+    exp = iff(not(parens(isNonCognitoIAMPrincipal)), exp);
+  }
```
sundersc, line 230: Can sandbox and IAM access be enabled at the time? It looks like if IAM access is enabled, API Key sandbox expression won't get included in the generated resolver.

packages/amplify-graphql-auth-transformer/src/__tests__/resolvers/__snapshots__/subscriptions.test.ts.snap
sobolk: here https://github.com/aws-amplify/amplify-category-api/blob/67290951da610989476d366703b7722229eedbf1/packages/amplify-graphql-auth-transformer/src/__tests__/resolvers/subscriptions.test.ts#L29-L35 .

packages/amplify-graphql-auth-transformer/src/vtl-generator/ddb/resolvers/field.ts
sobolk: > Can sandbox and IAM access be enabled at the time?

Yes.

> It looks like if IAM access is enabled, API Key sandbox expression won't get included in the generated resolver.

Sandbox works in such a way that:
1. necessary service directives are added to types and operations
2. `postAuth` resolver is added to handle sandbox auth. See sample `postAuth` resolvers handling sandbox
   1. https://github.com/aws-amplify/amplify-category-api/blob/67290951da610989476d366703b7722229eedbf1/packages/amplify-graphql-auth-transformer/src/__tests__/__snapshots__/iam-access.test.ts.snap#L34-L41
   2. https://github.com/aws-amplify/amplify-category-api/blob/11336ca59d035e5b57b62e11c583158291c9b8b4/packages/amplify-graphql-searchable-transformer/src/__tests__/__snapshots__/generate-resolver-vtl.test.ts.snap#L459-L466
   
   
I can add couple of tests that focus on this particular aspect.

packages/amplify-graphql-auth-transformer/src/vtl-generator/rds/resolvers/common.ts
```diff
@@ -211,10 +212,19 @@ export const constructAuthorizedInputStatement = (keyName: string): Expression =
...
+  let exp: Expression = methodCall(ref('util.unauthorized'));
+  if (sandboxEnabled) {
+    exp = iff(notEquals(methodCall(ref('util.authType')), str(API_KEY_AUTH_TYPE)), exp);
+  }
+  if (genericIamAccessEnabled) {
+    const isNonCognitoIAMPrincipal = and([
+      equals(ref('util.authType()'), str(IAM_AUTH_TYPE)),
+      methodCall(ref('util.isNull'), ref('ctx.identity.cognitoIdentityPoolId')),
+      methodCall(ref('util.isNull'), ref('ctx.identity.cognitoIdentityId')),
+    ]);
```
sundersc, line 225: I see this block is used atleast in two places. Can we move it to a helper method?

packages/amplify-graphql-auth-transformer/src/__tests__/resolvers/__snapshots__/subscriptions.test.ts.snap
sobolk: addressed in https://github.com/aws-amplify/amplify-category-api/pull/2331/commits/a10d9bbc008fffbfdf691e7e57f33f2e6e7c6403

packages/amplify-graphql-auth-transformer/src/vtl-generator/ddb/resolvers/field.ts
sobolk: added couple of tests here :

https://github.com/aws-amplify/amplify-category-api/pull/2331/commits/355ebb07272c4ecb53f475be76529e0469718fe2

https://github.com/aws-amplify/amplify-category-api/pull/2331/commits/c970c52ed8c9a492b48761ecb00bea37690f327f

packages/amplify-graphql-auth-transformer/src/vtl-generator/rds/resolvers/common.ts
sobolk: I moved things around in https://github.com/aws-amplify/amplify-category-api/pull/2331/commits/b70d224ab3ce645a0ac1e3598611efd5ee7fcd8e .

Notice that:
1. I didn't change model transformer as it doesn't share code with auth transformer (I don't think it's worth moving this piece to transformer core, there's already precedent of auth stuff being copy-simplify-paste to model transformer)
2. I also moved other expression that is now shared between rds and ddb.

phani-srikar (approved):

scripts/split-e2e-tests.ts
```diff
@@ -153,6 +153,9 @@ const RUN_IN_NON_OPT_IN_REGIONS: (string | RegExp)[] = [
   /src\/__tests__\/rds-.*\.test\.ts/,
   // Searchable tests
   /src\/__tests__\/.*searchable.*\.test\.ts/,
+  // Tests that use Auth Construct
+  'src/__tests__/ddb-iam-access.test.ts',
+  'src/__tests__/sql-models.test.ts',
```
sobolk, line 158: see https://github.com/aws-amplify/amplify-backend/issues/1194

packages/amplify-graphql-auth-transformer/src/__tests__/resolvers/__snapshots__/subscriptions.test.ts.snap
sundersc: The linked code block doesn't explain where we set the stash variable.

packages/amplify-graphql-auth-transformer/src/vtl-generator/ddb/resolvers/field.ts
sundersc: Thanks for adding more cases. Left a comment to avoid nested auth check.

packages/amplify-graphql-auth-transformer/src/vtl-generator/ddb/resolvers/helpers.ts
```diff
@@ -145,30 +143,54 @@ export const lambdaExpression = (roles: Array<RoleDefinition>): Expression =>
...
+ *
+ * Note: This function assumes that caller already checked that auth type is IAM. Should not be used outside of IAM helpers.
+ */
+export const generateIAMAccessCheck = (enableIamAccess: boolean, expression: Expression): Expression => {
+  if (!enableIamAccess) {
+    // No-op if generic IAM access is not enabled.
+    return expression;
+  }
+
+  const isGenericIamAccess = and([
```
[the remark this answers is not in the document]
sobolk: It's because all these calls are nested under `iamExpression` so we'd be repeating auth type check.
See https://github.com/aws-amplify/amplify-category-api/blob/82ae3aac071d64acbe4727631850ef7ee8f88324/packages/amplify-graphql-auth-transformer/src/__tests__/__snapshots__/iam-access.test.ts.snap#L645-L654 .

I.e. 
instead of 
```
#if( $util.authType() == \\"IAM Authorization\\" )
  #if( $util.isNull($ctx.identity.cognitoIdentityPoolId) && $util.isNull($ctx.identity.cognitoIdentityId) )
    $util.qr($ctx.stash.put(\\"hasAuth\\", true))
    #set( $isAuthorized = true )
  #else
    #if( $ctx.identity.userArn == $ctx.stash.unauthRole )
      #set( $isAuthorized = true )
    #end
  #end
#end
```

We'd have (double `$util.authType() == \\"IAM Authorization\\"`).
```
#if( $util.authType() == \\"IAM Authorization\\" )
  #if( $util.authType() == \\"IAM Authorization\\" && $util.isNull($ctx.identity.cognitoIdentityPoolId) && $util.isNull($ctx.identity.cognitoIdentityId) )
    $util.qr($ctx.stash.put(\\"hasAuth\\", true))
    #set( $isAuthorized = true )
  #else
    #if( $ctx.identity.userArn == $ctx.stash.unauthRole )
      #set( $isAuthorized = true )
    #end
  #end
#end
```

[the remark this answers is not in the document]
sobolk: Having said that. It's fine to reuse `isNonCognitoIAMPrincipal` if we don't mind double-check (I guess we don't but looking for confirmation).

packages/amplify-graphql-model-transformer/src/graphql-model-transformer.ts
```diff
@@ -319,22 +319,35 @@ export class ModelTransformer extends TransformerModelBase implements Transforme
...
-        });
-        subscriptionsFields.forEach((operationField) => {
-          const operationName = operationField.name.value;
-          addDirectivesToOperation(ctx, ctx.output.getSubscriptionTypeName()!, operationName, apiKeyDirArray);
-        });
+      if (!hasAuth) {
+        const serviceDirectiveNames = new Set<string>();
+
+        if (ctx.transformParameters.sandboxModeEnabled && ctx.synthParameters.enableIamAccess) {
+          serviceDirectiveNames.add(API_KEY_DIRECTIVE);
```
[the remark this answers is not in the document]
sobolk: If there are multiple (2+) access modes enabled then all models must be annotated with respective service directives explicitly regardless of default to allow both access.
Enabling both iam and sandbox mean that both IAM and api key are supposed to work.

The (appsync) rules are:
1. If model is not annotated then only default mode works for it. (hence checks below that check for situations that desired mode is not default)
2. If model is annotated with any service directive then local annotation wins and default mode is taken into account. (hence need to add both api key and iam if both are enabled).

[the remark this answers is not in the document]
sobolk: > Also, I may have missed it in the tests above, but do we have tests for sandbox mode?

Yes we do. I expanded them to add (sandbox + iam) cases.

packages/amplify-graphql-auth-transformer/src/__tests__/__snapshots__/iam-access.test.ts.snap
```diff
@@ -0,0 +1,3542 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`ddb generates related type sandbox resolvers when both sandbox and iam access enabled 1`] = `
+"## [Start] Sandbox Mode Enabled. **
+#if( !$ctx.stash.get(\\"hasAuth\\") )
+  #if( !($util.authType() == \\"IAM Authorization\\" && $util.isNull($ctx.identity.cognitoIdentityPoolId) && $util.isNull($ctx.identity.cognitoIdentityId)) )
+    #if( $util.authType() != \\"API Key Authorization\\" )
+      $util.unauthorized()
+    #end
+  #end
```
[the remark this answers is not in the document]
sobolk: I'll give it a try.

packages/amplify-graphql-auth-transformer/src/__tests__/resolvers/__snapshots__/subscriptions.test.ts.snap
sobolk: The call stack is:
1. https://github.com/aws-amplify/amplify-category-api/blob/11336ca59d035e5b57b62e11c583158291c9b8b4/packages/amplify-graphql-auth-transformer/src/vtl-generator/ddb/resolvers/helpers.ts#L56-L57
2. https://github.com/aws-amplify/amplify-category-api/blob/11336ca59d035e5b57b62e11c583158291c9b8b4/packages/amplify-graphql-auth-transformer/src/vtl-generator/ddb/resolvers/helpers.ts#L165
3. https://github.com/aws-amplify/amplify-category-api/blob/11336ca59d035e5b57b62e11c583158291c9b8b4/packages/amplify-graphql-auth-transformer/src/vtl-generator/ddb/resolvers/subscriptions.ts#L173
4. https://github.com/aws-amplify/amplify-category-api/blob/11336ca59d035e5b57b62e11c583158291c9b8b4/packages/amplify-graphql-auth-transformer/src/__tests__/resolvers/subscriptions.test.ts#L103

image

sobolk: Discussed offline.
This appears to be relevant place https://github.com/aws-amplify/amplify-category-api/blob/11336ca59d035e5b57b62e11c583158291c9b8b4/packages/amplify-graphql-auth-transformer/src/graphql-auth-transformer.ts#L1103 .
I'll update test values to match it.

packages/graphql-transformers-e2e-tests/src/IAMHelper.ts
```diff
@@ -163,12 +165,57 @@ export class IAMHelper {
...
+  async createAppSyncDataPolicy(policyName: string, region: string, appsyncApiIds: Array<string>) {
+    const accountDetails = await this.sts.getCallerIdentity({}).promise();
+    const currentAccountId = accountDetails.Account;
+    const policyStatement = {
+      Version: '2012-10-17',
+      Statement: [
+        {
+          Effect: 'Allow',
+          Action: ['appsync:GraphQL'],
+          Resource: appsyncApiIds.map((appsyncApiId) => `arn:aws:appsync:${region}:${currentAccountId}:apis/${appsyncApiId}/*`),
```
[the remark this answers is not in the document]
palpatim: Fair point. The case I'm specifically concerned about is a regression where an IAM control we accidentally apply fine-grained auth rules _in addition to_ the IAM rules. It may be that this case is covered by existing tests, but the schema I remember from this PR are pretty coarse grained, and the case I'm wondering about is a schema with field-level auth:

```graphql
type Todo @model @auth(rules: [
{ allow: owner },
{ allow: public, operations: [read] },
]) {
  id: ID!
  name: String!
  description: String
  secret: String @auth(rules: [{allow: owner}])
}
```

packages/amplify-graphql-auth-transformer/src/vtl-generator/ddb/resolvers/helpers.ts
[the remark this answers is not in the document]
palpatim: I see your point. My concern is that someone is going to incorrectly use `generateIAMAccessCheck` outside of a block that has already asserted IAM auth exists. This discussion suggests that we should probably separate this out further, but I'm reluctant to take on any more refactoring as part of this change.

My vote would be that it's OK to duplicate the generated code in the interests of a secure-by-default utility method.

packages/graphql-transformers-e2e-tests/src/IAMHelper.ts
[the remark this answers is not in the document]
sobolk: I can add such test.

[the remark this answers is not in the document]
sobolk: However, after looking once again. it seems that this is covered by `AuthV2ExhaustiveT3B.test.ts` and `AuthV2ExhaustiveT3D.test.ts` which generate schemas with "partial access" .
I.e. I copied existing `AuthV2ExhaustiveT2...` tests that cover `iam` provider in auth directive and run them with iam access setting enabled. They assert that for cognito traffic directive works as before.

packages/amplify-graphql-auth-transformer/src/__tests__/resolvers/__snapshots__/subscriptions.test.ts.snap
sobolk: Addressed here https://github.com/aws-amplify/amplify-category-api/pull/2331/commits/078fd69f3eac0c8080f144fb3825b2cbc9cfa6cf

packages/amplify-graphql-model-transformer/src/graphql-model-transformer.ts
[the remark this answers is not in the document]
sobolk: Added comment https://github.com/aws-amplify/amplify-category-api/pull/2331/commits/8b2eca12d7fb9a2548479096294daa9264bbbef0 .

packages/amplify-graphql-auth-transformer/src/vtl-generator/ddb/resolvers/helpers.ts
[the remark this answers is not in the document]
sobolk: changed in https://github.com/aws-amplify/amplify-category-api/pull/2331/commits/7eae99b5dc8e5965842a38ddc41efe4410ba614b

packages/graphql-transformers-e2e-tests/src/IAMHelper.ts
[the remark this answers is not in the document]
sobolk: also added tests in 
https://github.com/aws-amplify/amplify-category-api/pull/2331/commits/ca778e9be3a204d96b8407e0d22c9bbd580e67b5
and 
https://github.com/aws-amplify/amplify-category-api/pull/2331/commits/8602e341692633ab65c106d9abed31072a2df297

palpatim (dismissed):

palpatim (approved):

sundersc (approved):

phani-srikar (approved):

sobolk:
Tests green except known `schema_searchable` problem. https://tiny.amazon.com/15wj5u0l/IsenLink

PR #2342, opened 2024-03-13
refactor: move directive definitions to a single source of truth

dpilch:
#### Description of changes

Move directive definitions to a single source of truth. Transform logic will still be contained in the transformers. This change will allow codegen to depend the `@aws-amplify/graphql-directives` package and use the definitions as the default definitions of model generation.

PR to use new package in codegen: https://github.com/aws-amplify/amplify-codegen/pull/796

N/A

#### Issue #, if available

N/A

* Unit tests
* E2E tests

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-graphql-auth-transformer/API.md
```diff
@@ -64,7 +64,7 @@ export interface AuthDirective {
 }
 
 // @public (undocumented)
-export const authDirectiveDefinition = "\n  directive @auth(rules: [AuthRule!]!) on OBJECT | FIELD_DEFINITION\n  input AuthRule {\n    allow: AuthStrategy!\n    provider: AuthProvider\n    identityClaim: String\n    groupClaim: String\n    ownerField: String\n    groupsField: String\n    groups: [String]\n    operations: [ModelOperation]\n  }\n  enum AuthStrategy {\n    owner\n    groups\n    private\n    public\n    custom\n  }\n  enum AuthProvider {\n    apiKey\n    iam\n    oidc\n    userPools\n    function\n  }\n  enum ModelOperation {\n    create\n    update\n    delete\n    read\n    list\n    get\n    sync\n    listen\n    search\n  }\n";
+export const authDirectiveDefinition: string;
```
dpilch, line 67: The type is changing to string here, but the value is still the same.

atierian (approved):

packages/amplify-graphql-directives/src/directives/v1/connection.ts
```diff
@@ -0,0 +1,20 @@
...
+    keyField: String
+    sortField: String
+    keyName: String
+    limit: Int
+    fields: [String!]
+  ) on FIELD_DEFINITION
+`;
+const defaults = {};
+
+export const ConnectionDirective: Directive = {
```
palpatim, line 16: I'm a bit torn, but I wonder if we should just go ahead & name this `ConnectionDirectiveV1`? I realize we don't technically need to disambiguate it, but having the built-in marker for these legacy directives is kind of nice. Thoughts?

packages/amplify-graphql-directives/src/directives/v1/key.ts
```diff
@@ -0,0 +1,13 @@
+import { Directive } from '../directive';
+
+const name = 'key';
+const definition = /* GraphQL */ `
+  directive @${name}(name: String, fields: [String!]!, queryField: String) repeatable on OBJECT
+`;
+const defaults = {};
+
+export const KeyDirective: Directive = {
```
palpatim, line 9: Ditto question about naming: `KeyDirectiveV1`?

packages/amplify-graphql-directives/src/directives/v1/versioned.ts
```diff
@@ -0,0 +1,17 @@
...
+const name = 'versioned';
+const defaults: VersionedDirectiveDefaults = {
+  versionField: 'version',
+  versionInput: 'expectedVersion',
+};
+const definition = /* GraphQL */ `
+  directive @${name}(versionField: String = "${defaults.versionField}", versionInput: String = "${defaults.versionInput}") on OBJECT
+`;
+
+export const VersionedDirective: Directive<VersionedDirectiveDefaults> = {
```
palpatim, line 13: ditto `VersionedDirectiveV1`?

packages/amplify-graphql-name-mapping-transformer/src/graphql-refers-to-transformer.ts
```diff
@@ -113,13 +108,15 @@ export const shouldBeAppliedToRDSModels = (
 ): void => {
   const modelName = definition.name.value;
   if (!isSqlModel(ctx, modelName)) {
-    throw new Error(`@${directiveName} is only supported on RDS models. ${modelName} is not an RDS model.`);
+    throw new Error(`@${RefersToDirective.name} is only supported on RDS models. ${modelName} is not an RDS model.`);
```
palpatim, line 111: While we're in here:
```suggestion
    throw new Error(`@${RefersToDirective.name} is only supported on SQL models. ${modelName} is not a SQL model.`);
```

packages/amplify-graphql-transformer-core/src/transformation/validation.ts
```diff
@@ -106,18 +116,18 @@ scalar BigInt
...
-directive @aws_api_key on FIELD_DEFINITION | OBJECT
-directive @aws_iam on FIELD_DEFINITION | OBJECT
-directive @aws_oidc on FIELD_DEFINITION | OBJECT
-directive @aws_cognito_user_pools(cognito_groups: [String!]) on FIELD_DEFINITION | OBJECT
-directive @aws_lambda on FIELD_DEFINITION | OBJECT
-
-# Allows transformer libraries to deprecate directive arguments.
-directive @deprecated(reason: String) on FIELD_DEFINITION | INPUT_FIELD_DEFINITION | ENUM | ENUM_VALUE
-`);
+export const EXTRA_DIRECTIVES_DOCUMENT = parse(
```
palpatim, line 119: Good find! I didn't see this when I was looking for definitions!

packages/amplify-graphql-transformer/src/graphql-transformer.ts
```diff
@@ -61,6 +61,7 @@ export const constructTransformerChain = (options?: TransformerFactoryArgs): Tra
   const indexTransformer = new IndexTransformer();
   const hasOneTransformer = new HasOneTransformer();
 
+  // The default list of transformers should match DefaultDirectives in packages/amplify-graphql-directives/src/index.ts
```
palpatim, line 64: I know what you're getting at, but this list doesn't match because we don't have transformers for the AppSync directives. A couple of things I can think of:
1. Create new structures to house AppSync directives and Amplify directives, and compose `DefaultDirectives` from it and the list of Amplify directives
2. Include the list of supported AppSync directives in a comment so at least the number of directive definitions matches the number of transformer definitions

I slightly favor 1, but not enough to fight about it since we don't yet have any validation beyond somebody eyeballing it.

packages/amplify-graphql-directives/src/directives/v1/connection.ts
dpilch: Thinking about this now again I think `ConnectionDirectiveV1` would be better. There isn't really a drawback to using the `V1` suffix.

palpatim (approved):

atierian (approved):

PR #2345, opened 2024-03-13
feat: Fetch SNS topic ARN from SQL manifest

phani-srikar:
#### Description of changes

This change removes the hard-coded account number of the SQL Lambda Pipeline that is used to construct the SNS Topic ARN, subscribed to by the SQL lambda patcher.

The alternative approach that is implemented here is to fetch the service vended SNS Topic ARN per region from the SQL manifest bucket using a custom resource that is deployed to the customer's account for Gen2 deployments. 
For Gen1 deployments, we read it from the manifest bucket directly during the build step since Async operations are allowed. The necessary backend changes are already deployed. This approach is no different from the one we use fetch the SQL Lambda layer mapping from the manifest bucket and was already discussed with the team.

**Related API changes:** The API changes supporting this will be very similar to the `LayerMapping` wordings since we took the same approach as fetching the Layer ARN we already had in place for Gen1.

[Full E2E passed](https://tiny.amazon.com/1fyaruowm/IsenLink)

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-e2e-tests/src/rds-v2-tests-common/rds-model-v2.ts
+++ b/packages/amplify-e2e-tests/src/rds-v2-tests-common/rds-model-v2.ts
@@ -18,7 +18,6 @@ import { existsSync, readFileSync } from 'fs-extra';
 import generator from 'generate-password';
 import { ObjectTypeDefinitionNode, parse } from 'graphql';
 import AWSAppSyncClient, { AUTH_TYPE } from 'aws-appsync';
-import { ResourceConstants } from 'graphql-transformer-common';
 import gql from 'graphql-tag';
 import {
   ImportedRDSType,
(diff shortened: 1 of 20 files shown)


dpilch (dismissed):

AaronZyLee (approved):
I am not quite familiar with the layer patching related e2e tests. I see you mentioned about the e2e validation. Are there any e2e tests covering this fetching process(especially the custom resource one in CDK workflow)?

packages/amplify-e2e-tests/src/rds-v2-tests-common/rds-model-v2.ts
```diff
@@ -154,7 +145,6 @@ export const testRDSModel = (engine: ImportedRDSType, queries: string[]): void =
       expect(rdsPatchingSubscription.Properties.Protocol).toEqual('lambda');
       expect(rdsPatchingSubscription.Properties.Endpoint).toBeDefined();
       expect(rdsPatchingSubscription.Properties.TopicArn).toBeDefined();
-      expect(rdsPatchingSubscription.Properties.TopicArn).toMatchObject(expectedTopicArn);
```
sundersc, line 157: Can we at least do a partial match here?

phani-srikar: Added a check for the value here.

phani-srikar:
> Are there any e2e tests covering this fetching process(especially the custom resource one in CDK workflow)?

Yes, the existing `sql-models` test success means that we're able to successfully establish a subscription from the SQL patching lambda to the SNS topic who ARN is fetched via a custom resource during deployment.

sundersc (dismissed):

AaronZyLee (dismissed):

phani-srikar:
I had to rebase this with `main` to address conflicts in CDK construct's `.jsii`, no other changes.

sundersc (dismissed):

dpilch (approved):

sundersc (approved):

svidgen (approved):

PR #2362, opened 2024-03-15
feat(graphql-relational-transformer): add gsi creation for ddb references queries

atierian:
#### Description of changes

- Adds the `updateTableForReferencesConnection` method, which will be invoked in `generateResolvers` in the incoming `HasManyDirectiveDDBReferencesTransformer` (subsequent PR). 

- Makes an initial pass at consolidating common parts of `updateTableForConnection` (fields) and `updateTableForReferencesConnection`.

There are few TODOs included with this change that will be resolved in later PRs.
Once we have a commit in `feat/ddb-references`, I'll open a draft PR against main and start tracking these as open items in feature branch PR description.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-graphql-relational-transformer/src/resolvers.ts
+++ b/packages/amplify-graphql-relational-transformer/src/resolvers.ts
@@ -31,7 +31,7 @@ export const updateTableForReferencesConnection = (
     // Ideally validate this further up the chain.
   }
 
-  if (references.length < 1) {
+  if (references.length < 1 || referenceNodes.length < 1) {
     throw new Error('references should not be empty here'); // TODO: better error message
   }
 
--- a/packages/amplify-graphql-relational-transformer/src/resolvers.ts
+++ b/packages/amplify-graphql-relational-transformer/src/resolvers.ts
@@ -27,8 +27,8 @@ export const updateTableForReferencesConnection = (
   const indexName = `gsi-${mappedObjectName}.${field.name.value}`;
   config.indexName = indexName;
 
-  const table = getTable(ctx, relatedType);
-  const gsis = table.globalSecondaryIndexes;
+  const relatedTable = getTable(ctx, relatedType);
+  const gsis = relatedTable.globalSecondaryIndexes;
   if (gsis.some((gsi: any) => gsi.indexName === indexName)) {
     // TODO: In the existing `fields` based implementation, this returns.
     // However, this is likely a schema misconfiguration in the `references`
@@ -48,7 +48,7 @@ export const updateTableForReferencesConnection = (
     ? getConnectedSortKeyAttributeDefinitionsForImplicitHasManyObject(ctx, object, field, referenceNodes.slice(1))
     : undefined;
 
-  addGlobalSecondaryIndex(table, {
+  addGlobalSecondaryIndex(relatedTable, {
     indexName: indexName,
     partitionKey: { name: partitionKeyName, type: partitionKeyType },
     sortKey: sortKey,
--- a/packages/amplify-graphql-relational-transformer/src/resolvers.ts
+++ b/packages/amplify-graphql-relational-transformer/src/resolvers.ts
@@ -8,6 +8,18 @@ import { getSortKeyFields } from './schema';
 import { HasManyDirectiveConfiguration } from './types';
 import { getConnectionAttributeName, getObjectPrimaryKey } from './utils';
 
+/**
+ * Creates a GSI on the table of the `relatedType` based on the config's `references` / `referenceNodes`
+ *
+ * @remarks
+ * This method sets the `indexName` property of the `config` to the GSI name created on the
+ * table of the `relatedType`
+ *
+ * Preconditions: `config.references >= 1` and `config.referenceNodes >= 1`
+ *
+ * @param config The `HasManyDirectiveConfiguration` for DDB references.
+ * @param ctx The `TransformerContextProvider` for DDB references.
+ */
 export const updateTableForReferencesConnection = (
   config: HasManyDirectiveConfiguration, // TODO: Add support for HasOneDirectiveConfiguration
   ctx: TransformerContextProvider,
--- a/packages/amplify-graphql-relational-transformer/src/resolvers.ts
+++ b/packages/amplify-graphql-relational-transformer/src/resolvers.ts
@@ -12,7 +12,7 @@ export const updateTableForReferencesConnection = (
   config: HasManyDirectiveConfiguration, // TODO: Add support for HasOneDirectiveConfiguration
   ctx: TransformerContextProvider,
 ): void => {
-  const { field, fieldNodes, indexName: incomingIndexName, object, references, relatedType } = config;
+  const { field, referenceNodes, indexName: incomingIndexName, object, references, relatedType } = config;
 
   if (incomingIndexName) {
     // TODO: log warning or throw that indexName isn't supported for DDB references
@@ -36,16 +36,16 @@ export const updateTableForReferencesConnection = (
     return;
   }
 
-  const fieldNode = fieldNodes[0];
-  const partitionKeyName = fieldNode.name.value;
+  const referenceNode = referenceNodes[0];
+  const partitionKeyName = referenceNode.name.value;
   // Grabbing the type of the related field.
   // TODO: Validate types of related field and primary's pk match
   // -- ideally further up the chain
-  const partitionKeyType = attributeTypeFromType(fieldNode.type, ctx);
+  const partitionKeyType = attributeTypeFromType(referenceNode.type, ctx);
   const respectPrimaryKeyAttributesOnConnectionField: boolean = ctx.transformParameters.respectPrimaryKeyAttributesOnConnectionField;
 
   const sortKey = respectPrimaryKeyAttributesOnConnectionField
-    ? getConnectedSortKeyAttributeDefinitionsForImplicitHasManyObject(ctx, object, field, fieldNodes.slice(1))
+    ? getConnectedSortKeyAttributeDefinitionsForImplicitHasManyObject(ctx, object, field, referenceNodes.slice(1))
     : undefined;
 
   addGlobalSecondaryIndex(table, {
--- a/packages/amplify-graphql-relational-transformer/src/resolvers.ts
+++ b/packages/amplify-graphql-relational-transformer/src/resolvers.ts
@@ -10,7 +10,7 @@ import { getConnectionAttributeName, getObjectPrimaryKey } from './utils';
 
 export const updateTableForReferencesConnection = (
   config: HasManyDirectiveConfiguration, // TODO: Add support for HasOneDirectiveConfiguration
-  ctx: TransformerContextProvider
+  ctx: TransformerContextProvider,
 ): void => {
   const { field, fieldNodes, indexName: incomingIndexName, object, references, relatedType } = config;
 
@@ -20,9 +20,7 @@ export const updateTableForReferencesConnection = (
   }
 
   if (references.length < 1) {
-    throw new Error( // TODO: better error message
-      'references should not be empty here'
-    )
+    throw new Error('references should not be empty here'); // TODO: better error message
   }
 
   const mappedObjectName = ctx.resourceHelper.getModelNameMapping(object.name.value);
@@ -38,17 +36,17 @@ export const updateTableForReferencesConnection = (
     return;
   }
 
-  const fieldNode = fieldNodes[0]
+  const fieldNode = fieldNodes[0];
   const partitionKeyName = fieldNode.name.value;
   // Grabbing the type of the related field.
   // TODO: Validate types of related field and primary's pk match
   // -- ideally further up the chain
-  const partitionKeyType = attributeTypeFromType(fieldNode.type, ctx)
+  const partitionKeyType = attributeTypeFromType(fieldNode.type, ctx);
   const respectPrimaryKeyAttributesOnConnectionField: boolean = ctx.transformParameters.respectPrimaryKeyAttributesOnConnectionField;
 
   const sortKey = respectPrimaryKeyAttributesOnConnectionField
-  ? getConnectedSortKeyAttributeDefinitionsForImplicitHasManyObject(ctx, object, field, fieldNodes.slice(1))
-  : undefined;
+    ? getConnectedSortKeyAttributeDefinitionsForImplicitHasManyObject(ctx, object, field, fieldNodes.slice(1))
+    : undefined;
 
   addGlobalSecondaryIndex(table, {
     indexName: indexName,
@@ -57,7 +55,7 @@ export const updateTableForReferencesConnection = (
     ctx: ctx,
     relatedTypeName: relatedType.name.value,
   });
-}
+};
 
 /**
  * adds GSI to the table if it doesn't already exists for connection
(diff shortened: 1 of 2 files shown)


packages/amplify-graphql-relational-transformer/src/resolvers.ts
```diff
@@ -8,6 +8,57 @@ import { getSortKeyFields } from './schema';
...
+    throw new Error( // TODO: better error message
+      'references should not be empty here'
+    )
+  }
+
+  const mappedObjectName = ctx.resourceHelper.getModelNameMapping(object.name.value);
+  const indexName = `gsi-${mappedObjectName}.${field.name.value}`;
+  config.indexName = indexName;
+
+  const table = getTable(ctx, relatedType);
```
palpatim, line 32: naming nit
```suggestion
  const relatedTable = getTable(ctx, relatedType);
```

packages/amplify-graphql-relational-transformer/src/resolvers.ts
```diff
@@ -8,6 +8,57 @@ import { getSortKeyFields } from './schema';
 import { HasManyDirectiveConfiguration } from './types';
 import { getConnectionAttributeName, getObjectPrimaryKey } from './utils';
 
+export const updateTableForReferencesConnection = (
```
palpatim, line 11: Lots of notes through this method. I think my confusion stems from my assumptions about the configuration of the `hasMany` directive coming into this method. Right now it's opaque what each of the members means.

Given a schema like

```graphql
type Primary @model { 
  content: String
  related: Related @hasMany(references: 'primaryId')
}
type Related @model {
  content: String
  primaryId: ID!
  primary: Primary @belongsTo(references: 'primaryId')
}
```

...my assumptions (which I partially validated by reading through related code, but can't claim they're exhaustive) are:

- `field` - the GraphQL FieldDefinitionNode on which the `hasMany` directive is declared (i.e., `Primary.related`)
- `fieldNodes` - the GraphQL FieldDefinitionNodes for the fields referenced by the `fields` attribute (invalid for references-style relationships)
- `indexName` - the customer-specified index name to use for the relationship (invalid for references-style relationships)
- `object` - the GraphQL ObjectTypeDefinitionNode of the **parent object** of the field on which the `hasMany` directive is declared (i.e., `Primary`)
- `references` - a list of strings: the value of the `references` attribute
- `relatedType` - the GraphQL ObjectTypeDefinitionNode of the type of the field on which the `hasMany` directive is declared (i.e., `Related`)

Further, I'm assuming the intent of this method is to create a GSI on the related table so we can resolve hasMany and belongsTo relationships by querying that GSI.

If those are accurate, then I don't follow that the logic in the method is creating a properly configured GSI on the related table.

packages/amplify-graphql-relational-transformer/src/resolvers.ts
```diff
@@ -8,6 +8,57 @@ import { getSortKeyFields } from './schema';
...
+    // Ideally validate this further up the chain.
+  }
+
+  if (references.length < 1) {
+    throw new Error( // TODO: better error message
+      'references should not be empty here'
+    )
+  }
+
+  const mappedObjectName = ctx.resourceHelper.getModelNameMapping(object.name.value);
```
palpatim, line 28: I believe the `getModelNameMapping` is handling models renamed with `mapsTo`, but since we're trying to construct a GSI name for the related table, we should be looking for the name of the related model, not the primary model, right?

packages/amplify-graphql-relational-transformer/src/resolvers.ts
```diff
@@ -42,20 +93,42 @@ export const updateTableForConnection = (config: HasManyDirectiveConfiguration,
...
+};
+
+const addGlobalSecondaryIndex = (
+  table: any,
+  props: {
+    indexName: string,
+    partitionKey: KeyAttributeDefinition,
+    sortKey: KeyAttributeDefinition | undefined,
+    ctx: TransformerContextProvider,
+    relatedTypeName: string,
```
palpatim, line 116: Can you double-check that this is working as expected with GSIs added to the related model? It looks like the `relatedTypeName` is used to override the index name in CFN, but I'm a bit fuzzy on why we need to. And since we're already creating the index properly I'm unclear what the correct behavior should be.

packages/amplify-graphql-relational-transformer/src/resolvers.ts
```diff
@@ -8,6 +8,57 @@ import { getSortKeyFields } from './schema';
 import { HasManyDirectiveConfiguration } from './types';
 import { getConnectionAttributeName, getObjectPrimaryKey } from './utils';
 
+export const updateTableForReferencesConnection = (
+  config: HasManyDirectiveConfiguration, // TODO: Add support for HasOneDirectiveConfiguration
```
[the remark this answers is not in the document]
atierian: Only supporting one relationship at a time. This method is invoked for each applicable (DDB references) relational directive defined on the field. The `TODO` essentially means:
```diff
- config: HasManyDirectiveConfiguration
+ config: HasManyDirectiveConfiguration | HasOneDirectiveConfiguration
```
and adding necessary properties to `HasOneDirectiveConfiguration`.

I believe that addresses your question.

packages/amplify-graphql-relational-transformer/src/resolvers.ts
atierian: > but since we're trying to construct a GSI name for the related table

Yes, this is the case.

> we should be looking for the name of the related model, not the primary model, right?

No, I believe it's correct as-is -- we should be using the `<Primary>.<Primary.FieldName>`. I'll break down what's happening here with an example.

Given the following schema.
```graphql
type Team @model {
  id: ID!
  name: String!
  members: [Member] @hasMany(references: ["teamID"])
}

type Member @model {
  id: ID!
  teamID: ID!
  team: Team @belongsTo(references: ["teamID"])
}
```

`mappedObjectName` is `Team` (Primary model) as you mentioned. This leads to a GSI with the name `gsi-Team.members`.

The `gsi-Team.members` GSI is used by the AppSync resolver when querying `Team` and including `members` in the selection set. For example:
```graphql
# Request
query Foo {
  getTeam(id: "001") {
    id
    name
    members { # invokes field resolver, querying `Member` table for `teamID` using `gsi-Team.members`
      items {
        id
        teamID
      }
    }
  }
}

# Response
{
  "data": {
    "getTeam": {
      "id": "001",
      "name": "Inter Miami",
      "members": {
        "items": [
          {
            "id": "12345678-1234-5678-1234-567812345678",
            "name": "Lionel Messi",
            "teamID": "001"
          }
        ]
      }
    }
  }
}
```

This GSI naming convention here makes sense considering it's describing the access pattern for querying the relateds (`members` in this example) through the primary (`Team` in this example). Additionally it's the naming convention used by DDB `fields` based relationships today, so unless it were egregiously incorrect, I'd be reluctant to change it.

packages/amplify-graphql-relational-transformer/src/resolvers.ts
atierian: Good call, addressed in 0d3bce66584dd652cabba1ecc7987cdae3de3154

packages/amplify-graphql-relational-transformer/src/resolvers.ts
```diff
@@ -8,6 +8,57 @@ import { getSortKeyFields } from './schema';
...
+
+  if (references.length < 1) {
+    throw new Error( // TODO: better error message
+      'references should not be empty here'
+    )
+  }
+
+  const mappedObjectName = ctx.resourceHelper.getModelNameMapping(object.name.value);
+  const indexName = `gsi-${mappedObjectName}.${field.name.value}`;
+  config.indexName = indexName;
```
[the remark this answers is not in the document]
atierian: Addressed in c8367fff6fb33eae099c58155985b797380edf86
Will add the `updateTableForConnection` docs in a separate PR to main (unless we end up consolidating the methods in a subsequent `feat/ddb-references` PR).

packages/amplify-graphql-relational-transformer/src/resolvers.ts
```diff
@@ -8,6 +8,57 @@ import { getSortKeyFields } from './schema';
...
+
+  const table = getTable(ctx, relatedType);
+  const gsis = table.globalSecondaryIndexes;
+  if (gsis.some((gsi: any) => gsi.indexName === indexName)) {
```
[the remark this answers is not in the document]
atierian: That's a good callout. For DDB fields, we currently just `return`, which is seems not ideal -- deploy would succeed(?) but queries likely to fail at runtime.

The safe moves I see are:
1. Throw and make the customer rename their GSI
2. Append / prepend something (UUID / `-01` /  `Amplify-` etc) to the GSI name.

I'm going to digest this a bit and come back to it in a subsequent PR -- I've added it in **Open Items** of the feature branch PR.

packages/amplify-graphql-relational-transformer/src/resolvers.ts
```diff
@@ -8,6 +8,57 @@ import { getSortKeyFields } from './schema';
...
+    // TODO: In the existing `fields` based implementation, this returns.
+    // However, this is likely a schema misconfiguration in the `references`
+    // world because we don't support specifying indexName.
+    return;
+  }
+
+  const fieldNode = fieldNodes[0]
```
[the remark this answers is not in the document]
atierian: > I think you're intending to find the primary key of the primary model?

No, this the `FieldDefinitionNode` representing the name and type information of the first`references` argument on the related type. 

For instance, take the following schema:
```graphql
type Team @model {
  id: ID!
  name: String!
  members: [Member] @hasMany(references: ["teamID"])
}

type Member @model {
  id: ID!
  teamID: ID!
  team: Team @belongsTo(references: ["teamID"])
}
```

The `fieldNode` (now named `referenceNode`) is:
```js
{
  kind: "FieldDefinition",
  description: undefined,
  name: {
    kind: "Name",
    value: "teamID",
    loc: { /* ... */ },
  },
  arguments: [],
  type: {
    kind: "NonNullType",
    type: {
      kind: "NamedType",
      name: {
        kind: "Name",
        value: "ID",
        loc: { /* ... */ },
      },
      loc: { /* ... */ },
    },
    loc: { /* ... */ },
  },
  directives: [],
  loc: { /* ... */ },
}
```

> 1/ Confirm how we know that fieldNodes[0] is guaranteed to be the pk, especially in models where PK is generated implicitly by appending to the existing field list.

I believe this question isn't relevant given the information above.

> 2/ adjust the naming of the variable

Good call. Renamed to `referenceNode` / `referenceNodes` for clarity in 649740ce6d6351b78a5f02e9f0ffcf2134229cdb

> 3/ add logic (or a todo) to handle composite PKs. Ideally we'd have a method that could take an model's GraphQL ObjectTypeDefinition node and return the PK/composite PK.

Similar to 1/ I believe this request isn't relevant given this isn't the PK.

> Make sure to run `yarn prettier --write`.

Thanks, addressed in fa9012d3794fce0eab7cd666ea5a6761fcd5e221

packages/amplify-graphql-relational-transformer/src/resolvers.ts
```diff
@@ -8,6 +8,57 @@ import { getSortKeyFields } from './schema';
...
+  const partitionKeyName = fieldNode.name.value;
```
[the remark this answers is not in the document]
atierian: > Shouldn't this be the name of the referenced field

Yes, that's exactly what it is.

packages/amplify-graphql-relational-transformer/src/resolvers.ts
atierian: `relatedTypeName` is being used within `overrideIndexAtCfnLevel` to determine the data source strategy for that model, which is then used to determine how to override the index. It's not being used to override the index name in CFN.

This existing comment may provide some clarity.
https://github.com/aws-amplify/amplify-category-api/pull/2362/files/0eeb2cf93594f7d9ee6e872d5244a5274edb1719#diff-7e54ffdc31c20805d81c17a3e63863a1553b9911a6a00daf1e21f027f2995d84L65-L67
```
  // At the L2 level, the CDK does not handle the way Amplify sets GSI read and write capacity
  // very well. At the L1 level, the CDK does not create the correct IAM policy for accessing the
  // GSI. To get around these issues, keep the L1 and L2 GSI list in sync.
```

packages/amplify-graphql-relational-transformer/src/resolvers.ts
```diff
@@ -8,6 +8,67 @@ import { getSortKeyFields } from './schema';
...
+  const referenceNode = referenceNodes[0];
+  const partitionKeyName = referenceNode.name.value;
+  // Grabbing the type of the related field.
+  // TODO: Validate types of related field and primary's pk match
+  // -- ideally further up the chain
+  const partitionKeyType = attributeTypeFromType(referenceNode.type, ctx);
+  const respectPrimaryKeyAttributesOnConnectionField: boolean = ctx.transformParameters.respectPrimaryKeyAttributesOnConnectionField;
+
+  const sortKey = respectPrimaryKeyAttributesOnConnectionField
+    ? getConnectedSortKeyAttributeDefinitionsForImplicitHasManyObject(ctx, object, field, referenceNodes.slice(1))
```
AaronZyLee, line 60: It seems that this method is not used correctly by the new ref relation. If you compare the last argument for both `ref` and `field`, the former passes the `referenceNodes.slice(1)` which is the related model fields, whereas the latter passes `getSortKeyFields(ctx, object)` which are the sort key fields of the primary. And [these lines of code](https://github.com/aws-amplify/amplify-category-api/pull/2362/files#diff-7e54ffdc31c20805d81c17a3e63863a1553b9911a6a00daf1e21f027f2995d84L98-L100) stays the same. I wonder if the sort key definition is expected.

AaronZyLee: We should be careful to reuse some methods which were intended for the `field` usage, such as this one. It will also be confusing to reuse it in the `ref` scene as there is not `implicitHasMany` concept here. I would suggest to have a new method instead of reusing the old one. The essence of this method is to get the sort key definition for the auto-generated index in related model.

AaronZyLee (commented):
Please add unit tests or mark them as TODOs to be finished in separate PRs

packages/amplify-graphql-relational-transformer/src/resolvers.ts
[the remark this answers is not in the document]
atierian: These types are already being used to represent both. You'll notice that the `references` property already exists to support SQL relationships. Inline code comments added in e27aae657e31ee19522f24832e65616b410c4df2. Thanks for the suggestion.

packages/amplify-graphql-relational-transformer/src/resolvers.ts
```diff
@@ -8,6 +8,67 @@ import { getSortKeyFields } from './schema';
...
+  const relatedTable = getTable(ctx, relatedType);
+  const gsis = relatedTable.globalSecondaryIndexes;
+  if (gsis.some((gsi: any) => gsi.indexName === indexName)) {
+    // TODO: In the existing `fields` based implementation, this returns.
+    // However, this is likely a schema misconfiguration in the `references`
+    // world because we don't support specifying indexName.
+    return;
+  }
+
+  const referenceNode = referenceNodes[0];
```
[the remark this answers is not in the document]
atierian: Good call, thanks! Addressed in e3efa51138b1057d93b0bd26cbb3256dad138307

packages/amplify-graphql-relational-transformer/src/resolvers.ts
atierian: You're absolutely right, great catch! I'm pushing that to a subsequent PR to keep the scope of this one as tight as possible. Removed and added a TODO in 5debc1a8e1c8e15b8f5aaaba341adf3ecfd959ac

atierian:
> Please add unit tests or mark them as TODOs to be finished in separate PRs

Tracking as an open item in the [feature branch PR](https://github.com/aws-amplify/amplify-category-api/pull/2365).

AaronZyLee (approved):

packages/amplify-graphql-relational-transformer/src/resolvers.ts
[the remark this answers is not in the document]
palpatim: FWIW, I vote for option 1.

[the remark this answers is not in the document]
atierian: Went with option 1

PR #2343, opened 2024-03-13
chore(api): upgrade cdk to 2.129.0

sundersc:
Upgrade CDK dependency from 2.80.0 to 2.129.0.

Ran full E2E pipeline and made sure all the E2E tests pass.

PR #2377, opened 2024-03-22
chore: upgrade CDK version to `2.129.0`

phani-srikar:
#### Description of changes

- Upgrades the CDK version used to `2.129.0` that we have been using to run our CDK tests. We have [previously determined](https://github.com/aws-amplify/amplify-category-api/pull/2343) that this version adds least migration friction for us and includes the custom lambda runtime upgrades that we wanted to consume.
- Fixes [an issue](https://github.com/aws-amplify/amplify-category-api/issues/2299) of missing `ecs:TagResource` permission during REST API deployment for API plugin Gen-1 customers. This is identical to the [fix in ECS's CDK construct](https://github.com/aws/aws-cdk/pull/29401/files#diff-df545f4b38d5b880dbaeb12b64f532e8f6bf6e88e48501bc88181f5ab8a3f1b2R1013).

#### Issue #, if available
https://github.com/aws-amplify/amplify-category-api/issues/2299

PR and E2E runs successfully.
[First E2E run](https://tiny.amazon.com/1gf4m7ivn/IsenLink)
[Second run with only one failed test](https://tiny.amazon.com/djacm0vc/IsenLink)

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/package.json
+++ b/package.json
@@ -235,7 +235,7 @@
     "cross-fetch": "^2.2.6",
     "trim-newlines": "^3.0.1",
     "aws-sdk": "^2.1141.0",
-    "aws-cdk-lib": "2.97.0",
+    "aws-cdk-lib": "2.129.0",
     "parse-path": "^5.0.0",
     "parse-url": "^8.1.0",
     "moment-timezone": "^0.5.35",
--- a/package.json
+++ b/package.json
@@ -235,6 +235,7 @@
     "cross-fetch": "^2.2.6",
     "trim-newlines": "^3.0.1",
     "aws-sdk": "^2.1141.0",
+    "aws-cdk-lib": "2.97.0",
     "parse-path": "^5.0.0",
     "parse-url": "^8.1.0",
     "moment-timezone": "^0.5.35",
--- a/packages/amplify-category-api/package.json
+++ b/packages/amplify-category-api/package.json
@@ -75,8 +75,8 @@
     "@aws-amplify/amplify-provider-awscloudformation": "^8.8.0",
     "amplify-headless-interface": "^1.17.4",
     "amplify-util-headless-input": "^1.9.15",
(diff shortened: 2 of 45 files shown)


package.json
```diff
@@ -235,6 +235,7 @@
     "cross-fetch": "^2.2.6",
     "trim-newlines": "^3.0.1",
     "aws-sdk": "^2.1141.0",
+    "aws-cdk-lib": "2.129.0",
```
phani-srikar, line 238: >What is the purpose of this declaration, and why is it pinned rather than compatible?

This is to ensure we are resolving to the updated version of the library in our `yarn.lock` even if some of the downstream dependencies like the Gen-1 `amplify-cli-core` are still pulling in the older version. I believe we can remove this override after CLI is upgraded too.

packages/amplify-category-api/package.json
```diff
@@ -75,8 +75,8 @@
     "@aws-amplify/amplify-provider-awscloudformation": "^8.8.0",
     "amplify-headless-interface": "^1.17.4",
     "amplify-util-headless-input": "^1.9.15",
-    "aws-cdk-lib": "^2.80.0",
-    "constructs": "^10.0.5"
+    "aws-cdk-lib": "^2.129.0",
```
phani-srikar, line 78: > I see we're declaring major version compatible here, but pinning at the top level. Is that going to cause conflicts?

All the `peerDependency` declarations were/are specified as major version compatible. What this essentially means is that the host i.e CLI can choose to install any `2.x` where `x>=129`. I don't have full context on why we were allowing for that flexibility earlier, but would prefer to keep it the same.

palpatim:
Closing in favor of https://github.com/aws-amplify/amplify-category-api/pull/2515

PR #2380, opened 2024-03-25
feat(graphql-relational-transformer): add ddb references support for hasOne directive

atierian:
#### Description of changes
Adds `references` support for DDB data sources when defining `@hasOne` relationships.
- Adds `HasOneDirectiveDDBReferencesTransformer`
- Fixes input types in `RelationalResolverGenerator` abstract class and some method signatures.
- Adds some logic in the `getHasOneDirectiveTransformer` to unlock use references implementation for `hasOne`.
- Adds `hasOne` resolver implementation.
- Adds (start of) unit test in `amplify-graphql-has-one-references-transformer.test.ts`

None

#### Issue #, if available
N/A

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-graphql-relational-transformer/src/has-one/has-one-directive-ddb-references-transformer.ts
+++ b/packages/amplify-graphql-relational-transformer/src/has-one/has-one-directive-ddb-references-transformer.ts
@@ -1,0 +1,47 @@
+import { DDB_DB_TYPE } from '@aws-amplify/graphql-transformer-core';
+import {
+  TransformerContextProvider,
+  TransformerPrepareStepContextProvider,
+  TransformerTransformSchemaStepContextProvider,
+} from '@aws-amplify/graphql-transformer-interfaces';
+import { DataSourceBasedDirectiveTransformer } from '../data-source-based-directive-transformer';
+import { DDBRelationalReferencesResolverGenerator } from '../resolver/ddb-references-generator';
+import { setFieldMappingResolverReference, updateTableForReferencesConnection } from '../resolvers';
+import { HasOneDirectiveConfiguration } from '../types';
+import { ensureReferencesArray, getReferencesNodes, registerHasOneForeignKeyMappings, validateParentReferencesFields } from '../utils';
+
+/**
+ * HasOneDirectiveDDBFieldsTransformer executes transformations based on `@hasOne(references: [String!])` configurations
+ * and surrounding TransformerContextProviders for DynamoDB data sources.
+ *
+ * This should not be used for `@hasOne(fields: [String!])` definitions.
+ */
+export class HasOneDirectiveDDBReferencesTransformer implements DataSourceBasedDirectiveTransformer<HasOneDirectiveConfiguration> {
+  dbType = DDB_DB_TYPE;
+
+  prepare = (context: TransformerPrepareStepContextProvider, config: HasOneDirectiveConfiguration): void => {
+    const modelName = config.object.name.value;
+    setFieldMappingResolverReference(context, config.relatedType?.name?.value, modelName, config.field.name.value, true);
+    registerHasOneForeignKeyMappings({
+      transformParameters: context.transformParameters,
+      resourceHelper: context.resourceHelper,
+      thisTypeName: modelName,
+      thisFieldName: config.field.name.value,
+      relatedType: config.relatedType,
+    });
+  };
+
+  transformSchema = (context: TransformerTransformSchemaStepContextProvider, config: HasOneDirectiveConfiguration): void => {
+    validateParentReferencesFields(config, context as TransformerContextProvider);
+  };
+
+  generateResolvers = (context: TransformerContextProvider, config: HasOneDirectiveConfiguration): void => {
+    updateTableForReferencesConnection(config, context);
+    new DDBRelationalReferencesResolverGenerator().makeHasOneGetItemConnectionWithKeyResolver(config, context);
+  };
+
+  validate = (context: TransformerContextProvider, config: HasOneDirectiveConfiguration): void => {
+    ensureReferencesArray(config);
+    config.referenceNodes = getReferencesNodes(config, context);
+  };
+}
--- a/packages/amplify-graphql-relational-transformer/src/has-one/has-one-directive-transformer-factory.ts
+++ b/packages/amplify-graphql-relational-transformer/src/has-one/has-one-directive-transformer-factory.ts
@@ -3,13 +3,16 @@ import { HasOneDirectiveConfiguration } from '../types';
 import { DataSourceBasedDirectiveTransformer } from '../data-source-based-directive-transformer';
 import { HasOneDirectiveDDBFieldsTransformer } from './has-one-directive-ddb-fields-transformer';
 import { HasOneDirectiveSQLTransformer } from './has-one-directive-sql-transformer';
+import { HasOneDirectiveDDBReferencesTransformer } from './has-one-directive-ddb-references-transformer';
 
 const hasOneDirectiveMySqlTransformer = new HasOneDirectiveSQLTransformer();
 const hasOneDirectivePostgresTransformer = new HasOneDirectiveSQLTransformer();
 const hasOneDirectiveDdbFieldsTransformer = new HasOneDirectiveDDBFieldsTransformer();
+const hasOneDirectiveDdbReferencesTransformer = new HasOneDirectiveDDBReferencesTransformer();
 
 export const getHasOneDirectiveTransformer = (
   dbType: ModelDataSourceStrategyDbType,
+  config: HasOneDirectiveConfiguration,
   // eslint-disable-next-line consistent-return
 ): DataSourceBasedDirectiveTransformer<HasOneDirectiveConfiguration> => {
   switch (dbType) {
@@ -18,6 +21,22 @@ export const getHasOneDirectiveTransformer = (
     case 'POSTGRES':
       return hasOneDirectivePostgresTransformer;
     case 'DYNAMODB':
+      // If references are passed to the directive, we'll use the references relational
+      // modeling approach.
+      if (config.references) {
+        // Passing both references and fields is not supported.
+        if (config.fields) {
+          // TODO: Better error message
+          throw new Error('Something went wrong >> cannot have both references and fields.');
+        }
+        if (config.references.length < 1) {
+          throw new Error(`Invalid @hasMany directive on ${config.field.name.value} - empty references list`);
+        }
+        return hasOneDirectiveDdbReferencesTransformer;
+      }
+
+      // fields based relational modeling is the default because it supports implicit
+      // field creation / doesn't require explicitly defining the fields in the directive.
       return hasOneDirectiveDdbFieldsTransformer;
   }
 };
--- a/packages/amplify-graphql-relational-transformer/src/resolver/ddb-generator.ts
+++ b/packages/amplify-graphql-relational-transformer/src/resolver/ddb-generator.ts
@@ -221,7 +221,15 @@ export class DDBRelationalResolverGenerator extends RelationalResolverGenerator 
    * @param config The connection directive configuration.
    * @param ctx The transformer context provider.
    */
-  makeHasOneGetItemConnectionWithKeyResolver = (
+  makeHasOneGetItemConnectionWithKeyResolver = (config: HasOneDirectiveConfiguration, ctx: TransformerContextProvider): void => {
+    this.makeHasOneBelongToGetItemConnectionWithKeyResolver(config, ctx);
+  };
+
+  makeBelongsToGetItemConnectionWithKeyResolver = (config: BelongsToDirectiveConfiguration, ctx: TransformerContextProvider): void => {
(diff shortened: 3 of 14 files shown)


packages/amplify-graphql-relational-transformer/src/resolver/ddb-generator.ts
```diff
@@ -221,7 +221,15 @@ export class DDBRelationalResolverGenerator extends RelationalResolverGenerator
    * @param config The connection directive configuration.
    * @param ctx The transformer context provider.
    */
-  makeHasOneGetItemConnectionWithKeyResolver = (
+  makeHasOneGetItemConnectionWithKeyResolver = (config: HasOneDirectiveConfiguration, ctx: TransformerContextProvider): void => {
```
atierian, line 224: This was necessary due to the changes made in the `makeHasOneBelongsToGetItemCommectionWithKeyResolver` signature within the `RelationalResolverGenerator` abstract class.

packages/amplify-graphql-relational-transformer/src/resolver/ddb-references-generator.ts
```diff
@@ -165,8 +165,117 @@ export class DDBRelationalReferencesResolverGenerator extends DDBRelationalResol
     ctx.resolvers.addResolver(object.name.value, field.name.value, resolver);
   };
 
-  makeHasOneGetItemConnectionWithKeyResolver = (_config: HasOneDirectiveConfiguration, _ctx: TransformerContextProvider): void => {
-    // TODO: Implement hasOne resolver -- this should be nearly identical to hasMany
+  makeHasOneGetItemConnectionWithKeyResolver = (config: HasOneDirectiveConfiguration, ctx: TransformerContextProvider): void => {
```
atierian, line 168: There's a lot of unnecessary duplicate code introduced here. The intent is to define the working implementation as a point of reference, get it in the feature branch, then follow up with improvements. Tracking this work in the feature branch PR description as:
>  Consolidate resolver implementations for ddb references to reduce duplicate code.

packages/amplify-graphql-relational-transformer/src/resolver/generator.ts
```diff
@@ -3,9 +3,6 @@ import { BelongsToDirectiveConfiguration, HasManyDirectiveConfiguration, HasOneD
 
 export abstract class RelationalResolverGenerator {
   abstract makeHasManyGetItemsConnectionWithKeyResolver(config: HasManyDirectiveConfiguration, ctx: TransformerContextProvider): void;
-  abstract makeHasOneGetItemConnectionWithKeyResolver(
-    config: HasOneDirectiveConfiguration | BelongsToDirectiveConfiguration,
-    ctx: TransformerContextProvider,
-  ): void;
-  abstract makeBelongsToGetItemConnectionWithKeyResolver(config: HasOneDirectiveConfiguration, ctx: TransformerContextProvider): void;
+  abstract makeHasOneGetItemConnectionWithKeyResolver(config: HasOneDirectiveConfiguration, ctx: TransformerContextProvider): void;
```
atierian, line 6: Changing the signature was necessary to prevent type issues in the DDB references resolver generation in `DDBRelationalReferencesResolverGenerator`. Independent from that, this change seems like the right thing to do.

packages/amplify-graphql-relational-transformer/src/utils.ts
```diff
@@ -193,6 +193,31 @@ export const ensureReferencesArray = (
...
+      - field.type.name.value == config.type
+
+
+  */
+  // const related = config.relatedType;
+  // const otherSide = related.fields?.find((field) => {
+  //   const hasDirective = field?.directives?.length >= 1
+  // })
+};
+
```
atierian, line 220: This is a shell of an implementation. Tracking pending work in feature branch PR description as:
> Add validation for relational bi-directionality of DDB references

packages/amplify-graphql-relational-transformer/src/utils.ts
```diff
@@ -254,6 +279,7 @@ export const getReferencesNodes = (
     const fieldNode = relatedType.fields!.find((field) => field.name.value === fieldName);
 
     if (!fieldNode) {
+      // TODO: include more directive / type information to help debugging.
```
atierian, line 282: After going through the error message assertions in our unit tests, there's a lot of room for improvement.
Tracking this work in feature branch PR as:
> Improve error messages surfaced by including more directive and type information to improve debugging.

packages/amplify-graphql-relational-transformer/src/utils.ts
```diff
@@ -193,6 +193,27 @@ export const ensureReferencesArray = (
   }
 };
 
+export const ensureReferencesBidirectionality = (
```
atierian, line 196: This is a shell of an implementation. Tracking pending work in feature branch PR description as:
> Add validation for relational bi-directionality of DDB references

packages/amplify-graphql-relational-transformer/src/resolver/ddb-references-generator.ts
```diff
@@ -165,8 +165,116 @@ export class DDBRelationalReferencesResolverGenerator extends DDBRelationalResol
...
+      MappingTemplate.s3MappingTemplateFromString(
+        print(
+          DynamoDBMappingTemplate.dynamoDBResponse(
+            false,
+            ifElse(
+              and([not(ref('ctx.result.items.isEmpty()')), equals(ref('ctx.result.scannedCount'), int(1))]),
+              toJson(ref('ctx.result.items[0]')),
+              compoundExpression([
+                iff(and([ref('ctx.result.items.isEmpty()'), equals(ref('ctx.result.scannedCount'), int(1))]), ref('util.unauthorized()')),
+                toJson(nul()),
```
AaronZyLee, line 267: For hasOne with reference, it could return the result `scannedCount` greater than `1` (as there might be multiple items by querying the index if more than two children with same parent reference id). Are we still returning `null` for this case in the response? Or is there a plan to prevent the child record to be created? For either case, it is better to have TODO comment here and address it in separate PR

packages/amplify-graphql-relational-transformer/src/has-one/has-one-directive-ddb-references-transformer.ts
```diff
@@ -0,0 +1,47 @@
...
+    validateParentReferencesFields(config, context as TransformerContextProvider);
+  };
+
+  generateResolvers = (context: TransformerContextProvider, config: HasOneDirectiveConfiguration): void => {
+    updateTableForReferencesConnection(config, context);
+    new DDBRelationalReferencesResolverGenerator().makeHasOneGetItemConnectionWithKeyResolver(config, context);
+  };
+
+  validate = (context: TransformerContextProvider, config: HasOneDirectiveConfiguration): void => {
+    ensureReferencesArray(config);
```
AaronZyLee, line 44: This method will throw an error when the `fields` are detected in the config but with the error message `Relational database...`. In this code block, it is related to the `references` usage of a DynamoDB, which is a non-relational database. We should modify/branch the error message to indicate the correct validation rule. (In this case it mostly validates `fields` cannot be used with `references` for Dynamodb)
https://github.com/aws-amplify/amplify-category-api/blob/ed44cd66d504e6d29fe74968d24ae1ce2f665d0a/packages/amplify-graphql-relational-transformer/src/utils.ts#L184

packages/amplify-graphql-relational-transformer/src/schema.ts
```diff
@@ -126,7 +126,10 @@ const ensureModelSortDirectionEnum = (ctx: TransformerContextProvider): void =>
 /**
  * ensureHasOneConnectionField
  */
-export const ensureHasOneConnectionField = (config: HasOneDirectiveConfiguration, ctx: TransformerContextProvider): void => {
+export const ensureHasOneConnectionField = (
+  config: HasOneDirectiveConfiguration | BelongsToDirectiveConfiguration,
```
AaronZyLee, line 130: Why is `BelongsTo..` added here? Is it being used by belongsTo process?

atierian: This is needed because we've added an `indexName` property to the `HasOneDirectiveConfiguration`, which results in the following error:

> Argument of type 'BelongsToDirectiveConfiguration' is not assignable to parameter of type 'HasOneDirectiveConfiguration'.
  Property 'indexName' is missing in type 'BelongsToDirectiveConfiguration' but required in type 'HasOneDirectiveConfiguration'.

in the existing [ensureBelongsToConnectionField](https://github.com/aws-amplify/amplify-category-api/blob/ed44cd66d504e6d29fe74968d24ae1ce2f665d0a/packages/amplify-graphql-relational-transformer/src/schema.ts#L262) method.

packages/amplify-graphql-relational-transformer/src/has-one/has-one-directive-ddb-references-transformer.ts
atierian: Thanks for calling this out. I already have an open item in the feature branch PR to track improving error messages. I've added your a link to your comment there to make sure it's included in that PR.

packages/amplify-graphql-relational-transformer/src/has-one/has-one-directive-transformer-factory.ts
```diff
@@ -18,6 +21,22 @@ export const getHasOneDirectiveTransformer = (
...
+      // modeling approach.
+      if (config.references) {
+        // Passing both references and fields is not supported.
+        if (config.fields) {
+          // TODO: Better error message
+          throw new Error('Something went wrong >> cannot have both references and fields.');
+        }
+        if (config.references.length < 1) {
+          throw new Error(`Invalid @hasMany directive on ${config.field.name.value} - empty references list`);
+        }
```
phani-srikar, line 34: would simplify this nested logic here

packages/amplify-graphql-relational-transformer/src/has-one/has-one-directive-ddb-references-transformer.ts
```diff
@@ -0,0 +1,47 @@
...
+      transformParameters: context.transformParameters,
+      resourceHelper: context.resourceHelper,
+      thisTypeName: modelName,
+      thisFieldName: config.field.name.value,
+      relatedType: config.relatedType,
+    });
+  };
+
+  transformSchema = (context: TransformerTransformSchemaStepContextProvider, config: HasOneDirectiveConfiguration): void => {
+    validateParentReferencesFields(config, context as TransformerContextProvider);
```
phani-srikar, line 35: why do we have validation steps here instead of in the `validate` method/step?

AaronZyLee (approved):

packages/amplify-graphql-relational-transformer/src/has-one/has-one-directive-ddb-references-transformer.ts
atierian: Great question. This was already being called in the `transformSchema` method. [[reference](https://github.com/aws-amplify/amplify-category-api/blob/aa815ff1f6c54a8983c5250d58cfed4e7931aacd/packages/amplify-graphql-relational-transformer/src/graphql-has-one-transformer.ts#L192)] This isn't the only case of validation being done outside of `validate`. I'm tracking this work in the [feature branch PR](https://github.com/aws-amplify/amplify-category-api/pull/2370):
>  Move all necessary validation to `validate` methods in relational directive transformers.

packages/amplify-graphql-relational-transformer/src/has-one/has-one-directive-transformer-factory.ts
atierian: I'm open to suggestions, do you have one?

We have a few states that we need to check to provide applicable error messages:

```mermaid
flowchart

REFARG{`references` arg passed?}
REFARG --> |yes| FIELDSARG
REFARG --> |no| FIELDS[`fields` behavior]
REFEMPTY{`references` arg empty?}
REFEMPTY --> |yes| E1[Error: `references` cannot be empty]
REFEMPTY --> |no| REF[`references` behavior]
FIELDSARG{`fields` arg passed?}
FIELDSARG --> |yes| E2[Error: cannot pass `references` and `fields`]
FIELDSARG --> |no| REFEMPTY
```

atierian: This is one option, but I don't see how it simplifies anything.
```ts
if (config.references && config.fields) { // error }
if (config.references) {
  if (config.references.length > 0) { // references }
  // error 
}
// fields
```
another option:
```ts
if (config.references && config.fields) { // error }
if (config.references?.length > 0) { // references }
// fields
```
would reduce the complexity. But it does so at the cost of making this `@hasOne(references: [])` use `fields` behavior, which we really don't want.

phani-srikar:
```
if (config.references && config.fields) { // error }
if (config.references && config.references.length <= 0) { //error }

if (config.references) { // references }
// fields
```
Is what I had in mind so the error conditions are upfront and easy to read, but this is nit picky and not worth a revision.

phani-srikar (approved):

PR #2398, opened 2024-03-27
feat: iam access support for searchable, index and sql directives

sobolk:
#### Description of changes

This is a follow up on https://github.com/aws-amplify/amplify-category-api/pull/2385 .

It turns out that sandbox logic present in `@aws-amplify/graphql-model-transformer` is also present in `@searchable` `@index` and `@sql` transformers and also requires amendments to allow iam access for types annotated with these directives (that are NOT annotated with `@auth`).

1. for `@searchable` service directive logic and sandbox resolver is amended.
2. for `@index` service directive logic and sandbox resolver is amended.
3. for `@sql` only resolver is amended, there's no logic to add service directives present.

This PR amends `@searchable` `@index` and `@sql` transformers by pattern matching changes from `graphql-model-transformer`.

Note: This PR does not attempt to remove this existing code duplication. That should be accomplished separately if desired.

Added unit tests.

Ran PR checks and e2e tests.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-graphql-index-transformer/src/__tests__/amplify-graphql-index-transformer.test.ts
+++ b/packages/amplify-graphql-index-transformer/src/__tests__/amplify-graphql-index-transformer.test.ts
@@ -1450,6 +1450,113 @@ describe('Index query resolver creation', () => {
       dataSources: {
         get: jest.fn(),
       },
+      synthParameters: {},
     };
   };
+});
+
+describe('auth', () => {
+  const API_KEY = 'API Key Authorization';
+  const IAM_AUTH_TYPE = 'IAM Authorization';
+
+  const schema = /* GraphQL */ `
+    type Test @model {
+      id: ID!
+      description: String @index(name: "index1")
+    }
+  `;
+
+  it('sandbox auth enabled should add apiKey if not default mode of auth', () => {
+    const out = testTransform({
+      schema,
+      transformers: [new ModelTransformer(), new IndexTransformer()],
+      transformParameters: {
+        sandboxModeEnabled: true,
+      },
+      synthParameters: {
+        enableIamAccess: false,
+      },
+      authConfig: {
+        defaultAuthentication: {
+          authenticationType: 'AMAZON_COGNITO_USER_POOLS',
+        },
+        additionalAuthenticationProviders: [
+          {
+            authenticationType: 'API_KEY',
+          },
+        ],
+      },
+    });
+    expect(out).toBeDefined();
+    expect(out.schema).toContain('aws_api_key');
+    expect(out.schema).not.toContain('aws_iam');
+    expect(out.schema).toMatchSnapshot();
+    expect(out.resolvers['Query.testsByDescription.postAuth.1.res.vtl']).toBeDefined();
+    expect(out.resolvers['Query.testsByDescription.postAuth.1.res.vtl']).toContain(API_KEY);
+    expect(out.resolvers['Query.testsByDescription.postAuth.1.res.vtl']).not.toContain(IAM_AUTH_TYPE);
+    expect(out.resolvers['Query.testsByDescription.postAuth.1.res.vtl']).toMatchSnapshot();
+  });
+
+  it('iam auth enabled should add aws_iam if not default mode of auth', () => {
+    const out = testTransform({
+      schema,
+      transformers: [new ModelTransformer(), new IndexTransformer()],
+      transformParameters: {
+        sandboxModeEnabled: false,
+      },
+      synthParameters: {
+        enableIamAccess: true,
+      },
+    });
+    expect(out).toBeDefined();
+    expect(out.schema).not.toContain('aws_api_key');
+    expect(out.schema).toContain('aws_iam');
+    expect(out.schema).toMatchSnapshot();
+    expect(out.resolvers['Query.testsByDescription.postAuth.1.res.vtl']).toBeDefined();
+    expect(out.resolvers['Query.testsByDescription.postAuth.1.res.vtl']).not.toContain(API_KEY);
+    expect(out.resolvers['Query.testsByDescription.postAuth.1.res.vtl']).toContain(IAM_AUTH_TYPE);
+    expect(out.resolvers['Query.testsByDescription.postAuth.1.res.vtl']).toMatchSnapshot();
+  });
+
+  it('iam and sandbox auth enabled should add aws_iam and aws_api_key if not default mode of auth', () => {
+    const out = testTransform({
+      schema,
+      transformers: [new ModelTransformer(), new IndexTransformer()],
+      transformParameters: {
+        sandboxModeEnabled: true,
+      },
+      synthParameters: {
+        enableIamAccess: true,
+      },
+    });
+    expect(out).toBeDefined();
+    expect(out.schema).toContain('aws_api_key');
+    expect(out.schema).toContain('aws_iam');
+    expect(out.schema).toMatchSnapshot();
+    expect(out.resolvers['Query.testsByDescription.postAuth.1.res.vtl']).toBeDefined();
+    expect(out.resolvers['Query.testsByDescription.postAuth.1.res.vtl']).toContain(API_KEY);
+    expect(out.resolvers['Query.testsByDescription.postAuth.1.res.vtl']).toContain(IAM_AUTH_TYPE);
+    expect(out.resolvers['Query.testsByDescription.postAuth.1.res.vtl']).toMatchSnapshot();
+  });
+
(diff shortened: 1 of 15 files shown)


packages/amplify-graphql-searchable-transformer/src/generate-resolver-vtl.ts
```diff
@@ -22,26 +22,42 @@ import {
...
-  let sandboxExp: Expression;
-  if (enabled) {
-    sandboxExp = ifElse(
-      notEquals(methodCall(ref('util.authType')), str(API_KEY)),
-      methodCall(ref('util.unauthorized')),
-      qref(methodCall(ref('ctx.stash.put'), str(allowedAggFieldsList), raw(JSON.stringify(fields)))),
-    );
-  } else {
-    sandboxExp = methodCall(ref('util.unauthorized'));
+export const sandboxMappingTemplate = (
```
palpatim, line 36: same note as above about "sandbox" naming

packages/amplify-graphql-searchable-transformer/src/generate-resolver-vtl.ts
```diff
@@ -22,26 +22,42 @@ import {
...
+  const expressions: Array<Expression> = [];
+  const ifAuthorizedExpression: Expression = compoundExpression([
+    qref(methodCall(ref('ctx.stash.put'), str(allowedAggFieldsList), raw(JSON.stringify(fields)))),
+    ret(toJson(obj({}))),
+  ]);
+  if (isSandboxModeEnabled) {
+    expressions.push(iff(equals(methodCall(ref('util.authType')), str(API_KEY)), ifAuthorizedExpression));
+  }
+  if (genericIamAccessEnabled) {
+    const isNonCognitoIAMPrincipal = and([
```
palpatim, line 50: same note as above about utility method

packages/amplify-graphql-sql-transformer/src/graphql-sql-transformer.ts
```diff
@@ -146,15 +149,26 @@ export class SqlTransformer extends TransformerPluginBase {
   };
 }
 
-const generateAuthExpressionForSandboxMode = (enabled: boolean): string => {
-  let exp;
+const generateAuthExpressionForSandboxMode = (isSandboxModeEnabled: boolean, genericIamAccessEnabled: boolean | undefined): string => {
```
palpatim, line 152: ditto naming & utility method comments, if applicable

packages/amplify-graphql-index-transformer/src/__tests__/amplify-graphql-index-transformer.test.ts
```diff
@@ -1450,6 +1450,113 @@ describe('Index query resolver creation', () => {
...
+    expect(out.schema).toContain('aws_api_key');
+    expect(out.schema).toContain('aws_iam');
+    expect(out.schema).toMatchSnapshot();
+    expect(out.resolvers['Query.testsByDescription.postAuth.1.res.vtl']).toBeDefined();
+    expect(out.resolvers['Query.testsByDescription.postAuth.1.res.vtl']).toContain(API_KEY);
+    expect(out.resolvers['Query.testsByDescription.postAuth.1.res.vtl']).toContain(IAM_AUTH_TYPE);
+    expect(out.resolvers['Query.testsByDescription.postAuth.1.res.vtl']).toMatchSnapshot();
+  });
+
+  it('iam and sandbox auth disable should not add service directives', () => {
```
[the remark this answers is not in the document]
sobolk: It's a prior art that's been on `main` for a while.
See https://github.com/search?q=repo%3Aaws-amplify%2Famplify-category-api+%22if+%28%21hasAuth+%26%26%22&type=code .

If I understand correctly the way this works is:
For types that don't have `@auth` but have either `@model` or `@searchable` or `@index` or `@sql` we do two things:
1. If sandbox is enabled we add necessary service directives to make appsync allow auth type for them
2. We generate `postAuth` resolver which checks for the following things:
    1. It checks if any `auth` resolver ran (auth resolvers leave `ctx.stash.hasAuth` set to true to mark their presence). If any `auth` resolver ran the logic is short circuited and access is granted.
    2. If no `auth` resolver ran (no `@auth` directive on `type`) then depending on sandbox setting it either lets API key access in or not.

One of requirements given was that enabling "generic iam access" should grant access to models without `@auth` as well (@renebrandel please correct me if I'm wrong here). Hence there's a need to amend these "sandbox" resolvers to account for that (otherwise they deny access).
I do agree that it makes naming a bit unfortunate and this should be changed (I'll elaborate more under other comment).

packages/amplify-graphql-index-transformer/src/resolvers/resolvers.ts
```diff
@@ -901,14 +903,26 @@ const generateSyncResolverInit = (): CompoundExpressionNode => {
...
+export const generateAuthExpressionForSandboxMode = (
+  isSandboxModeEnabled: boolean,
+  genericIamAccessEnabled: boolean | undefined,
+): string => {
+  const expressions: Array<Expression> = [];
+  if (isSandboxModeEnabled) {
+    expressions.push(iff(equals(methodCall(ref('util.authType')), str(API_KEY)), ret(toJson(obj({})))));
+  }
+  if (genericIamAccessEnabled) {
+    const isNonCognitoIAMPrincipal = and([
```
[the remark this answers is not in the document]
sobolk: I did. However it's only scoped to `amplify-graphql-auth-transformer` see https://github.com/aws-amplify/amplify-category-api/pull/2331#discussion_r1534758875 .

We could push it one level up and make it shared via `transformer-core` . Which is tradeoff between exposing primitives in public api VS duplication of code.

I consider it not worth polluting `transformer-core` with that detail, so I kept one copy per package.

This sandbox (aka postAuth) support for non-auth types already suffers from duplication (there are at least 4 copies of similar solution). I mentioned in PR description that I'm deliberately avoiding de-duplication - this should be separate dedicated effort. And perhaps there `isNonCognitoIAMPrincipal` should find proper place behind better abstraction.

packages/amplify-graphql-index-transformer/src/resolvers/resolvers.ts
```diff
@@ -901,14 +903,26 @@ const generateSyncResolverInit = (): CompoundExpressionNode => {
...
-  else exp = methodCall(ref('util.unauthorized'));
+export const generateAuthExpressionForSandboxMode = (
+  isSandboxModeEnabled: boolean,
+  genericIamAccessEnabled: boolean | undefined,
+): string => {
+  const expressions: Array<Expression> = [];
+  if (isSandboxModeEnabled) {
+    expressions.push(iff(equals(methodCall(ref('util.authType')), str(API_KEY)), ret(toJson(obj({})))));
+  }
+  if (genericIamAccessEnabled) {
```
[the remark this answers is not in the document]
sobolk: I agree. This should be called `generatePostAuthExpression` or something like that.

And it if we do this we should refactor similar places I touched in previous PR.

My suggestion would be to take this PR as is and do renaming in separate dedicated PR - it's going to be rather big.

palpatim (approved):

packages/amplify-graphql-sql-transformer/src/__tests__/__snapshots__/amplify-graphql-sql-transformer.test.ts.snap
```diff
@@ -1,5 +1,53 @@
...
+"## [Start] Sandbox Mode Disabled. **
+#if( !$ctx.stash.get(\\"hasAuth\\") )
+  $util.unauthorized()
+#end
+$util.toJson({})
+## [End] Sandbox Mode Disabled. **"
+`;
+
+exports[`sql directive tests Post auth resolver Renders post auth resolver if sandbox is disabled and iam access enabled 1`] = `
+"## [Start] Sandbox Mode Disabled. **
```
phani-srikar, line 13: nit: readability comment - This resolver snippet b/w [Start] and [End] depends on both sandbox mode and iam access states. Making this `## [Start] Sandbox Mode Disabled and IAM access Enabled **` would make it more clear.

phani-srikar (approved):

packages/amplify-graphql-sql-transformer/src/__tests__/__snapshots__/amplify-graphql-sql-transformer.test.ts.snap
sobolk: I'll make it part of renaming PR discussed above https://github.com/aws-amplify/amplify-category-api/pull/2398#discussion_r1543340299 .

PR #2406, opened 2024-03-29
feat: allow multiple graphql APIs in a cdk app

dpilch:
#### Description of changes

* Allow multiple GraphQL APIs to be created in a single CDK app.
* This was previously explicitly disallowed to prevent non-helpful error messages.
  * `storeOutput` for Amplify gen 2 stores outputs in CfnOutput resources with non-unique names.
    * Multiple GraphQL APIs can still not share the same parent because of this. If the APIs are in different nested stacks it is fine.
  * Transformer resources are written to temporary directory. The directory was reused for each GraphQL API in the CDK synth resulting in some resources being overwritten.
    * The construct now uses a unique temporary directory for each API

Internal changes only.

#### Issue #, if available

N/A

* Test in sample app
* Unit tests
* E2E tests

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-graphql-api-construct/src/internal/asset-manager.ts
```diff
@@ -11,9 +12,13 @@ const RESOLVER_PREFIX = 'resolvers';
  * with the path method which is used in CDK.
  */
 export class AssetManager {
```
palpatim, line 14: I recognize that this is not an architecture that you're changing as part of this PR, but I don't understand the relationship between the construct's AssetManager and the AssetManager in the transformer core--despite their shared name, they have different interfaces (and presumably, responsibilities). Can you expand the doc comment above with some extra detail:
- What is the responsibility of each of the AssetManager classes?
- What is the relationship, if any, between the two?
- I don't see that the construct AM conforms to the transformer's AssetManagerProvider interface, so I'm unsure how the construct AM logic is actually invoked.

My hope is to come away with a high-level understanding of how the core asset generation logic uses the construct's AM

packages/amplify-graphql-transformer-interfaces/src/graphql-api-provider.ts
```diff
@@ -113,13 +114,13 @@ export interface InlineMappingTemplateProvider {
```
palpatim, line 114: Why doesn't `InlineMappingTemplateProvider` above also need an AssetManagerProvider? Do we not write it as an asset anywhere during processing?

dpilch: The `InlineTemplate` does not store any asset.

https://github.com/aws-amplify/amplify-category-api/blob/a5a233cfc1aa0859a9adf74f319b53c9479b38d4/packages/amplify-graphql-transformer-core/src/cdk-compat/template-asset.ts#L86-L103

packages/amplify-graphql-api-construct/src/internal/asset-manager.ts
dpilch: I'll expand more later but the basic gist is that the AssetManager in the construct is utilized by the AssetManager in the transformer. The transformer AssetManager knows that it is supposed to store assets, but it doesn't know the details of how this is done. The transformer AssetManager relies on the construct AssetManager to implement the details.

The `createAsset` function in the transformer AssetManager:
https://github.com/aws-amplify/amplify-category-api/blob/a5a233cfc1aa0859a9adf74f319b53c9479b38d4/packages/amplify-graphql-transformer-core/src/transformer-context/asset-manager.ts#L7-L9

is calling the provide function:
https://github.com/aws-amplify/amplify-category-api/blob/b044a26366653e4c505b040a7809e3bd56471594/packages/amplify-graphql-api-construct/src/amplify-graphql-api.ts#L193-L194

Which is then calling the `addAsset` function from the construct AssetManager:
https://github.com/aws-amplify/amplify-category-api/blob/b044a26366653e4c505b040a7809e3bd56471594/packages/amplify-graphql-api-construct/src/internal/asset-manager.ts#L22-L32

packages/amplify-graphql-api-construct/src/internal/asset-provider.ts
```diff
@@ -0,0 +1,64 @@
+import * as fs from 'fs';
```
dpilch, line 1: This file was moved from `packages/amplify-graphql-api-construct/src/internal/asset-manager.ts`. There are enough line changes that git is representing this as removing one file and adding another.

packages/amplify-graphql-api-construct/src/amplify-graphql-api.ts
```diff
@@ -401,7 +396,7 @@ export class AmplifyGraphqlApi extends Construct {
  * @param scope the scope this construct is created in.
  */
 const validateNoOtherAmplifyGraphqlApiInStack = (scope: Construct): void => {
-  const rootStack = getStackForScope(scope, true);
+  const rootStack = getStackForScope(scope, false);
```
dpilch, line 399: The second argument here is to check the whole tree (when true) or to only check the parent (when false).

https://github.com/aws-amplify/amplify-category-api/blob/a5a233cfc1aa0859a9adf74f319b53c9479b38d4/packages/amplify-graphql-api-construct/src/internal/construct-tree.ts#L9-L16

phani-srikar (approved):

palpatim (approved):

PR #2410, opened 2024-03-30
fix(amplify-table): describe ttl rate limit

AaronZyLee:
#### Description of changes
Changes to the DynamoDB API call `DescribeTimeToLive` to avoid throttling error when a large number of amplify tables are being deployed

- The API call is staggered by a random delay within 10s based on an interval of 100ms
- The exponential backoff is applied upon the throttling exception

Unit and e2e tests
Local CDK app deploying 60+ models
#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes


packages/amplify-graphql-api-construct-tests/src/__tests__/backends/amplify-table/rate-limit/updateTableTtl/disabled/app.ts
```diff
@@ -0,0 +1,33 @@
...
+
+const packageJson = require('../package.json');
+
+const app = new App();
+const stack = new Stack(app, packageJson.name.replace(/_/g, '-'), {
+  env: { region: process.env.CLI_REGION || 'us-west-2' },
+});
+const schema =
+  `input AMPLIFY { globalAuthRule: AuthRule = { allow: public } }\n` +
+  Array.from({ length: 60 }, (_, i) => i + 1)
```
AaronZyLee, line 15: There is a hard limit for `2500` resources for nested stack within one CFN operation (CRUD). For one nested stack of a model type, there are ~40 resources, which indicates the test for `60` is reasonable. In the real use case, this number will be smaller as there will be additional resources in the nested stack(such as function)
image

packages/amplify-graphql-model-transformer/src/resources/amplify-dynamodb-table/amplify-table-manager-lambda/amplify-table-manager-handler.ts
```diff
@@ -726,6 +733,24 @@ export const extractTableInputFromEvent = (
...
+ */
+export const extractOldTableInputFromEvent = (
+  event: AWSCDKAsyncCustomResource.OnEventRequest | AWSCDKAsyncCustomResource.IsCompleteRequest,
+): CustomDDB.Input => {
+  // isolate the resource properties from the event and remove the service token
+  const resourceProperties = { ...event.OldResourceProperties } as Record<string, any> & { ServiceToken?: string };
+  delete resourceProperties.ServiceToken;
+
+  // cast the remaining resource properties to the DynamoDB API call input type
+  const tableDef = convertStringToBooleanOrNumber(resourceProperties) as CustomDDB.Input;
```
palpatim, line 750: A little bit of a tangent, but I'm unclear why we're casting this to CustomDDB.Input. If I had to guess, I'd say that the incoming event is emitted by CDK, and the `OldResourceProperties` represented the event that last created/updated the resource. Since we know that the resource was created using our handler, we can cast it as a `CustomDDB.Input`. Is that right?

If that's the case, how are we ensuring that we maintain this input contract? Is the `CustomDDB.Input` API documented as public and locked so we don't accidentally break existing installations by changing a field name or adding a new required field that old installations might not have?

packages/amplify-graphql-model-transformer/src/resources/amplify-dynamodb-table/amplify-table-manager-lambda/amplify-table-manager-handler.ts
```diff
@@ -942,6 +967,80 @@ const isKeySchemaModified = (currentSchema: Array<KeySchemaElement>, endSchema:
...
+};
+
+/**
+ * Get time to live specification for the given table
+ * This API call has a limit rate. The call is staggered randomly within 10s and called with exponential retries
+ * @param tableName table name
+ * @returns time to live specification object
+ */
+const getTtlStatus = async (tableName: string): Promise<DescribeTimeToLiveCommandOutput> => {
+  const initialDelay = Math.floor(Math.random() * 10 * 1000); // between 0 to 10s
```
palpatim, line 996: Why would we implement an initial delay? The usual way of handling this would be to start our exponential retries if and when we hit a throttling limit. It seems counterintuitive, not to mention punitive, to introduce a 10 second delay for a 1-table update.

packages/amplify-graphql-model-transformer/src/resources/amplify-dynamodb-table/amplify-table-manager-lambda/amplify-table-manager-handler.ts
```diff
@@ -942,6 +967,80 @@ const isKeySchemaModified = (currentSchema: Array<KeySchemaElement>, endSchema:
...
+};
+
+/**
+ * Makes an API call with error retries and exponential backoff.
+ * @param requestData The data or parameters needed for the API call.
+ * @param maxRetries Maximum number of retries before giving up.
+ * @param retryDelayMs Initial delay between retries, in milliseconds.
+ * @returns A promise that resolves with the API call result or rejects after exceeding max retries.
+ */
+const makeApiCallWithExponentialRetriesOnThrottlingException = async <T>(
```
palpatim, line 1016: Can we reuse our [existing retrier](https://github.com/aws-amplify/amplify-category-api/blob/64a1b7f46ee2edd23aa9364106ad74374ff76aae/packages/amplify-graphql-model-transformer/src/resources/amplify-dynamodb-table/amplify-table-manager-lambda/amplify-table-manager-handler.ts#L969) rather than coding a new one?

packages/amplify-graphql-model-transformer/package.json
```diff
@@ -67,7 +67,7 @@
     "coverageThreshold": {
       "global": {
         "branches": 65,
-        "functions": 77,
+        "functions": 76,
```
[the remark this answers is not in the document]
AaronZyLee: This is addressed by merging the method to `retry` and also remove the test coverage for type definition files with `.d.ts` suffix

packages/amplify-graphql-model-transformer/src/resources/amplify-dynamodb-table/amplify-table-manager-lambda/amplify-table-manager-handler.ts
AaronZyLee: I've finally use the combination of `maxRetry=5` and `delay=5s`. Some datapoints are as follows.
| max retry | initial delay (s) | total deployment time (s)|
|---|---|---|
|5|10|407.1|
|5|0|383.82|
|5|5|382.45|

packages/amplify-graphql-model-transformer/src/resources/amplify-dynamodb-table/amplify-table-manager-lambda/amplify-table-manager-handler.ts
AaronZyLee: I've finally use the combination of `maxRetry=5` and `delay=5s`. Some datapoints are as follows.
| max retry | initial delay (s) | total deployment time (s)|
|---|---|---|
|5|10|407.1|
|5|0|383.82|
|5|5|382.45|

packages/amplify-graphql-model-transformer/src/resources/amplify-dynamodb-table/amplify-table-manager-lambda/amplify-table-manager-handler.ts
AaronZyLee: I've refactored the `retry` method to have exponential backoff

packages/amplify-graphql-model-transformer/src/resources/amplify-dynamodb-table/amplify-table-manager-lambda/amplify-table-manager-handler.ts
```diff
@@ -993,54 +993,23 @@ export const isTtlModified = (
  * @returns time to live specification object
  */
 const getTtlStatus = async (tableName: string): Promise<DescribeTimeToLiveCommandOutput> => {
-  const initialDelay = Math.floor(Math.random() * 10 * 1000); // between 0 to 10s
+  const initialDelay = Math.floor(Math.random() * 5 * 1000); // between 0 to 5s
```
palpatim, line 996: Thank you for testing this. Let's add a brief doc comment explaining why we're using an initial delay rather than just retrying on failure.

packages/amplify-graphql-model-transformer/src/resources/amplify-dynamodb-table/amplify-table-manager-lambda/amplify-table-manager-handler.ts
```diff
@@ -993,54 +993,23 @@ export const isTtlModified = (
...
   console.log(`Waiting for ${initialDelay} ms`);
   await sleep(initialDelay);
   // Max retry is 3
   // Retry delay is 1000ms
```
palpatim, line 1000: These comments are now inaccurate, and not terribly helpful in any case, since the var names are descriptive

packages/amplify-graphql-model-transformer/src/resources/amplify-dynamodb-table/amplify-table-manager-lambda/amplify-table-manager-handler.ts
palpatim: Thanks. Please add a doc comment in the CustomDDB.Input type that this type defines how data is stored in CloudFormation, so changes to the structure must be backward compatible with all previous versions.

packages/amplify-graphql-model-transformer/src/resources/amplify-dynamodb-table/amplify-table-manager-lambda/amplify-table-manager-handler.ts
palpatim: Given that the existing retrier doesn't have an initial delay, let's keep the custom retry method.

packages/amplify-graphql-model-transformer/src/resources/amplify-dynamodb-table/amplify-table-manager-lambda/amplify-table-manager-handler.ts
AaronZyLee: Added in [83c56ae](https://github.com/aws-amplify/amplify-category-api/pull/2410/commits/83c56aebf1507b757b28b981f741fc1030fb2a0a)

palpatim (approved):

phani-srikar (approved):

PR #2416, opened 2024-04-02
fix: propagate mutation context to relational fields

palpatim:
#### Description of changes

Propagates the `__operation` field to relational field resolvers.

- [x] DDB, hasOne
- [x] DDB, hasMany
- [x] DDB, belongsTo
- [x] DDB, manyToMany
- [x] SQL, hasOne
- [x] SQL, hasMany
- [x] SQL, belongsTo
- [x] Unit tests
- [x] [E2E tests](https://us-east-1.console.aws.amazon.com/codesuite/codebuild/594813022831/projects/amplify-category-api-e2e-workflow/batch/amplify-category-api-e2e-workflow:e036d324-9498-43b2-9aa8-2683b9bdaa44?region=us-east-1) (including newly added tests) pass, except for searchable test with known "invalid instance" errors.

The E2E tests added in this change include the full schema, API.ts, and operations generated from `amplify codegen --maxDepth 4`. You should feel free to disregard those files. We use them as test fixtures because they allow the newly-added [AppSync graphql util](https://github.com/aws-amplify/amplify-category-api/pull/2416/files#diff-a0561e6bca343f4fd344be36a06719c6a410228924fe29f47d673769fdaf753aR1) to provide a typesafe graphql invocation method that doesn't require the Amplify library.

We have separate E2Es for homogeneous DDB and SQL environments -- we want to test gen1-style DDB references, and in any case, `references` is only supported by SQL data sources currently. In Gen2, we'll extend these E2Es to test relationships in heterogeneous environments.

The E2Es do not test subscriptions, since they simply propagate the returned selection set of the mutation. As long as the mutation fields are properly redacted, the subscriptions will also be redacted.

The pattern we use for these E2Es is to set up the "associated" records of a relationship first, then mutate the "source" record and assert that its restricted fields, and those of all associated records, are null.

**Additional changes**

- Test coverage thresholds failed in graphql-transformer-core, even though this PR doesn't change anything in that package. I removed an unused file and added a small test to get coverage passing locally, but it doesn't seem to affect coverage in CI. Reduced coverage threshold in the package to work around.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-graphql-api-construct-tests/src/__tests__/graphql-schemas/restricted-field-auth/ddb-only/API.ts
+++ b/packages/amplify-graphql-api-construct-tests/src/__tests__/graphql-schemas/restricted-field-auth/ddb-only/API.ts
@@ -3,3171 +3,3169 @@
 //  This file was automatically generated and should not be edited.
 
 export type CreatePrimaryInput = {
-  id?: string | null,
-  secret?: string | null,
-  primaryRelatedOneId?: string | null,
+  id?: string | null;
+  secret?: string | null;
+  primaryRelatedOneId?: string | null;
 };
 
 export type ModelPrimaryConditionInput = {
-  secret?: ModelStringInput | null,
-  and?: Array< ModelPrimaryConditionInput | null > | null,
-  or?: Array< ModelPrimaryConditionInput | null > | null,
-  not?: ModelPrimaryConditionInput | null,
-  createdAt?: ModelStringInput | null,
-  updatedAt?: ModelStringInput | null,
-  primaryRelatedOneId?: ModelIDInput | null,
-  owner?: ModelStringInput | null,
+  secret?: ModelStringInput | null;
+  and?: Array<ModelPrimaryConditionInput | null> | null;
+  or?: Array<ModelPrimaryConditionInput | null> | null;
+  not?: ModelPrimaryConditionInput | null;
+  createdAt?: ModelStringInput | null;
+  updatedAt?: ModelStringInput | null;
+  primaryRelatedOneId?: ModelIDInput | null;
+  owner?: ModelStringInput | null;
 };
 
 export type ModelStringInput = {
-  ne?: string | null,
-  eq?: string | null,
-  le?: string | null,
-  lt?: string | null,
-  ge?: string | null,
-  gt?: string | null,
-  contains?: string | null,
-  notContains?: string | null,
-  between?: Array< string | null > | null,
-  beginsWith?: string | null,
-  attributeExists?: boolean | null,
-  attributeType?: ModelAttributeTypes | null,
-  size?: ModelSizeInput | null,
+  ne?: string | null;
+  eq?: string | null;
+  le?: string | null;
+  lt?: string | null;
+  ge?: string | null;
+  gt?: string | null;
+  contains?: string | null;
+  notContains?: string | null;
+  between?: Array<string | null> | null;
+  beginsWith?: string | null;
+  attributeExists?: boolean | null;
+  attributeType?: ModelAttributeTypes | null;
+  size?: ModelSizeInput | null;
 };
 
 export enum ModelAttributeTypes {
-  binary = "binary",
-  binarySet = "binarySet",
-  bool = "bool",
-  list = "list",
-  map = "map",
-  number = "number",
-  numberSet = "numberSet",
-  string = "string",
-  stringSet = "stringSet",
-  _null = "_null",
+  binary = 'binary',
+  binarySet = 'binarySet',
+  bool = 'bool',
+  list = 'list',
+  map = 'map',
+  number = 'number',
+  numberSet = 'numberSet',
+  string = 'string',
+  stringSet = 'stringSet',
+  _null = '_null',
 }
 
-
 export type ModelSizeInput = {
-  ne?: number | null,
-  eq?: number | null,
-  le?: number | null,
-  lt?: number | null,
-  ge?: number | null,
-  gt?: number | null,
-  between?: Array< number | null > | null,
+  ne?: number | null;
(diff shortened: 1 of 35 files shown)


packages/graphql-transformer-core/package.json
```diff
@@ -57,7 +57,7 @@
     "coverageProvider": "v8",
     "coverageThreshold": {
       "global": {
-        "branches": 66,
+        "branches": 65,
```
palpatim, line 60: See discussion for reasoning -- threshold passes locally but fails in CI

packages/graphql-transformer-core/src/__tests__/tableNameMap.test.ts
```diff
@@ -0,0 +1,33 @@
+import { getTableNameForModel } from '../tableNameMap';
```
palpatim, line 1: added this to try to get coverage threshold to pass

packages/amplify-graphql-relational-transformer/src/__tests__/__snapshots__/amplify-graphql-belongs-to-transformer.test.ts.snap
```diff
@@ -321,7 +321,12 @@ exports[`@belongsTo directive with RDS datasource composite key should generate
 #if( $ctx.error )
   $util.error($ctx.error.message, $ctx.error.type)
 #else
-  $util.toJson($ctx.result)
+  #set( $resultValue = $ctx.result )
+  #set( $operation = $util.defaultIfNull($ctx.source.get(\\"__operation\\"), null) )
+  #if( $operation == \\"Mutation\\" )
```
phani-srikar, line 326: here and other places:
Instead of checking for null and defaulting to null again, does the below simplification work?
```
#if( $ctx.source.get(\\"__operation\\") == \\"Mutation\\" )
```

packages/amplify-graphql-relational-transformer/src/resolver/ddb-generator.ts
```diff
@@ -205,7 +207,21 @@ export class DDBRelationalResolverGenerator extends RelationalResolverGenerator
...
+              iff(raw('!$result'), set(ref('result'), ref('ctx.result'))),
+
+              // Make sure each retrieved item has the __operation field, so the individual type resolver can appropriately redact fields
+              compoundExpression([
+                iff(
+                  equals(
+                    methodCall(ref('util.defaultIfNull'), methodCall(ref('ctx.source.get'), str(OPERATION_KEY)), nul()),
+                    str('Mutation'),
+                  ),
+                  forEach(ref('item'), ref('result.items'), [qref(methodCall(ref('item.put'), str(OPERATION_KEY), str('Mutation')))]),
```
phani-srikar, line 220: here we're directly mutating the `$ctx.result` object and returning it.

packages/amplify-graphql-relational-transformer/src/resolver/rds-generator.ts
```diff
@@ -149,10 +153,41 @@ export class RDSRelationalResolverGenerator extends RelationalResolverGenerator
...
+
+  generateListConnectionLambdaResponseMappingTemplate = (): string => {
+    const statements: Expression[] = [];
+    statements.push(
+      ifElse(
+        ref('ctx.error'),
+        methodCall(ref('util.error'), ref('ctx.error.message'), ref('ctx.error.type')),
+        // Make sure each retrieved item has the __operation field, so the individual type resolver can appropriately redact fields
+        compoundExpression([
+          set(ref('resultValue'), ref('ctx.result')),
```
phani-srikar, line 183: here we're assigning the `ctx.result` to a new value, mutating the new value and returning it.
Functionally I don't think it makes a difference, but why have 2 approaches? Treat this as a nit.

packages/amplify-graphql-api-construct-tests/src/utils/auth.ts
```diff
@@ -0,0 +1,116 @@
+import {
```
phani-srikar, line 1: Thanks for adding these utilities. I've noticed that these were missing as well on Friday.

packages/amplify-graphql-relational-transformer/src/resolver/rds-generator.ts
palpatim: I wondered that myself, and elected to leave it because it was closer to the style of the surrounding code. When we merge this fix into the `feature/gen2-release` branch we can look at normalizing this.

packages/amplify-graphql-api-construct-tests/src/__tests__/graphql-schemas/restricted-field-auth/ddb-only/API.ts
AaronZyLee, line 1: Do we use the `API.ts` file in the e2e tests? I notice the GQL operations generated by `amplify codegen statements` are used for running GQL queries but cannot find any import for `API.ts`. If it is not used, this file can be removed.

dpilch (approved):

phani-srikar (approved):

iartemiev (approved):

AaronZyLee (approved):

packages/amplify-graphql-api-construct-tests/src/__tests__/graphql-schemas/restricted-field-auth/ddb-only/graphql/schema.json
```diff
@@ -0,0 +1,7775 @@
+{
```
sundersc, line 1: Looks these are codegen generated files. Are we sure we want to check-in these files? Should we generate at runtime instead?

PR #2236, opened 2024-02-01
feat: add implicit fields to filter input

dpilch:
#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes
- [ ] Tests are [changed or added](https://github.com/aws-amplify/amplify-cli/blob/master/CONTRIBUTING.md#tests)
- [ ] Relevant documentation is changed or added (and PR referenced)
- [ ] New AWS SDK calls or CloudFormation actions have been added to relevant test and service IAM policies
- [ ] Any CDK or CloudFormation parameter changes are called out explicitly

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #2435, opened 2024-04-08
fix: remove null timestamp fields from filter input

dpilch:
#### Description of changes

This bug was introduced in https://github.com/aws-amplify/amplify-category-api/pull/2236

* A field with null name is included when timestamps are removed with `@model(timestamps: { createdAt/updatedAt: null })`.
```
type Foo @model(timestamps: { updatedAt: null }) {
  id: ID!
  title: String!
}
```
This results in schema deploy error because of field named `null`.
```
input ModelFooConditionInput {
  title: ModelStringInput
  and: [ModelFooConditionInput]
  or: [ModelFooConditionInput]
  not: ModelFooConditionInput
  createdAt: ModelStringInput
  null: ModelStringInput // should not be included
}
```
* Bad filter inputs are added when timestamps are removed with `@model(timestamps: null)`
```
type Foo @model(timestamps: null) {
  id: ID!
  title: String!
}
```

Schema deploys fine in this case.
```
input ModelFooConditionInput {
  title: ModelStringInput
  and: [ModelFooConditionInput]
  or: [ModelFooConditionInput]
  not: ModelFooConditionInput
  createdAt: ModelStringInput // should not be included
  updatedAt: ModelStringInput // should not be included
}
```

N/A

#### Issue #, if available

N/A (reported through AWS support)

* Unit tests
* E2E tests

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-graphql-model-transformer/src/__tests__/model-transformer.test.ts
+++ b/packages/amplify-graphql-model-transformer/src/__tests__/model-transformer.test.ts
@@ -1686,7 +1686,7 @@ describe('ModelTransformer:', () => {
   describe('remove null timestamp fields from input', () => {
     it('updatedAt null', () => {
       const validSchema = `
-        type Foo @model(timestamps: { updatedAt: null }) {
+        type UpdatedAtNull @model(timestamps: { updatedAt: null }) {
             id: ID!
             title: String!
         }
@@ -1704,7 +1704,7 @@ describe('ModelTransformer:', () => {
 
     it('createdAt null', () => {
       const validSchema = `
-        type Bar @model(timestamps: { createdAt: null }) {
+        type CreatedAtNull @model(timestamps: { createdAt: null }) {
             id: ID!
             title: String!
         }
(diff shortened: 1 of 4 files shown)


palpatim (approved):

packages/amplify-graphql-model-transformer/src/__tests__/model-transformer.test.ts
```diff
@@ -1682,4 +1682,33 @@ describe('ModelTransformer:', () => {
       });
     });
   });
+
+  it('should remove null timestamp fields from input', () => {
+    const validSchema = `
+      type Foo @model(timestamps: { updatedAt: null }) {
```
sundersc, line 1688: Can we add the below test cases?
- Both values explicitly mentioned? (timestamps: { createdAt: "createdOn", updatedAt: "updatedOn" })
- Both values explicitly mentioned and one is set to null? (timestamps: { createdAt: "createdOn", updatedAt: null })

sundersc (approved):

palpatim (approved):

PR #2440, opened 2024-04-09
fix(api): add identifier to generated ts data schema

sundersc:
#### Description of changes

Add identifier to the generated typescript data schema file. This identifier is later used with `defineData` to construct an unique datasource strategy ID.

NA

#### Issue #, if available
NA

- New unit tests
- Manual test

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-graphql-schema-generator/src/__tests__/generate-ts-data-schema.test.ts
+++ b/packages/amplify-graphql-schema-generator/src/__tests__/generate-ts-data-schema.test.ts
@@ -10,15 +10,15 @@ jest.mock('../utils', () => ({
       securityGroupIds: ['sg0', 'sg1', 'sg2'],
       subnetAvailabilityZoneConfig: [
         {
-          subnetId: 'sg0',
+          subnetId: 'sb0',
           availabilityZone: 'az0',
         },
         {
-          subnetId: 'sg1',
+          subnetId: 'sb1',
           availabilityZone: 'az1',
         },
         {
-          subnetId: 'sg2',
+          subnetId: 'sb2',
           availabilityZone: 'az2',
         },
       ],
--- a/packages/amplify-graphql-schema-generator/src/__tests__/generate-ts-data-schema.test.ts
+++ b/packages/amplify-graphql-schema-generator/src/__tests__/generate-ts-data-schema.test.ts
@@ -3,6 +3,29 @@ import { generateTypescriptDataSchema } from '../ts-schema-generator/generate-ts
 import { DataSourceConfig } from '../ts-schema-generator/helpers';
 import { TypescriptDataSchemaGenerator } from '../ts-schema-generator/ts-schema-generator';
 
+jest.mock('../utils', () => ({
+  getHostVpc: jest.fn(() => {
+    return {
+      vpcId: 'abc',
+      securityGroupIds: ['sb0', 'sb1', 'sb2'],
+      subnetAvailabilityZoneConfig: [
+        {
+          subnetId: 'sb1',
+          availabilityZone: 'az1',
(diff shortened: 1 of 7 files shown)


packages/amplify-graphql-schema-generator/src/__tests__/generate-ts-data-schema.test.ts
```diff
@@ -1,8 +1,57 @@
 import { Engine, Field, Model, Schema } from '../schema-representation';
 import { generateTypescriptDataSchema } from '../ts-schema-generator/generate-ts-schema';
 import { DataSourceConfig } from '../ts-schema-generator/helpers';
+import { TypescriptDataSchemaGenerator } from '../ts-schema-generator/ts-schema-generator';
+
+jest.mock('../utils', () => ({
+  getHostVpc: jest.fn(() => {
+    return {
+      vpcId: 'abc',
+      securityGroupIds: ['sb0', 'sb1', 'sb2'],
```
palpatim, line 10: nit: "sg" instead of "sb"

packages/amplify-graphql-schema-generator/src/ts-schema-generator/ts-schema-generator.ts
```diff
@@ -54,4 +56,17 @@ export class TypescriptDataSchemaGenerator {
     models.forEach((m) => schema.addModel(m));
     return schema;
   };
+
+  private static createIdentifier = (config: TypescriptDataSchemaGeneratorConfig): string => {
+    const { host, database, engine, port } = config;
+    const identifierString = host.concat(engine, database, port.toString());
+    const identifierHash = createHash('md5').update(identifierString).digest('base64');
```
palpatim, line 63: Do we need to be using MD5? I know we're not doing anything particularly sensitive, but it'll inevitably cause false positives in security scanners that we'll have to address. SHA2 would be a better choice if the length constraints support it (which we should validate with tests). If not, add a doc comment here to explain how we're using MD5 and why it's not a security-sensitive use case

sundersc: Added comments explaining why we use 'md5' here. Considering our intention here is to generate an unique id and not to check the integrity of the data, it is okay to use 'md5'. Also, we only use the hostname, engine, port number and database to generate the ID, this doesn't post a security threat.

packages/amplify-graphql-schema-generator/src/__tests__/generate-ts-data-schema.test.ts
sundersc: Updated.

packages/amplify-graphql-schema-generator/src/__tests__/generate-ts-data-schema.test.ts
```diff
@@ -7,19 +7,19 @@ jest.mock('../utils', () => ({
   getHostVpc: jest.fn(() => {
     return {
       vpcId: 'abc',
-      securityGroupIds: ['sb0', 'sb1', 'sb2'],
+      securityGroupIds: ['sg0', 'sg1', 'sg2'],
       subnetAvailabilityZoneConfig: [
         {
-          subnetId: 'sb1',
-          availabilityZone: 'az1',
+          subnetId: 'sg0',
```
palpatim, line 13: subnets can stay "sb" -- i was trying to say that we needed to disambiguate your identifiers so that we don't confuse security group IDs with subnet IDs. :)

palpatim (approved):

phani-srikar (approved):

PR #2447, opened 2024-04-10
fix: propagate operation in references-style ddb resolver

palpatim:
#### Description of changes

This change ports the fix from https://github.com/aws-amplify/amplify-category-api/pull/2416 into the DDB references-style resolvers. Other changes:

- Updated DDB references snapshot to reflect this change
- Add E2Es to test redaction in relationships with heterogeneous data sources
- Refactor restricted-auth tests to support reuse of gen2 schemas
- Removed unused schema.json from test fixtures, per earlier PR comment

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] [E2E tests](https://us-east-1.console.aws.amazon.com/codesuite/codebuild/594813022831/projects/amplify-category-api-e2e-workflow/batch/amplify-category-api-e2e-workflow:40298f33-7a99-4abc-a444-2a0ee3f9a59e?region=us-east-1) pass


packages/amplify-graphql-api-construct-tests/src/__tests__/restricted-field-auth-sql-only.test.ts
```diff
@@ -1,447 +0,0 @@
-import * as path from 'path';
```
palpatim, line 1: This was renamed to `packages/amplify-graphql-api-construct-tests/src/__tests__/restricted-field-auth/restricted-field-auth-gen2.test.ts` and pretty heavily refactored to add new describe blocks and reuse the test methods.

packages/amplify-graphql-api-construct-tests/src/__tests__/restricted-field-auth/graphql-schemas/gen2/schema-primary.graphql
```diff
@@ -0,0 +1,7 @@
+type Primary @model @auth(rules: [{ allow: public, operations: [read] }, { allow: owner }]) {
```
palpatim, line 1: Broke the gen2 schema into 2 files so we can mix & match primary & related between our data sources

packages/amplify-graphql-api-construct-tests/src/__tests__/restricted-field-auth/restricted-field-auth-gen2.test.ts
```diff
@@ -0,0 +1,440 @@
+import * as path from 'path';
```
palpatim, line 1: This is a rename & refactor of the "sql-only" test. Primary changes:
- Add new cases for heterogeneous data sources
- Add new case for homogeneous DDB relationships using Gen2 references style relationships
- Refactored test implementations to promote reuse.

Note that we can't use the implementations in the gen1 test, because they rely on the gen2 operations & API.ts.

palpatim (commented):
marker review

palpatim:
E2E test failure cleaning up RDS instance, otherwise all green

palpatim:
> 1. The lack of subscription tests means we're assuming (I believe safely) that fields redacted in mutation responses are redacted in subscription payloads. Is that correct and is that a safe assumption (i.e. do we need test E2E tests using subscriptions here)?

We will need subscription tests but I plan on including those separately (we need them for runtime filtering validation as well as other cases). The field redaction logic is sufficiently validated with the mutation tests.

> 2. What are your thoughts on adding mutations that go >1 layer of relational nesting in the selection set to ensure we don't have regressions around the recursive setting of `__operation`?

We should definitely test that, which is why the 'hasOne' tests include a selection set and assertion for `expect(relatedOne.primary.relatedMany.items[0].secret).toBeNull();` :)

svidgen (approved):

PR #2462, opened 2024-04-15
test: add association field tests

palpatim:
#### Description of changes

Refactors CDK tests to make test files all setup, no logic. Test implementations are tied to a specific schema and reused for a single test suite targeted toward a particular auth area.

Added tests for associated fields.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-graphql-api-construct-tests/src/__tests__/owner-auth/assoc-field/API.ts
```diff
@@ -0,0 +1,1090 @@
+/* tslint:disable */
```
palpatim, line 1: Note that I've edited the queries to removed createdAt, updatedAt, and __typename fields to clean up the interface and remove the need to model those in SQL DBs.

packages/amplify-graphql-api-construct-tests/src/__tests__/backends/configurable-stack/app.ts
```diff
@@ -33,9 +37,12 @@ const combineTestDefinitionsInDirectory = (directory: string): IAmplifyGraphqlDe
   return AmplifyGraphqlDefinition.combine(definitions);
 };
 
-// Keeping this short so we don't bump up against resource length limits (e.g., 140 characters for lambda layers).
+const projRoot = path.normalize(path.join(__dirname, '..'));
+
+// Read the prefix to use when naming stack assets. Keep this short so you don't bump up against resource length limits (e.g., 140
+// characters for lambda layers).
```
phani-srikar, line 43: Since these prefixes are currently hand-written, 
- how do we know what's the max prefix length that won't cross 140?
- can we trim it to say 100 chars long which we know will succeed before deploying? I don't think we will need those many chars since these are essentially like test IDs.

packages/amplify-graphql-api-construct-tests/src/__tests__/backends/configurable-stack/app.ts
```diff
@@ -62,15 +69,14 @@ const userPoolClient = userPool.addClient(`${PREFIX}UserPoolClient`, {
...
-const combinedDefinition = combineTestDefinitionsInDirectory(testDefinitionDir);
+const combinedDefinition = combineTestDefinitionsInDirectory(projRoot);
 
 new AmplifyGraphqlApi(stack, `${PREFIX}Api`, {
   definition: combinedDefinition,
   authorizationModes: {
     defaultAuthorizationMode: 'AMAZON_COGNITO_USER_POOLS',
     userPoolConfig: { userPool },
-    apiKeyConfig: { expires: Duration.days(360) },
+    apiKeyConfig: { expires: Duration.days(7) },
```
phani-srikar, line 79: Since these are test apps meant to be torn down, why not go with the minimum i.e 1 day?

packages/amplify-graphql-api-construct-tests/src/__tests__/owner-auth/assoc-field/assoc-field.test.ts
```diff
@@ -0,0 +1,720 @@
...
+      test('getRelatedMany shows relations if created by same owner', async () => {
+        await testGetRelatedManyVisibleForSameOwner(currentId, apiEndpoint, accessToken);
+      });
+
+      test('getRelatedMany redacts relations if created by same owner', async () => {
+        await testGetRelatedManyRedactedForDifferentOwners(currentId, apiEndpoint, accessToken, accessToken2);
+      });
+    });
+  });
+});
```
phani-srikar, line 720: There's a lot of tests including different data source types in this suite. Wondering if it hits the token expiry limit or was it reliably succeeding on CI?

packages/amplify-graphql-auth-transformer/src/__tests__/owner-auth.test.ts
```diff
@@ -560,6 +560,58 @@ describe('owner based @auth', () => {
...
+  test('@auth in bidirectional relational schema using references', () => {
+    const inputSchema = `
+      type Primary @model @auth(rules: [{ allow: owner }]) {
+        id: String! @primaryKey
+        owner: String
+        relatedMany: [RelatedMany] @hasMany(references: ["primaryId"])
+        relatedOne: RelatedOne @hasOne(references: ["primaryId"])
+      }
+
+      type RelatedMany @model @auth(rules: [{ allow: owner }]) {
```
phani-srikar, line 572: If not already covered in other tests, it would be nice to have the following additional case either in this or a separate schema:
- using the connected field i.e `primaryId` as the `ownerField`

packages/amplify-graphql-api-construct-tests/src/__tests__/backends/configurable-stack/app.ts
palpatim: Reasonable. Will fix in upcoming PR

packages/amplify-graphql-api-construct-tests/src/__tests__/owner-auth/assoc-field/assoc-field.test.ts
palpatim: That's a very good question. It succeeds reliably on my local machine, but I haven't yet run it in CI. I probably need to make sure this suite is run solo though

packages/amplify-graphql-auth-transformer/src/__tests__/owner-auth.test.ts
palpatim: That's my next case :)

packages/amplify-graphql-api-construct-tests/src/__tests__/backends/configurable-stack/app.ts
palpatim: The prefix must be way shorter than 140 -- the example is the lambda resource that broke last time I ran into the length limit. In an upcoming PR I'll truncate to <= 20 chars

packages/amplify-graphql-api-construct-tests/src/__tests__/owner-auth/assoc-field/assoc-field.test.ts
palpatim: If we hit token expiry, IN an upcoming PR, I'll split each of the "source model" variations into their own test file so they can be run independently.

phani-srikar (approved):

packages/amplify-graphql-api-construct-tests/src/__tests__/owner-auth/assoc-field/test-implementations.ts
```diff
@@ -0,0 +1,576 @@
+import { AppSyncGraphqlResponse, doAppSyncGraphqlMutation, doAppSyncGraphqlQuery } from '../../../utils';
+import {
```
AaronZyLee, line 2: nit: A code comment can be added to indicate that the `API.ts` and queries, mutations and subscriptions in `./graphql/` are generated by `amplify codegen --maxDepth 4`

AaronZyLee (approved):

PR #2464, opened 2024-04-16
feat: allow override table name

pavinthan:
#### Description of changes

Allow overwritable DynamoDB table name 

#### Issue #, if available

N/A

1. Test in sample app

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-graphql-api-construct/src/amplify-dynamodb-table-wrapper.ts
+++ b/packages/amplify-graphql-api-construct/src/amplify-dynamodb-table-wrapper.ts
@@ -119,6 +119,13 @@ export class AmplifyDynamoDbTableWrapper {
   }
 
   /**
+   * Enforces a particular physical table name.
+   */
+  set tableName(tableName: string) {
+    this.resource.addPropertyOverride('tableName', tableName);
+  }
+
+  /**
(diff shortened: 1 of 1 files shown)


renebrandel:
@pavinthan  - overriding the table name like this could be tricky to maintain long-term. Are there alternatives, you've potentially explored using the `amplifyApi.resources.cfnResources`?

Separately, we're currently working on a feature to allow you to describe you model but "point to an existing DynamoDB table". I think that might setup the same use case as you're trying to solve here.

pavinthan:
@renebrandel, thanks for the feedback! I’m excited about the feature to link models to existing DynamoDB tables, but I still see a unique value in allowing custom table names right from the start—especially for new projects without existing tables. This avoids the need for additional CDK apps or tools.

On the concern about name conflicts across different environments, like multiple sandboxes in the same region, I suggest leveraging the empty cfnTables in cfnResources. We could allow users to include Amplify-specific identifiers (appID, branch name, and environment name) from defineBackend in their table names. This would enable a flexible yet standardized naming convention.

Incorporating this feature would give us tighter control over resource naming from the get-go, making projects easier to manage as they grow.

PR #2470, opened 2024-04-17
fix: auto generated id when timestamps: null

dpilch:
#### Description of changes

The auto generated ID is missing when timestamps are removed `@model(timestamps: null)`. This PR fixes that issue. This PR also remove auto id for id fields with type Float, Int, and Boolean.

The auto ID is used in these cases (see test cases for schemas):
* explicit id with type ID
* explicit id with type String
  * auto id is not used for other types (Float, Int, Boolean)
* implicit id

N/A

#### Issue #, if available

Resolves: https://github.com/aws-amplify/amplify-category-api/issues/978

* Unit tests
* E2E tests: https://us-east-1.console.aws.amazon.com/codesuite/codebuild/594813022831/projects/amplify-category-api-e2e-workflow/batch/amplify-category-api-e2e-workflow:bcb694d0-590a-4c36-9137-33cb41354d32?region=us-east-1

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-graphql-index-transformer/src/__tests__/__snapshots__/amplify-graphql-index-transformer.test.ts.snap
+++ b/packages/amplify-graphql-index-transformer/src/__tests__/__snapshots__/amplify-graphql-index-transformer.test.ts.snap
@@ -4,6 +4,7 @@ exports[`@index with a single sort key adds a query field and GSI correctly 1`] 
 Object {
   "Mutation.createTest.init.1.req.vtl": "## [Start] Initialization default values. **
 $util.qr($ctx.stash.put(\\"defaultValues\\", $util.defaultIfNull($ctx.stash.defaultValues, {})))
+$util.qr($ctx.stash.defaultValues.put(\\"id\\", $util.autoId()))
 #set( $createdAt = $util.time.nowISO8601() )
 $util.qr($ctx.stash.defaultValues.put(\\"createdAt\\", $createdAt))
 $util.qr($ctx.stash.defaultValues.put(\\"updatedAt\\", $createdAt))
@@ -609,6 +610,7 @@ exports[`@index with multiple sort keys adds a query field and GSI correctly 1`]
 Object {
   "Mutation.createTest.init.1.req.vtl": "## [Start] Initialization default values. **
 $util.qr($ctx.stash.put(\\"defaultValues\\", $util.defaultIfNull($ctx.stash.defaultValues, {})))
+$util.qr($ctx.stash.defaultValues.put(\\"id\\", $util.autoId()))
 #set( $createdAt = $util.time.nowISO8601() )
 $util.qr($ctx.stash.defaultValues.put(\\"createdAt\\", $createdAt))
 $util.qr($ctx.stash.defaultValues.put(\\"updatedAt\\", $createdAt))
--- a/packages/amplify-graphql-index-transformer/src/__tests__/__snapshots__/amplify-graphql-index-transformer.test.ts.snap
+++ b/packages/amplify-graphql-index-transformer/src/__tests__/__snapshots__/amplify-graphql-index-transformer.test.ts.snap
@@ -5,7 +5,6 @@ Object {
   "Mutation.createTest.init.1.req.vtl": "## [Start] Initialization default values. **
 $util.qr($ctx.stash.put(\\"defaultValues\\", $util.defaultIfNull($ctx.stash.defaultValues, {})))
 #set( $createdAt = $util.time.nowISO8601() )
-$util.qr($ctx.stash.defaultValues.put(\\"id\\", $util.autoId()))
 $util.qr($ctx.stash.defaultValues.put(\\"createdAt\\", $createdAt))
 $util.qr($ctx.stash.defaultValues.put(\\"updatedAt\\", $createdAt))
 $util.toJson({
@@ -611,7 +610,6 @@ Object {
   "Mutation.createTest.init.1.req.vtl": "## [Start] Initialization default values. **
 $util.qr($ctx.stash.put(\\"defaultValues\\", $util.defaultIfNull($ctx.stash.defaultValues, {})))
 #set( $createdAt = $util.time.nowISO8601() )
-$util.qr($ctx.stash.defaultValues.put(\\"id\\", $util.autoId()))
 $util.qr($ctx.stash.defaultValues.put(\\"createdAt\\", $createdAt))
 $util.qr($ctx.stash.defaultValues.put(\\"updatedAt\\", $createdAt))
 $util.toJson({
@@ -1323,8 +1321,8 @@ exports[`@index with no sort key field adds a query field and GSI correctly 1`] 
 Object {
   "Mutation.createTest.init.1.req.vtl": "## [Start] Initialization default values. **
 $util.qr($ctx.stash.put(\\"defaultValues\\", $util.defaultIfNull($ctx.stash.defaultValues, {})))
-#set( $createdAt = $util.time.nowISO8601() )
 $util.qr($ctx.stash.defaultValues.put(\\"id\\", $util.autoId()))
+#set( $createdAt = $util.time.nowISO8601() )
 $util.qr($ctx.stash.defaultValues.put(\\"createdAt\\", $createdAt))
 $util.qr($ctx.stash.defaultValues.put(\\"updatedAt\\", $createdAt))
 $util.toJson({
@@ -3048,7 +3046,6 @@ Object {
   "Mutation.createTest.init.1.req.vtl": "## [Start] Initialization default values. **
 $util.qr($ctx.stash.put(\\"defaultValues\\", $util.defaultIfNull($ctx.stash.defaultValues, {})))
 #set( $createdAt = $util.time.nowISO8601() )
-$util.qr($ctx.stash.defaultValues.put(\\"id\\", $util.autoId()))
 $util.qr($ctx.stash.defaultValues.put(\\"createdAt\\", $createdAt))
 $util.qr($ctx.stash.defaultValues.put(\\"updatedAt\\", $createdAt))
 $util.toJson({
@@ -3758,7 +3755,6 @@ Object {
   "Mutation.createItem.init.1.req.vtl": "## [Start] Initialization default values. **
 $util.qr($ctx.stash.put(\\"defaultValues\\", $util.defaultIfNull($ctx.stash.defaultValues, {})))
 #set( $createdAt = $util.time.nowISO8601() )
-$util.qr($ctx.stash.defaultValues.put(\\"id\\", $util.autoId()))
 $util.qr($ctx.stash.defaultValues.put(\\"createdAt\\", $createdAt))
 $util.qr($ctx.stash.defaultValues.put(\\"updatedAt\\", $createdAt))
 $util.toJson({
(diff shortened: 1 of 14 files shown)


packages/amplify-graphql-index-transformer/src/__tests__/__snapshots__/amplify-graphql-index-transformer.test.ts.snap
```diff
@@ -5,7 +5,6 @@ Object {
   "Mutation.createTest.init.1.req.vtl": "## [Start] Initialization default values. **
 $util.qr($ctx.stash.put(\\"defaultValues\\", $util.defaultIfNull($ctx.stash.defaultValues, {})))
 #set( $createdAt = $util.time.nowISO8601() )
-$util.qr($ctx.stash.defaultValues.put(\\"id\\", $util.autoId()))
```
dpilch, line 8: In these test cases there is no `id` field (implicit or explicit) on the model. Removing this default value will have no effect.

packages/amplify-graphql-index-transformer/src/__tests__/__snapshots__/amplify-graphql-index-transformer.test.ts.snap
```diff
@@ -3048,7 +3048,6 @@ Object {
...
```
dpilch, line 3051: In these test cases there is no id field (implicit or explicit) on the model. Removing this default value will have no effect.

It would be worth double checking my work here though.

packages/amplify-graphql-model-transformer/src/__tests__/model-transformer.test.ts
```diff
@@ -1906,4 +1906,239 @@ describe('ModelTransformer:', () => {
...
+        const out = testTransform({
+          schema,
+          transformers: [new ModelTransformer()],
+        });
+        expect(out.resolvers['Mutation.createPost.init.1.req.vtl']).toContain(
+          '$util.qr($ctx.stash.defaultValues.put("id", $util.autoId()))',
+        );
+      });
+
+      it('should include autoId when timestamps are null', async () => {
```
palpatim, line 1945: what about timestamps null and implicit id?

```graphql
type Post @model(timestamps: null) {
  title: String!
}
```

packages/amplify-graphql-model-transformer/src/__tests__/model-transformer.test.ts
```diff
@@ -1906,4 +1906,239 @@ describe('ModelTransformer:', () => {
...
+          schema,
+          transformers: [new ModelTransformer(), new PrimaryKeyTransformer()],
+          dataSourceStrategies: constructDataSourceStrategies(schema, makeStrategy(MYSQL_DB_TYPE)),
+        });
+        expect(out.resolvers['Mutation.createPost.init.1.req.vtl']).toContain(
+          '$util.qr($ctx.stash.defaultValues.put("id", $util.autoId()))',
+        );
+      });
+
+      it('should not include autoId for implicit ID', async () => {
```
palpatim, line 2050: This isn't an implicit id, this is an explicit, right? I would not expect an `id` field to be generated here

packages/amplify-graphql-model-transformer/src/__tests__/model-transformer.test.ts
```diff
@@ -1906,4 +1906,239 @@ describe('ModelTransformer:', () => {
...
+          schema,
+          transformers: [new ModelTransformer(), new PrimaryKeyTransformer()],
+          dataSourceStrategies: constructDataSourceStrategies(schema, makeStrategy(MYSQL_DB_TYPE)),
+        });
+        expect(out.resolvers['Mutation.createPost.init.1.req.vtl']).not.toContain(
+          '$util.qr($ctx.stash.defaultValues.put("id", $util.autoId()))',
+        );
+      });
+
+      it('should include autoId when timestamps are null', async () => {
```
palpatim, line 2068: Is it expected that we would generate an autoId for an explicitly declared `id` field even when there's an explicit primaryKey?

packages/amplify-graphql-model-transformer/src/__tests__/model-transformer.test.ts
```diff
@@ -1906,4 +1906,239 @@ describe('ModelTransformer:', () => {
...
+          transformers: [new ModelTransformer(), new PrimaryKeyTransformer()],
+        });
+        expect(out.resolvers['Mutation.createPost.init.1.req.vtl']).not.toContain(
+          '$util.qr($ctx.stash.defaultValues.put("id", $util.autoId()))',
+        );
+      });
+    });
+
+    describe('sql', () => {
+      it('should include autoId for basic ID', async () => {
```
palpatim, line 2031: Is it expected that we would generate an autoId for an explicitly declared `id` field even though there is an explicit primaryKey?

packages/amplify-graphql-model-transformer/src/resolvers/dynamodb/mutation.ts
```diff
@@ -234,9 +234,11 @@ export const generateCreateInitSlotTemplate = (modelConfig: ModelDirectiveConfig
     ),
   ];
 
+  if (initializeIdField) {
+    statements.push(qref(methodCall(ref('ctx.stash.defaultValues.put'), str('id'), methodCall(ref('util.autoId')))));
```
palpatim, line 238: Can we extract the qref expression into a standalone utility & reuse it everywhere we set an autoId?

palpatim (commented):
1. I'm a bit confused that none of the snapshots have any changes other than a reordering of the createdAt field. Is it expected that no snapshot would be impacted by the changes in this PR?
2. I'm more than a bit confused about the rules behind:
    - When we generate an implicit ID field in the first place, and
    - When it is expected that we will auto-populate an ID field with a UUID

    Are these rules published somewhere for customers? It is strange to me that we're auto-generating ID fields in some cases even when there's an explicit primary key, and strange that we're auto populating an explicitly declared `id` field that's NOT a primary key.

packages/amplify-graphql-model-transformer/src/__tests__/model-transformer.test.ts
dpilch: That is the current behavior (as long as timestamps are not null) for dynamodb and sql. Changing this would be breaking now.

packages/amplify-graphql-model-transformer/src/__tests__/model-transformer.test.ts
dpilch: Fixed the name. This test is for no `id` field now.

dpilch:
1. We do not have a test that takes a snapshot of createModel init resolver for a model with `timestamps: null`. This a bug that has existed for at least 4 years and I found 1 bug report for it (https://github.com/aws-amplify/amplify-category-api/issues/978). We just did not realize that we would need to test for this.
2. Implicit ID is added for all dynamodb models not using `@primaryKey`. Implicit ID is never added for sql models.

Behavior before this change:
* Auto id is used on explicit id with any type
  * This throws an error for Float and Int when trying to use the create mutation without supplying the id in the input because the casting fails.
* Auto id is used implicit id
  * implicit id will only appear on dynamodb models not using `@primaryKey`.
  * implicit id will only have type ID
* Auto id is not used for implicit or explicit id for dynamodb or sql when the `timestamps: null` is set

AFAIK these are not documented. I found this behavior by creating the unit tests.

palpatim (dismissed):

packages/amplify-graphql-model-transformer/src/__tests__/model-transformer.test.ts
```diff
@@ -1906,4 +1906,255 @@ describe('ModelTransformer:', () => {
...
+        const out = testTransform({
+          schema,
+          transformers: [new ModelTransformer()],
+        });
+        expect(out.resolvers['Mutation.createPost.init.1.req.vtl']).toContain(
+          '$util.qr($ctx.stash.defaultValues.put("id", $util.autoId()))',
+        );
+      });
+
+      it('should include autoId when id is type Int, Float, or Boolean', async () => {
```
AaronZyLee, line 1995:
```suggestion
      it('should not include autoId when id is type Int, Float, or Boolean', async () => {
```

packages/amplify-graphql-model-transformer/src/__tests__/model-transformer.test.ts
```diff
@@ -1906,4 +1906,255 @@ describe('ModelTransformer:', () => {
       expect(out.schema).toMatchSnapshot();
     });
   });
+
+  describe.only('autoId', () => {
```
AaronZyLee, line 1910:
```suggestion
  describe('autoId', () => {
```

packages/amplify-graphql-model-transformer/src/__tests__/model-transformer.test.ts
AaronZyLee: Same question here. Also the schema used is different from the same case in DDB which has the `postId: ID! @primaryKey` defined. Does the DDB act the same as SQL with the same schema in such case?

AaronZyLee: Just see the comments below. Looks like the DDB is the same and changing the behavior is breaking.

packages/amplify-graphql-model-transformer/src/__tests__/model-transformer.test.ts
```diff
@@ -1906,4 +1906,255 @@ describe('ModelTransformer:', () => {
...
+          schema,
+          transformers: [new ModelTransformer(), new PrimaryKeyTransformer()],
+          dataSourceStrategies: constructDataSourceStrategies(schema, makeStrategy(MYSQL_DB_TYPE)),
+        });
+        expect(out.resolvers['Mutation.createPost.init.1.req.vtl']).toContain(
+          '$util.qr($ctx.stash.defaultValues.put("id", $util.autoId()))',
+        );
+      });
+
+      it('should include autoId when id is type Int, Float, or Boolean', async () => {
```
AaronZyLee, line 2122:
```suggestion
      it('should not include autoId when id is type Int, Float, or Boolean', async () => {
```

AaronZyLee (commented):
Could you also add the unit test for the explicit id (either of type `ID!` or `String!`) with `primaryKey`? It seems this case is missing but it should be an important one especially in the SQL where the `primaryKey` is required in the schema.

Sample schema
```
  type Post @model {
      id: ID! @primaryKey
      title: String!
  }
```

packages/amplify-graphql-model-transformer/src/__tests__/model-transformer.test.ts
dpilch: The schema for SQL is different because SQL schemas require an `@primaryKey`.

```
InvalidDirectiveError: SQL model "Post" must contain a primary key field
```

dpilch: I've just checked again to make sure there are not breaking changes. I reverted my changes, but kept the tests cases. The failing tests cases are:

* autoId › dynamodb › should include autoId when timestamps are null with explicit id
  * expected: change to timestamp: null handling
* autoId › dynamodb › should include autoId when timestamps are null with implicit id
  * expected: change to timestamp: null handling
* autoId › dynamodb › should not include autoId when id is type Int, Float, or Boolean
  * expected: change to non ID or String types
* autoId › dynamodb › should not include autoId when using a custom primary key
  * expected: Removed the auto id statement because implicit id will not be added when using primary key
* autoId › sql › should include autoId when timestamps are null
  * expected: change to timestamp: null handling
* autoId › sql › should not include autoId when id is type Int, Float, or Boolean
  * expected: change to non ID or String types

palpatim (approved):

AaronZyLee (approved):

PR #2475, opened 2024-04-19
test: add list and get E2Es for simple owner auth

palpatim:
#### Description of changes

- Added list and get cases to simple owner auth using references-style relations
- Added 'assoc-field' tests to the RUN_SOLO block so it isn't combined with any other cases. If we encounter timeouts on this file, I'll split the permutations out into separate test files that will also be solo (See https://github.com/aws-amplify/amplify-category-api/pull/2462/#discussion_r1567894253)
- Right now, tests that call `list` queries against a SQL data source fail because of a bug in the SQL lambda layer. Once we resolve that, those cases will pass.
- Addressed previous PR feedback:
  - https://github.com/aws-amplify/amplify-category-api/pull/2462/#discussion_r1567889907
  - https://github.com/aws-amplify/amplify-category-api/pull/2462/#discussion_r1567891099
  - https://github.com/aws-amplify/amplify-category-api/pull/2462/#discussion_r1567960000

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-graphql-api-construct-tests/src/__tests__/backends/configurable-stack/app.ts
```diff
@@ -76,7 +76,7 @@ new AmplifyGraphqlApi(stack, `${PREFIX}Api`, {
   authorizationModes: {
     defaultAuthorizationMode: 'AMAZON_COGNITO_USER_POOLS',
     userPoolConfig: { userPool },
-    apiKeyConfig: { expires: Duration.days(7) },
+    apiKeyConfig: { expires: Duration.days(2) },
```
palpatim, line 79: Unexpectedly, a Duration of 1 fails to deploy. Rather than spend a bunch of time troubleshooting, I went with 2, which is still less than 7. :)

packages/amplify-graphql-api-construct-tests/src/__tests__/utils.test.ts
```diff
@@ -0,0 +1,54 @@
+import { isOperationAuthInputApiKey } from '../utils';
```
palpatim, line 1: I needed to verify some predicate type behavior, so I left this in place. It only runs during actual E2E builds, so it's not much use as a unit test, but it does verify that the predicate function is working as expected.

packages/amplify-graphql-api-construct-tests/src/utils/auth.ts
```diff
@@ -114,3 +114,34 @@ export const signInCognitoUser = async (options: SignInCognitoUserInput): Promis
     username,
   };
 };
+
+export const getUsernameClaimFromJwt = (accessToken: string): string => {
```
palpatim, line 118: Added these for an aborted foray into owner==primary key testing. We don't actually support that case well, but I've left the utilities for upcoming cases.

scripts/split-e2e-tests.ts
```diff
@@ -135,15 +135,16 @@ const RUN_SOLO: (string | RegExp)[] = [
   /src\/__tests__\/sql-.*\.test\.ts/,
   // CDK tests
   /src\/__tests__\/base-cdk.*\.test\.ts/,
+  'src/__tests__/admin-role.test.ts',
```
palpatim, line 138: I sorted the CDK tests section as well as adding my new owner-auth

palpatim (commented):
marker review

atierian (approved):

AaronZyLee (approved):

palpatim:
Overriding required admin approval since the only triggering file is `split-e2e-tests.ts`, which we just set to not require admin approval in https://github.com/aws-amplify/amplify-category-api/pull/2479

PR #2481, opened 2024-04-19
feat: support multiple connection Uris for SQL databases

phani-srikar:
#### Description of changes

With this feature, we will allow the customers of Amplify GraphQL CDK construct to specify more than one SSM paths that hold the connection URI for the SQL database. This enables customers of Amplify Console to have both shared and branch specific secrets for database connection URIs.
The provided SSM paths are resolved in the order that is specified in the array by the customer and based on the availability to read them at runtime during execution of the SQL Lambda datasource.
This PR also updates the error messages to make them more readable.
The change is backwards compatible since existing customers can continue to specify the connection URI SSM path as a single string.

Added unit and E2E tests. The added E2E test runs successfully on CI.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-graphql-api-construct/src/model-datasource-strategy-types.ts
+++ b/packages/amplify-graphql-api-construct/src/model-datasource-strategy-types.ts
@@ -131,7 +131,12 @@ export type SqlModelDataSourceDbConnectionConfig =
  * @experimental
  */
 export interface SqlModelDataSourceSsmDbConnectionStringConfig {
-  /** The SSM Path to the secure connection string used for connecting to the database. **/
+  /**
+   * The SSM Path to the secure connection string used for connecting to the database. If more than one path is provided,
+   * the SQL Lambda will attempt to retrieve connection information from each path in order until it finds a valid
+   * path entry, then stop. If the connection information contained in that path is invalid, the SQL Lambda will not
+   * attempt to retrieve connection information from subsequent paths in the array.
+   **/
   readonly connectionUriSsmPath: string | string[];
 }
 
--- a/packages/amplify-graphql-api-construct/src/model-datasource-strategy-types.ts
+++ b/packages/amplify-graphql-api-construct/src/model-datasource-strategy-types.ts
@@ -132,7 +132,7 @@ export type SqlModelDataSourceDbConnectionConfig =
  */
 export interface SqlModelDataSourceSsmDbConnectionStringConfig {
   /** The SSM Path to the secure connection string used for connecting to the database. **/
-  readonly connectionUriSsmPath: string;
+  readonly connectionUriSsmPath: string | string[];
 }
 
 /**
--- a/packages/amplify-graphql-model-transformer/rds-lambda/handler.ts
+++ b/packages/amplify-graphql-model-transformer/rds-lambda/handler.ts
@@ -144,6 +144,36 @@ const getSecretManagerValue = async (secretArn: string | undefined): Promise<{ u
   }
 }
 
+const resolveConnectionStringValue = async (jsonConnectionString: string): Promise<string> => {
+  const parsedJsonConnectionString = JSON.parse(jsonConnectionString);
+  const ssmRequestError = 'Unable to connect to the database. Check the logs for more details.';
+  const ssmLoggedError = 'Unable to fetch the connection Uri from SSM for the provided paths.';
+  console.log(`connection uri SSM paths: ${parsedJsonConnectionString}`);
+  if (Array.isArray(parsedJsonConnectionString)) {
+    for (const connectionUriSsmPath of parsedJsonConnectionString) {
+      console.log(`Trying to fetch connection string from SSM path: ${connectionUriSsmPath}`);
+      try {
+        return await getSSMValue(connectionUriSsmPath);
+      }
+      catch (e) {
+        // try the next secret path;
+        continue;
+      }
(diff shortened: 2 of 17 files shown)


packages/amplify-graphql-api-construct/src/model-datasource-strategy-types.ts
```diff
@@ -132,7 +132,7 @@ export type SqlModelDataSourceDbConnectionConfig =
  */
 export interface SqlModelDataSourceSsmDbConnectionStringConfig {
   /** The SSM Path to the secure connection string used for connecting to the database. **/
```
palpatim, line 134: Add additional documentation for the resolution strategy if multiple values are specified
```suggestion
  /**
   * The SSM Path to the secure connection string used for connecting to the database. If more than one path is provided,
   * the SQL Lambda will attempt to retrieve connection information from each path in order until it finds a valid 
   * path entry, then stop. If the connection information contained in that path is invalid, the SQL Lambda will not 
   * attempt to retrieve connection information from subsequent paths in the array.
   **/
```

palpatim (approved):
Approved contingent on @atierian 's approval of public API change.

atierian (approved):

packages/amplify-graphql-model-transformer/rds-lambda/handler.ts
```diff
@@ -144,6 +144,36 @@ const getSecretManagerValue = async (secretArn: string | undefined): Promise<{ u
   }
 }
 
+const resolveConnectionStringValue = async (jsonConnectionString: string): Promise<string> => {
+  const parsedJsonConnectionString = JSON.parse(jsonConnectionString);
+  const ssmRequestError = 'Unable to connect to the database. Check the logs for more details.';
+  const ssmLoggedError = 'Unable to fetch the connection Uri from SSM for the provided paths.';
+  console.log(`connection uri SSM paths: ${parsedJsonConnectionString}`);
```
sundersc, line 151: We may need to remove this line? or print only if debug more is enabled?

packages/amplify-graphql-model-transformer/rds-lambda/handler.ts
```diff
@@ -144,6 +144,36 @@ const getSecretManagerValue = async (secretArn: string | undefined): Promise<{ u
...
+  if (Array.isArray(parsedJsonConnectionString)) {
+    for (const connectionUriSsmPath of parsedJsonConnectionString) {
+      console.log(`Trying to fetch connection string from SSM path: ${connectionUriSsmPath}`);
```
sundersc, line 154: All the console logs - let's print only if debug mode is enabled.

packages/amplify-graphql-model-transformer/rds-lambda/handler.ts
```diff
@@ -144,6 +144,36 @@ const getSecretManagerValue = async (secretArn: string | undefined): Promise<{ u
...
+    }
+    console.log(ssmLoggedError);
+    throw new Error(ssmRequestError);
+  }
+  else {
+    try {
+      return await getSSMValue(parsedJsonConnectionString);
+    }
+    catch (e) {
+      console.log(ssmLoggedError);
```
sundersc, line 171: These logs may be redundant. IIRC, exceptions are recorded in CW logs.

packages/amplify-graphql-model-transformer/rds-lambda/handler.ts
phani-srikar: This is for debugging E2E, I will remove this in next commit.

packages/amplify-graphql-model-transformer/rds-lambda/handler.ts
phani-srikar: This is for debugging E2E, I will remove this in next commit.

packages/amplify-graphql-model-transformer/rds-lambda/handler.ts
phani-srikar: These ones are intentional since the logged message has more details than the error message we send back to the GraphQL client.

sundersc: Did a quick testing, these log statements are redundant. Throwing an error automatically records the full error object as shown in the screenshot.

```ts
  const customErr = 'This is a custom error';
  console.log(customErr);
  throw new Error(customErr);
```
image

phani-srikar: That is expected. The situation here is different from above case since the message we log is different from the error we throw which also gets logged but has less information than the additionally added log.
```
const ssmRequestError = 'Unable to connect to the database. Check the logs for more details.';
const ssmLoggedError = 'Unable to fetch the connection Uri from SSM for the provided paths.';

console.log(ssmLoggedError);
throw new Error(ssmRequestError);
```
So the log message is useful since it additionally tells the developer that: We failed to connect to DB because we couldn't **fetch the connection Uri from SSM**.
The reason this isn't put behind `DEBUG_FLAG` setting is because this is a custom error and does not print out any sensitive information other than what we already expect from our public docs and saves the developer time to debug the issue without needing to update the environment variable to debug.

palpatim (approved):

AaronZyLee (approved):

PR #2516, opened 2024-04-30
fix: enable ssl for generate schema command

sundersc:
#### Description of changes

Enforce SSL for generate schema command.

NA

#### Issue #, if available
NA

- Manual test
- E2E test is in progress.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-graphql-schema-generator/src/certs/cert.ts
```diff
@@ -0,0 +1,33 @@
+import * as fs from 'fs';
+import * as path from 'path';
+
+// This logic to get the SSL config must match with the SQL Lambda Layer.
+// If you are changing this, please make sure to change the logic in the SQL Lambda Layer as well.
+export const getSSLConfig = (
```
sundersc, line 6: This logic is copied from lambda layer. For now, it is duplicated. We can think about storing this in one place in the future.

packages/amplify-graphql-schema-generator/vpc-db-lambda/src/connection.ts
```diff
@@ -7,7 +9,10 @@ export const establishDBConnection = (config: any): any => {
     port: config.port,
     user: config.username,
     password: config.password,
-    ssl: { rejectUnauthorized: false },
+    ssl: {
+      rejectUnauthorized: true,
+      ca: getRDSCertificate(),
+    },
```
sundersc, line 15: This code gets executed only if the provided host belongs to RDS. So there is no need to check if it is a RDS host or not.

packages/amplify-graphql-schema-generator/vpc-db-lambda/src/aws-rds-global-bundle.pem
```diff
@@ -0,0 +1,3052 @@
+-----BEGIN CERTIFICATE-----
```
phani-srikar, line 1: I don't recommend storing these certificates in the library code, because of the following reasons:
- If these certificates were to be refreshed, it would require **all** our customers to upgrade their dependency to the `latest` which can be painful.
- There is no edit protection for these files in terms of external contributions, so if we accidentally merge a PR that updates these, that breaks our customers.

The second can be mitigated using `CODEOWNERS` in Github, but want to know if we thought through the first.

phani-srikar (changes requested):

packages/amplify-graphql-schema-generator/src/certs/download-certs.sh
```diff
@@ -0,0 +1,26 @@
...
+#
+# RDS global bundle is used to connect directly to clusters and/or instances
+# Root certs are to connect to RDS Proxies
+CERTS=(
+https://truststore.pki.rds.amazonaws.com/global/global-bundle.pem
+https://www.amazontrust.com/repository/AmazonRootCA1.pem
+https://www.amazontrust.com/repository/AmazonRootCA2.pem
+https://www.amazontrust.com/repository/AmazonRootCA3.pem
+https://www.amazontrust.com/repository/AmazonRootCA4.pem
+https://www.amazontrust.com/repository/SFSRootCAG2.pem
```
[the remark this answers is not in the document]
sundersc: Yes, these are for RDS instances, clusters and proxies. There is an explanation on lines 8-13. Is there anything specific you are looking for?

palpatim (dismissed):

packages/amplify-graphql-schema-generator/vpc-db-lambda/src/aws-rds-global-bundle.pem
sundersc: For your first point, this file contains only the certificates provided by RDS service. They don't change often, the default provided certificate `rds-ca-rsa2048-g1` is valid for 40 years and (`rds-ca-rsa4096-g1` and `rds-ca-ecc384-g1`) are valid for 100 years.

If RDS is updating their global CA, we should be okay to push it on a next release. Since these don't change often, I don't think it is a problem we need to worry about right now.

We can consider pulling these certificates dynamically at runtime but I don't want to do this now unless we clearly understand all the side effects of that approach.

For your second point, I agree, we can add it to the `codeowners` file.

phani-srikar: > We can consider pulling these certificates dynamically at runtime but I don't want to do this now unless we clearly understand all the side effects of that approach.

Lets discuss this option and agree we can do this without breaking changes in subsequent releases.

> For your second point, I agree, we can add it to the codeowners file.

Do you want to make this change in the PR?

phani-srikar (approved):

palpatim (approved):

PR #2517, opened 2024-04-30
chore: add local publish helper for release verification against amplify-backend

palpatim:
#### Description of changes

New helper function to publish local workspace to local Verdaccio registry.

#### Checklist

- [x] PR description included

### DIFF
--- a/shared-scripts.sh
+++ b/shared-scripts.sh
@@ -164,6 +164,14 @@ function _publishToLocalRegistry {
 
     _generateChangeLog
 }
+function _publishLocalWorkspace {
+    source codebuild_specs/scripts/local_publish_helpers.sh
+    startLocalRegistry "$(pwd)/codebuild_specs/scripts/verdaccio.yaml"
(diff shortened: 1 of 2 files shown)


phani-srikar (approved):

shared-scripts.sh
```diff
@@ -164,6 +164,14 @@ function _publishToLocalRegistry {
 
     _generateChangeLog
 }
+function _publishLocalWorkspace {
```
phani-srikar, line 167: Do we want to expose a top level script in the `package.json` to be able to easily run this? Like `yarn publish-to-local`?

palpatim: I went back & forth on this and landed on "not right now." I want to see it in action a few times first, and then we can make sure that a yarn task properly encompasses all the steps. For now, it's not too burdensome; involving a couple of bash invocations rather than a yarn task.

iartemiev (approved):

PR #2523, opened 2024-05-02
fix: auth to use validateUsingSource in place of auth filter

dpilch:
#### Description of changes

Validate the result from the SQL lambda against auth rules using `validateUsingSource`. This will enable unauthorized queries to contain an unauthorized error.

Query:
```
query MyQuery {
  getPrimary(id: "b7d23954-2017-4ddf-bfc7-a0c9b9d3a371") {
    id
    owner
  }
}
```

Current behavior result:
```
{
  "data": {
    "getPrimary": null
  }
}
```

New behavior result:
```
{
  "data": {
    "getPrimary": null
  },
  "errors": [
    {
      "path": [
        "getPrimary"
      ],
      "data": null,
      "errorType": "Unauthorized",
      "errorInfo": null,
      "locations": [
        {
          "line": 2,
          "column": 3,
          "sourceName": null
        }
      ],
      "message": "Not Authorized to access getPrimary on type Query"
    }
  ]
}
```

N/A

#### Issue #, if available

N/A

* Unit tests
* E2E tests
* Manual testing

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-graphql-name-mapping-transformer/src/__tests__/__integ__/__snapshots__/with-has-many.test.ts.snap
```diff
@@ -247,9 +247,7 @@ exports[`@refersTo with @hasMany for RDS Models model table names are mapped 4`]
 #set( $lambdaInput.operationName = \\"getEmployee\\" )
 #set( $lambdaInput.args.metadata = {} )
 #set( $lambdaInput.args.metadata.keys = [] )
-#if( !$util.isNullOrEmpty($ctx.stash.authFilter) )
-  #set( $lambdaInput.args.metadata.authFilter = $ctx.stash.authFilter )
-#end
+#set( $lambdaInput.args.metadata.authFilter = {} )
```
palpatim, line 250: Do we have confidence that this change is safe? Are we sure that there will never be data in `$lambdaInput.args.metadata.authFilter`?

palpatim (approved):
LGTM overall, but one behavior confirmation requested

packages/amplify-graphql-name-mapping-transformer/src/__tests__/__integ__/__snapshots__/with-has-many.test.ts.snap
dpilch: In this case we want the authFilter to be cleared because we want the lambda to return the record to the resolver even it is not authorized. We are now using the `validateUsingSource` util function against the result to verify the authorization.

In this case there is information in the authFilter but we are shifting that information to `authRules` and `validateUsingSource`.

sundersc (approved):

PR #2536, opened 2024-05-08
fix: redact relational field in mutation & subscription based on model auth rules

AaronZyLee:
#### Description of changes
The relational field is redacted in mutation result and subsequent subscription from another client based on the comparison result of

- Read role definitions for relational field
- Read role definitions for the related model

The comparison will occur if the relational field is not protected by any field level auth rules (as the field is redacted when field level auth is added currently). The field redaction will only happen when the model subscription level is on in either case

#### Issue #, if available

N/A

* Unit tests
* E2E tests

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-graphql-relational-transformer/src/resolver/ddb-generator.ts
```diff
@@ -185,7 +185,7 @@ export class DDBRelationalResolverGenerator extends RelationalResolverGenerator
       MappingTemplate.s3MappingTemplateFromString(
         print(
           compoundExpression([
-            iff(ref('ctx.stash.deniedField'), raw('#return($util.toJson(null))')),
+            iff(ref('ctx.stash.deniedField'), compoundExpression([set(ref('result'), obj({ items: list([]) })), raw('#return($result)')])),
```
AaronZyLee, line 188: For a `hasMany` relation, when the field is redacted, the `{ item: [] }` should be returned instead of a `null` value. Otherwise the following error will be thrown by AppSync
Screenshot 2024-05-08 at 10 59 03 AM

packages/amplify-graphql-auth-transformer/src/__tests__/relational-field-redaction.test.ts
```diff
@@ -0,0 +1,263 @@
...
+  };
+  return expectation;
+};
+
+export type TestTableRow = [string, string[], string[], boolean];
+
+/**
+ * Primary auth rules - Related auth rules - Redact relational field
+ */
+const testCases: [string[], string[], boolean][] = [
```
AaronZyLee, line 213: These test cases are the main part of this unit test, which defines the different auth rules on both relation models and whether to redact the relational field. These cases are tested below in the dimension of data source types and relation types(field or reference)

packages/amplify-graphql-relational-transformer/src/resolver/ddb-generator.ts
palpatim: Thanks for the explanation. for simplicity, could you do ?

```
raw('#return($util.toJson([]))'))
```

packages/amplify-graphql-auth-transformer/src/graphql-auth-transformer.ts
```diff
@@ -716,55 +716,133 @@ export class AuthTransformer extends TransformerAuthBase implements TransformerA
...
+        }
+        // relational field read roles are already processed with filter in parent call
+        const fieldReadRoleDefinitions = fieldRoles.map((r) => this.roleMap.get(r)!);
+        /**
+         * Compare if two auth roles are identical
+         * @param role1 auth role 1
+         * @param role2 auth role 2
+         * @returns boolean for two roles are identical
+         */
+        const isIdenticalAuthRole = (role1: RoleDefinition, role2: RoleDefinition): boolean =>
```
palpatim, line 767: nit: can you add this as a utliity near the RoleDefinition type itself, and add some unit tests to catch edge cases. Especially to make sure we handle undefined elements in one or both operands.

packages/amplify-graphql-auth-transformer/src/graphql-auth-transformer.ts
```diff
@@ -716,55 +716,133 @@ export class AuthTransformer extends TransformerAuthBase implements TransformerA
...
+         * @param fieldRole relational field auth role
+         * @param relatedModelRoles related model auth roles
+         * @returns boolean for the field can access both sides or not
+         */
+        const isFieldRoleHavingAccessToBothSide = (fieldRole: RoleDefinition, relatedModelRoles: RoleDefinition[]): boolean =>
+          relatedModelRoles.some(
+            (relatedRole) =>
+              isIdenticalAuthRole(relatedRole, fieldRole) ||
+              (relatedRole.provider === fieldRole.provider &&
+                (relatedRole.provider === 'userPools' || relatedRole.provider === 'oidc') &&
```
palpatim, line 786: 1. Wouldn't `identityPools` be a valid provider for this case?
2. As above, put this utility in RoleDefinition & make sure we have plenty of tests for it

packages/amplify-graphql-auth-transformer/src/graphql-auth-transformer.ts
```diff
@@ -716,55 +716,133 @@ export class AuthTransformer extends TransformerAuthBase implements TransformerA
...
+                (fieldRole.strategy === 'private' || relatedRole.strategy === 'private')),
+          );
+        /**
+         * Determine the given auth role is either dynamic auth role or custom auth role
+         * These auth client authorization depends on the dynamic field or custom auth logic apart from the auth type
+         * Which authenticates differently based on the given dynamic field value
+         * @param role auth role definition
+         * @returns boolean for the auth role is a dynamic or custom auth role
+         */
+        const isDynamicAuthOrCustomAuth = (role: RoleDefinition): boolean =>
```
palpatim, line 796: as above, put this utility in RoleDefinition & make sure we have plenty of tests for it

packages/amplify-graphql-auth-transformer/src/graphql-auth-transformer.ts
```diff
@@ -716,55 +716,133 @@ export class AuthTransformer extends TransformerAuthBase implements TransformerA
...
+         * Loop through the field read role
+         * Once there is one role detected to have access on both side, the auth role definitions will be compared to determine whether
+         * to redac the field or not
+         */
+        for (let fieldRole of fieldReadRoleDefinitions) {
+          // When two role definitions have an overlap
+          if (isFieldRoleHavingAccessToBothSide(fieldRole, filteredRelatedModelReadRoleDefinitions)) {
+            // Check if two role definitions are identical without dynamic auth role or custom auth role
+            // If not, redact the relational field
+            redactRelationalField = !(
```
palpatim, line 808: This filter is pretty hard to parse (a negation of a multi-layer boolean test that includes 2 loops). Can you separate it out for readibility?

packages/amplify-graphql-auth-transformer/src/__tests__/relational-field-redaction.test.ts
```diff
@@ -0,0 +1,263 @@
...
+    owner: String
+    owners: [String]
+    singleGroup: String
+    groups: [String]
+    relatedOnePrimaryId: ID
+    primary: Primary @belongsTo(fields: ["relatedOnePrimaryId"])
+  }
+`;
+
+const referenceSchemaTempalte = `
```
palpatim, line 105: nit
```suggestion
const referenceSchemaTemplate = `
```

packages/amplify-graphql-auth-transformer/src/__tests__/relational-field-redaction.test.ts
```diff
@@ -0,0 +1,263 @@
...
+    const expectation = makeTransformationExpectation(ddbDataSourceStrategies, referenceSchemaTempalte);
+    const testTable: TestTableRow[] = [];
+    for (const strategyName of Object.keys(ddbDataSourceStrategies)) {
+      testCases.forEach((testCase) => {
+        testTable.push([strategyName, ...testCase] as TestTableRow);
+      });
+    }
+    test.each(testTable)('%s - Primary %s - Related %s should redact relational field - %s', expectation);
+  });
+  describe('RDS datasources - reference relation', () => {
```
palpatim, line 253: It "shouldn't" make a difference, but we should test heterogeneous relationships as well (DDB Primary, SQL Related; and SQL Primary, DDB Related)

packages/amplify-graphql-relational-transformer/src/resolver/ddb-generator.ts
sundersc: Is it because the related field is not null. I believe we don't this today if the auth rules on the relational field is restrictive than model rules.

palpatim (commented):
As noted, please add e2es that verify that we do not redact if subscriptions are turned off

palpatim (approved):

phani-srikar (approved):

PR #2541, opened 2024-05-09
fix: added @default schema validations

bzsurbhi:
#### Description of changes
Adde schema validations for
- @default directive
    - The @default directive may only be added to object definitions annotated with @model
    - The @default directive may only be added to scalar or enum field types
    - The @default directive only takes a value property
    - String cannot represent a non string value: the @default directive has a non String value

- ownerField should be of type String

N/A

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-schema-validator/src/__tests__/schemas/valid-owner-field-type-string.graphql
```diff
@@ -0,0 +1,7 @@
+type Inicio
+  @model
+  @auth(
+    rules: [{ allow: owner, operations: [create, delete, update, read], provider: userPools, ownerField: "autores", groups: ["MeuCPD"] }]
```
sundersc, line 4: Have you intentionally added `groups` for owner rule? Are there any validations in transformers to stop this?

bzsurbhi: No, one of our customers had this schema. So just used the same schema for our test. The validation on our end only checks for ownerField type
I'm not sure if any validations exists on the data side to stop this.

sundersc (approved):

phani-srikar (approved):

dpilch (approved):

sundersc (approved):

phani-srikar (approved):

PR #2548, opened 2024-05-10
fix: yarn clean preserve cert file in schema discovery lambda

sundersc:
`yarn clean` is removing the SQL cert file because it is placed in `lib` folder.

This PR moves it to `certs` directory under `src` directory and updates the build script to copy it before bundling the zip file.


phani-srikar (approved):

packages/amplify-graphql-schema-generator/vpc-db-lambda/src/connection.ts
```diff
@@ -37,5 +37,6 @@ export const establishDBConnection = (config: any): any => {
 const getRDSCertificate = (): string => {
   // This certificate file is copied from the parent folder `amplify-graphql-schema-generator/certs/aws-rds-global-bundle.pem`.
   const RDS_CERT_FILE_NAME = 'aws-rds-global-bundle.pem';
-  return fs.readFileSync(RDS_CERT_FILE_NAME, 'utf-8');
+  const certPath = `${__dirname}/certs/${RDS_CERT_FILE_NAME}`;
```
palpatim, line 40: Is this going to cause any problems on Windows where `__dirname` may have backslashes instead of forward slashes? I understand that Node uses forward slashes transparently; I'm asking specifically about a mixed-mode path. Why not `path.join`?

palpatim (approved):

packages/amplify-graphql-schema-generator/vpc-db-lambda/src/connection.ts
sundersc: Updated.

phani-srikar (approved):

PR #2587, opened 2024-05-31
fix: fix selection sets for SQL models with optional dependencies

palpatim:
#### Description of changes

This PR changes the resolution of null reference field values from empty string (`""`) to null for relationships involving SQL data sources.

In https://github.com/aws-amplify/amplify-category-api/issues/2580, we see that defaulting to an empty string breaks non-string primary keys like UUIDs, because the database engine performs validation of the input and returns an error. By letting the value pass through as null, the primary key comparison becomes an `IS NULL` as opposed to `= ""`, which passes validation just fine, even if though the primary key can never be null.

#### Issue #, if available

https://github.com/aws-amplify/amplify-category-api/issues/2580

New E2Es & updated snapshots.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] [E2E tests](https://us-east-1.console.aws.amazon.com/codesuite/codebuild/594813022831/projects/amplify-category-api-e2e-workflow/batch/amplify-category-api-e2e-workflow:fb5aa8c5-f908-4463-b105-ab5fb76e3733?region=us-east-1) pass
  - `verify_api_extract` failure in the E2E run fixed in [9ad51a9](https://github.com/aws-amplify/amplify-category-api/pull/2587/commits/9ad51a912368158455ea52939f52708e51a184cc)


palpatim (commented):
marker review

phani-srikar (dismissed):

phani-srikar (approved):

PR #2614, opened 2024-06-10
ci: don't use me-south-1 region for e2e tests when using beta layer

dpilch:
#### Description of changes

Some regions may not have the beta lambda layer deployed. Currently this is just `me-south-1`. The E2E tests should not use this region when using the beta lambda layer. Technically this is only necessary for tests that use SQL, but it would be more difficult/error prone to maintain a list of tests that this applies to.

This change will filter out `me-south-1` as a candidate region for tests. Run `yarn split-e2e-tests-beta` to remove `me-south-1` as an available region.

N/A

#### Issue #, if available

N/A

* `yarn split-e2e-tests-beta` and then `git diff HEAD`.
* E2E tests: https://us-east-1.console.aws.amazon.com/codesuite/codebuild/594813022831/projects/amplify-category-api-e2e-workflow/batch/amplify-category-api-e2e-workflow%3Ab5128979-17d9-4848-9ccc-52d3b72cc700?region=us-east-1

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


scripts/split-e2e-tests.ts
```diff
@@ -7,18 +7,23 @@ type TestRegion = {
   name: string;
   optIn: boolean;
   cognitoSupported: boolean;
+  betaLayerDeployed: boolean; // is the beta layer deployed in this region
```
palpatim, line 10: nit: The comment is redundant with the well-named variable :)

```suggestion
  betaLayerDeployed: boolean;
```

palpatim (approved):
non-blocking nit. If you get another review that requires you to edit this PR, please remove the comment as noted, otherwise, feel free to ship & remove the comment in a future PR

scripts/split-e2e-tests.ts
```diff
@@ -245,7 +255,8 @@ const splitTests = (baseJobLinux: any, testDirectory: string, pickTests?: (testS
 
       if (RUN_SOLO.find((solo) => test === solo || test.match(solo))) {
         if (RUN_IN_ALL_REGIONS.find((allRegionsTest) => test === allRegionsTest || test.match(allRegionsTest))) {
-          const candidateRegions = filterCandidateRegions(test, testRegions);
+          // always run these jobs in regions that do not have the beta layer deployed
+          const candidateRegions = filterCandidateRegions(test, testRegions, false);
```
phani-srikar, line 259: The canaries run solo across regions and the SQL ones might fail if run in non-supported beta regions

dpilch: The canaries do not use the lambda layer so they will pass fine in this region. We can see successfull runs here: 
* https://us-east-1.console.aws.amazon.com/codesuite/codebuild/594813022831/projects/amplify-category-api-e2e-workflow/build/amplify-category-api-e2e-workflow%3A3c17047b-e41c-433f-9d6f-86be971fdd9a?region=us-east-1
* https://us-east-1.console.aws.amazon.com/codesuite/codebuild/594813022831/projects/amplify-category-api-e2e-workflow/build/amplify-category-api-e2e-workflow%3Aff30ddf6-99cd-4b84-b1b2-bf0926e207a4?region=us-east-1

phani-srikar (approved):

PR #2634, opened 2024-06-18
feat: import existing table to amplify managed table

dpilch:
#### Description of changes

Add a feature to import existing DynamoDB tables to an Amplify managed table. Import validation will come with a later PR.

* New data source strategy: `IMPORTED_AMPLIFY_TABLE`.
* New construct prop: `importedAmplifyDynamoDBTableMap`.
* Final naming on these can be updated before merging to main.

`importedAmplifyDynamoDBTableMap` will define a mapping of model name to imported table name. All models defined with the `IMPORTED_AMPLIFY_TABLE` should have a matching key in the `importedAmplifyDynamoDBTableMap` prop.

```typescript
import * as cdk from "aws-cdk-lib";
import { Construct } from "constructs";
import {
  AmplifyGraphqlApi,
  AmplifyGraphqlDefinition,
} from "@aws-amplify/graphql-api-construct";
import * as path from "path";

export class BackendStack extends cdk.Stack {
  constructor(scope: Construct, id: string, props?: cdk.StackProps) {
    super(scope, id, props);

    const amplifyApi = new AmplifyGraphqlApi(this, "MyNewApi", {
      definition: AmplifyGraphqlDefinition.combine([
        AmplifyGraphqlDefinition.fromString(
          `
            type Todo @model @auth(rules: [{ allow: public }]) {
              title: String
            }
          `,
          {
            dbType: "DYNAMODB",
            provisionStrategy: "DEFAULT",
          },
        ),
        AmplifyGraphqlDefinition.fromString(
          `
            type Blog @model @auth(rules: [{ allow: public }]) {
              title: String
              content: String
              authors: [String]
            }
          `,
          {
            dbType: "DYNAMODB",
            provisionStrategy: "IMPORTED_AMPLIFY_TABLE",
          },
        ),
      ]),
      importedAmplifyDynamoDBTableMap: {
        // existing table
        Blog: 'Blog-xxxx-NONE',
      },
      authorizationModes: {
        defaultAuthorizationMode: "API_KEY",
        apiKeyConfig: {
          expires: cdk.Duration.days(30),
        },
      },
    });
  }
}
```

#### Issue #, if available

N/A

* Manual testing
* Unit tests, more to come in future PRs
* This is merging to a feature branch and E2E tests will be added with a separate PR

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-graphql-api-construct/src/types.ts
```diff
@@ -752,6 +752,12 @@ export interface AmplifyGraphqlApiProps {
    * For more information, refer to https://docs.amplify.aws/lib/datastore/getting-started/q/platform/js/
    */
   readonly dataStoreConfiguration?: DataStoreConfiguration;
+
+  // TODO: decide final naming before merging to main
+  /**
+   * The table map for the imported tables. The key is the model type name defined in schema; the value is the table name of the existing table
```
palpatim, line 758: 1. This can happen in a followup PR, but let's make sure to highlight some of the major constraints. Off the top of my head: 1/ the existing table must have been created by Amplify Gen1, 2/ the existing table must be in the same account and region as the API.
2. For parallelism with the existing `stackMappings` input property, we could consider a name like `importedAmplifyDynamoDBTableMapping`. I'm not 100% sold on that though, so it's just a suggestion to consider for naming review.
3. While we're in this file (again can happen in a followup PR), can you correct the typo in the comment on L746? `If no outputStorageStrategey` -> `If no outputStorageStrategy`

packages/amplify-graphql-model-transformer/src/__tests__/amplify-dynamodb-table-generator.test.ts
```diff
@@ -3,10 +3,11 @@ import {
   DDB_AMPLIFY_MANAGED_DATASOURCE_STRATEGY,
   DDB_DEFAULT_DATASOURCE_STRATEGY,
   validateModelSchema,
+  IMPORTED_DDB_AMPLIFY_MANAGED_DATASOURCE_STRATEGY,
```
palpatim, line 6: very nit: sort import

packages/amplify-graphql-api-construct/src/__tests__/internal/imported-tables.test.ts
```diff
@@ -0,0 +1,125 @@
...
+        `,
+        {
+          dbType: 'DYNAMODB',
+          provisionStrategy: 'IMPORTED_AMPLIFY_TABLE',
+        },
+      );
+      expect(() => validateImportedTableMap(definition)).toThrow('Table mapping is missing for imported Amplify DynamoDB table strategy.');
+    });
+  });
+});
```
palpatim, line 125: We should also have multi-strategy tests. One case should ensure that a model that uses a different strategy throws an error if it is included in the `importedAmplifyDynamoDBTableMap`. Ideally, the error would note that the strategy is invalid, rather than a generic "missing" or "invalid" error. E.g., from your generator test below:

```typescript
dataSourceStrategies: {
        Comment: DDB_DEFAULT_DATASOURCE_STRATEGY,
        Post: DDB_AMPLIFY_MANAGED_DATASOURCE_STRATEGY,
        Author: IMPORTED_DDB_AMPLIFY_MANAGED_DATASOURCE_STRATEGY,
      },
      importedAmplifyDynamoDBTableMap: {
        Author: 'Author-myApiId-myEnv',
      },
```

If you add `Post` to the map, we'd hope to see an error like `"Table mapping includes models that do not use an imported strategy (Post)."` or similar

packages/amplify-graphql-api-construct/src/internal/imported-tables.ts
```diff
@@ -0,0 +1,37 @@
...
+    }
+    const missingModels = importedModels.filter((model) => !modelsInMap.includes(model));
+    if (missingModels.length > 0) {
+      throw new Error(`Cannot find imported Amplify DynamoDB table mapping for models ${missingModels.join(', ')}.`);
+    }
+  }
+};
+
+/**
+ * Type predicate that returns true if `obj` is a AmplifyDynamoDbModelDataSourceStrategy
```
palpatim, line 27:
```suggestion
 * Type predicate that returns true if `obj` is an ImportedAmplifyDynamoDbModelDataSourceStrategy
```

packages/amplify-graphql-model-transformer/src/graphql-model-transformer.ts
```diff
@@ -871,6 +874,12 @@ export class ModelTransformer extends TransformerModelBase implements Transforme
       generator = this.resourceGeneratorMap.get(ITERATIVE_TABLE_GENERATOR);
     } else if (isSqlStrategy(strategy)) {
       generator = this.resourceGeneratorMap.get(SQL_LAMBDA_GENERATOR);
+    } else if (isImportedAmplifyDynamoDbModelDataSourceStrategy(strategy)) {
+      if (ctx.importedAmplifyDynamoDBTableMap && ctx.importedAmplifyDynamoDBTableMap[typeName]) {
```
palpatim, line 878: I'd prefer to handle the validations as invariants to the various checks we have to scatter through the code. E.g., when we introduced SQL support, we made strong assertions that every single model would have a DataSourceStrategy, and wrote our `is*` checks with that in mind.

What conditions would lead to this code being invoked with an invalid `importedAmplifyDynamoDBTableMap`? How can we catch those earlier so we can build in the strong assumptions through the rest of the code?

Perhaps a different internal structure work better: e.g., the imported strategy definition must be created with the mapped table name, so:

```typescript
export interface ImportedAmplifyDynamoDbModelDataSourceStrategy {
  readonly dbType: 'DYNAMODB';
  readonly provisionStrategy: 'IMPORTED_AMPLIFY_TABLE';
  readonly tableName: string;
}
```

Thus no import strategy could even be created without a valid mapping.

That might not be appropriate for the _public_ interface, if we want to allow a batch-oriented DX, but it would be completely appropriate for the _internal_ interface, where we want that strong consistency guarantee. But my initial thought is that even the public interface should be specified this way. That would be consistent with how we require SQL data sources to be specified--the strategy contains db connection information & db type, fully self-contained.

packages/amplify-graphql-model-transformer/src/resources/amplify-dynamodb-table/amplify-dynamo-model-resource-generator.ts
```diff
@@ -149,7 +154,9 @@ export class AmplifyDynamoModelResourceGenerator extends DynamoModelResourceGene
   protected createModelTable(scope: Construct, def: ObjectTypeDefinitionNode, context: TransformerContextProvider): void {
     const modelName = def!.name.value;
     const tableLogicalName = ModelResourceIDs.ModelTableResourceID(modelName);
-    const tableName = context.resourceHelper.generateTableName(modelName);
+    const importedTableMap = context.importedAmplifyDynamoDBTableMap ?? {};
+    const isTableImported = importedTableMap[modelName] !== undefined;
```
palpatim, line 158: Piling onto my above comments -- here you check for the existence of the importedTableMap, but _don't_ validate the strategy. If we have conditions that can result in inconsistency b/t strategy & mapping, this is a source of potential future bugs.

packages/amplify-graphql-model-transformer/src/resources/amplify-dynamodb-table/amplify-table-manager-lambda/amplify-table-manager-handler.ts
```diff
@@ -130,6 +130,27 @@ const processOnEvent = async (event: AWSCDKAsyncCustomResource.OnEventRequest):
...
+      if (tableDef.isImported) {
+        // TODO: Add import validation
+        console.log('Initiating table import process');
+        console.log('Fetching current table state');
+        console.log(`Table name: ${tableDef.tableName}`);
+        const describeTableResult = await ddbClient.describeTable({ TableName: tableDef.tableName });
+        if (!describeTableResult.Table) {
+          throw new Error(`Could not find ${tableDef.tableName} to update`);
+        }
+        log('Current table state: ', describeTableResult);
```
palpatim, line 142: Any sensitive data to worry about in the `describeTableResult`? I'm assuming not, but let's be careful about what we log.

packages/amplify-graphql-model-transformer/src/resources/amplify-dynamodb-table/amplify-table-manager-lambda/amplify-table-manager-handler.ts
```diff
@@ -130,6 +130,27 @@ const processOnEvent = async (event: AWSCDKAsyncCustomResource.OnEventRequest):
   let result;
   switch (event.RequestType) {
     case 'Create':
+      if (tableDef.isImported) {
```
palpatim, line 133: 1. rather than inlining this big `if` block, can we refactor to a standalone function to make the "Create" case a bit easier to parse?
2. I don't see anything to suggest otherwise, but please confirm that this won't cause any issues with the Tagging support for Amplify-managed tables that we're introducing in https://github.com/aws-amplify/amplify-category-api/pull/2694

packages/amplify-graphql-model-transformer/src/resources/amplify-dynamodb-table/amplify-table-manager-lambda/amplify-table-manager-handler.ts
```diff
@@ -130,6 +130,27 @@ const processOnEvent = async (event: AWSCDKAsyncCustomResource.OnEventRequest):
...
+        // TODO: Add import validation
```
palpatim, line 134: What validation needs to be added?

dpilch: The validation is to ensure the DDB properties align (Example: KeySchema, SSESpecification, etc.). WIP PR: https://github.com/aws-amplify/amplify-category-api/pull/2696

packages/amplify-graphql-model-transformer/src/resources/amplify-dynamodb-table/amplify-table-manager-lambda/amplify-table-manager-handler.ts
dpilch: 1. Will do.
2. I don't see any issues here, but I'll add a test to cover this case before merging to main.

palpatim (approved):

packages/amplify-graphql-model-transformer/src/__tests__/amplify-dynamodb-table-generator.test.ts
```diff
@@ -107,7 +106,29 @@ describe('ModelTransformer:', () => {
...
+      schema: validSchema,
+      transformers: [new ModelTransformer()],
+      dataSourceStrategies: {
+        Post: {
+          dbType: 'DYNAMODB' as const,
+          provisionStrategy: 'IMPORTED_AMPLIFY_TABLE' as const,
+        },
+      },
+    };
+    // @ts-expect-error
```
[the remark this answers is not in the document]
dpilch: Yes, updated to clarify

packages/amplify-graphql-transformer-core/src/utils/model-datasource-strategy-utils.ts
```diff
@@ -60,6 +61,21 @@ export const getModelDataSourceStrategy = (ctx: DataSourceStrategiesProvider, ty
...
+ */
+export const isImportedAmplifyDynamoDbModelDataSourceStrategy = (
+  strategy: ModelDataSourceStrategy,
+): strategy is ImportedAmplifyDynamoDbModelDataSourceStrategy => {
+  return (
+    isDynamoDbType(strategy.dbType) &&
+    typeof (strategy as any)['provisionStrategy'] === 'string' &&
+    (strategy as any)['provisionStrategy'] === 'IMPORTED_AMPLIFY_TABLE' &&
+    typeof (strategy as any)['tableName'] === 'string' &&
+    (strategy as any)['tableName'] !== ''
```
phani-srikar, line 75: Should we turn this "empty imported table name" scenario handling into a validation check and fail instead of treating this as a greenfield scenario?

packages/amplify-graphql-model-transformer/src/resources/amplify-dynamodb-table/amplify-table-manager-lambda/amplify-table-manager-handler.ts
```diff
@@ -1282,4 +1286,31 @@ const retry = async <T>(
...
+
+/**
+ * Imports an existing DDB table.
+ * @param tableDef Tabel definition of imported table.
+ */
+const importTable = async (tableDef: CustomDDB.Input): Promise<AWSCDKAsyncCustomResource.OnEventResponse> => {
+  // TODO: Add import validation
+  console.log('Initiating table import process');
+  console.log('Fetching current table state');
+  console.log(`Table name: ${tableDef.tableName}`);
```
phani-srikar, line 1298: minor nit: (since you already have a TODO up there):
Can combine above two lines to `console.log('Fetching current state of table ${tableDef.tableName}');` to make it more readable in the logs.

dpilch: I'll update this in the next PR.

packages/amplify-graphql-transformer-core/src/utils/model-datasource-strategy-utils.ts
dpilch: This would actually fail with `No resource generator assigned for Post with dbType DYNAMODB`. The error message could be improved though. I can tackle this in the next PR.

phani-srikar (approved):

palpatim (approved):

PR #2646, opened 2024-06-20
fix: add warning message for related auth subscriptions

dpilch:
#### Description of changes

Add warning message to required relational fields with differing auth roles that the subscriptions will inherit the related auth.

N/A

#### Issue #, if available

N/A

* Unit testing

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


palpatim (approved):
Approved contingent on doc update

sundersc (changes requested):
This behavior would add more confusion to what the new FF is going to do. Can we discuss the other options? Since we are planning to release the FF fix with a major version bump, we can use this opportunity to handle this as build time validation and prompt to change the relational fields to optional if we can't correctly redact.

PR #2666, opened 2024-06-26
feat(function-transformer): add support for event invocation types

atierian:
## Description of changes
Adds support for [AppSync asynchronous Lambda function invocations](https://aws.amazon.com/about-aws/whats-new/2024/05/aws-appsync-events-asynchronous-lambda-function-invocations/) via the `@function` directive.

### API Changes
```diff
directive @function(
  name: String!,
  region: String,
  accountId: String,
+  invocationType: InvocationType = RequestResponse
)

+ enum InvocationType { 
+   RequestResponse
+   Event
+ }
```

#### `appendError` vs `error`
The response mapping template for `Event` invocation types uses `util.appendError` rather than `util.error` used by `RequestResponse` (default) invocation types.

`util.error` has always been the behavior of the `@function` directive. I was unable to find any justification for the decision in the original PR:
- https://github.com/aws-amplify/amplify-cli/pull/1346

`util.error` immediately exits preventing any further execution in the response (also request but n/a here) mapping template **and** subsequent resolvers in the same pipeline. This is reasonable in the context of single function resolvers, but is unnecessarily constraining in a multi-function pipeline -- using `util.appendError` could¹ allow the subsequent Lambda function to make a handling decision for the previous error based on business logic.

Changing this behavior for existing `RequestResponse` (default) `@function` directives would be a breaking change. However, we have an opportunity to move in a more flexible direction for the newly introduced `Event` based `@function` directives. We're using `util.appendError` in the response mapping template of `Event` `@function` directives for two reasons:
1. **It allows us to return the `EventInvocationResponse` regardless of whether the Lambda function execution failed or not.**
```json
{ 
  "data": { 
    "myMutation": { 
      "success": true
    }
  }
}
```
```json
{ 
  "data": { 
    "myMutation": { 
      "success": false
    }
  },
  "errors": [
    {  }
  ]
}
```
For comparison, if `Event` based `@function` directives used `util.error`, this would be the failure response:
```json
{ 
  "data": { 
    "myMutation": null
  },
  "errors": [
    {  }
  ]
}
```

2. **It allows execution of subsequent resolvers in the pipeline.**

¹ This would require passing `ctx.error` in the Lambda invocation payload; this is **not** done today and is not included in this PR.

While we're on the topic, I'm going to document the behavior of `util.error` and `util.appendError` in pipeline resolvers for future reference.

- `util.error` immediately exits, returning the error and only the error to the caller. Neither subsequent lines in the request/response mapping template nor subsequent resolvers in a pipeline are executed after `util.error`
- `util.appendError` continues execution and returns the appended error in the `errors` block of the request.
- Appending an error using `util.appendError`, then calling `util.error` results in the appended error being **excluded** in the `errors` block -- only the error passed to `util.error` is returned in the `errors` block.
 
### Response Types
Using `@function(name: 'name', invocationType: Event)` requires that the response type of that field is `EventInvocationResponse`, which must be explicitly defined in the schema as:
```gql
type EventInvocationResponse {
  success: Boolean!
}
```
This applies when the `Event` invocation type `@function` directive is the final `@function` directive defined on that field. For example:
```gql
type Mutation {
  asyncStuff(msg: String): EventInvocationResponse @function(name: "stuff", invocationType: Event)
}
```
```gql
type Mutation {
  asyncStuff(msg: String): EventInvocationResponse  
  @function(name: "stuff")
  @function(name: "other-stuff", invocationType: Event)
}
```
But **not** when a 'RequestResponse' invocation is the final `@function` directive. This is fine:
```gql
type Mutation {
  asyncStuff(msg: String): String
  @function(name: "stuff", invocationType: Event)
  @function(name: "other-stuff")
}
```

### Constraints
`Event` invocation type `@function` directives can only be used on `Mutation` or `Query` fields. This is is more constraining than `RequestResponse` (default) invocation types.

- Snapshot testing.
- Manually deployed and testing GraphQL API.

Note: This PR is into the `feature/async-lambda` branch where a tagged release will be cut. A full E2E test run will happen prior to merging this to `main`.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-graphql-function-transformer/src/graphql-function-transformer.ts
```diff
@@ -195,3 +255,256 @@ const lambdaArnKey = (name: string, region?: string, accountId?: string): string
...
+};
+
+/**
+ * Use this to generate a field description for contextual error messages in the form of:
+ * `field-name(optional-args): return-type `@function`(name: name, invocationType: invocation-type)`
+ * @param config the {@link FunctionDirectiveConfiguration} of the `@function` directive.
+ * @param field the {@link FieldDefinitionNode} on which the `@function` directive is defined.
+ * @returns a description of the field with `@function` directive appropriate for use in error messages.
+ */
+const fieldDebugDescription = (config: FunctionDirectiveConfiguration, field: FieldDefinitionNode): string => {
```
atierian, line 485: We may want to consider adding this to `amplify-graphql-transformer-core/utils`. If so, I'll take that on in a follow up.

packages/amplify-graphql-function-transformer/src/graphql-function-transformer.ts
```diff
@@ -178,6 +193,51 @@ export class FunctionTransformer extends TransformerPluginBase {
...
+      ]),
+    );
+  }
+
+  /*
+    #if( $ctx.error )
+      $util.error($ctx.error.message, $ctx.error.type)
+    #end
+    $util.toJson($ctx.result)
+  */
```
palpatim, line 232: nit for future PR: remove or explain

packages/amplify-graphql-function-transformer/src/graphql-function-transformer.ts
```diff
@@ -178,6 +193,51 @@ export class FunctionTransformer extends TransformerPluginBase {
...
+      #set( $success = true )
+      #if( $ctx.error )
+        $util.appendError($ctx.error.message, $ctx.error.type)
+        #set( $success = false )
+      #end
+      #set( $response = {
+        "success": $success
+      } )
+      $util.toJson($response)
+    */
```
palpatim, line 214: nit for future PR: remove or explain

palpatim (approved):
minor nits about stray comments, otherwise LGTM

packages/amplify-graphql-function-transformer/src/graphql-function-transformer.ts
atierian: I find these comment blocks helpful to understand the actual request/response mapping templates we're generating (much easier to read than the VTL helpers) within the context of the transformer logic, i.e. without having to peruse test cases to find applicable snapshots. But I'm fine removing them if you feel they just add noise.

sundersc (approved):

PR #2668, opened 2024-06-27
feat: get datasource map for migration

dpilch:
#### Description of changes

* Add transformer option `enableGen2Migration`.
    * Set dynamodb table data source removal policy to retain when true.
    * Output a mapping of model name to table name when true.
    * This will be exposed as a feature flag to gen 1 customers.

New behavior is not exposed through the construct because migration is not supported from construct to gen 2.

#### Issue #, if available

N/A

* Add transformer option `enableGen2Migration`.
    * Set dynamodb table data source removal policy to retain when true.
    * Output a mapping of model name to table name when true.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-graphql-model-transformer/src/graphql-model-transformer.ts
```diff
@@ -323,9 +324,31 @@ export class ModelTransformer extends TransformerModelBase implements Transforme
...
+    const dataSourceMapping = {};
     this.resourceGeneratorMap.forEach((generator) => {
       generator.generateResources(context);
+      Object.assign(
+        dataSourceMapping,
+        Object.fromEntries(
+          Object.entries(generator.getDatasourceMap())
+            .filter(([, datasource]) => datasource.ds.type === 'AMAZON_DYNAMODB')
+            .map(([modelName, datasource]) => {
+              if (datasource.ds.dynamoDbConfig && !cdk.isResolvableObject(datasource.ds.dynamoDbConfig)) {
```
dpilch, line 336: I'm not sure there is a case when `isResolvableObject` would be true.

palpatim: Why is the check necessary? My read of the this (which might be wrong--see comment above) is that we're trying to get to the table name, which might not be resolved at synth time? What happens if the DS is a resolvable--will we skip resolver generation? How would a customer resolve in that situation?

packages/amplify-graphql-model-transformer/src/graphql-model-transformer.ts
```diff
@@ -323,9 +324,31 @@ export class ModelTransformer extends TransformerModelBase implements Transforme
   };
 
   generateResolvers = (context: TransformerContextProvider): void => {
+    const dataSourceMapping = {};
     this.resourceGeneratorMap.forEach((generator) => {
       generator.generateResources(context);
+      Object.assign(
```
palpatim, line 330: Rather than a chain of nested assigns & loops & filters, could you break this out so it's a bit easier to understand the logic here? Reading through this I'm not quite sure what's supposed to be happening in this block.

packages/amplify-graphql-model-transformer/src/graphql-model-transformer.ts
```diff
@@ -323,9 +324,31 @@ export class ModelTransformer extends TransformerModelBase implements Transforme
...
+            }),
+        ),
+      );
     });
+    if (context.transformParameters.enableGen2Migration && context.transformParameters.enableTransformerCfnOutputs) {
+      const { scope } = context.stackManager;
+      new cdk.CfnOutput(cdk.Stack.of(scope), 'DataSourceMappingOutput', {
+        value: cdk.Stack.of(scope).toJsonString(dataSourceMapping),
+        description: 'Mapping of model name to data source table name.',
+        exportName: cdk.Fn.join(':', [cdk.Aws.STACK_NAME, 'DataSourceMappingOutput']),
```
palpatim, line 349: Let's review the name of this output, since it's a new public API (my first thought is that it's too broad--it's specifically a mapping of Amplify managed tables, and possibly just of imported tables?

packages/amplify-graphql-model-transformer/src/graphql-model-transformer.ts
dpilch: The check is necessary for TypeScript here. `datasource.ds.dynamoDbConfig` is of type `DynamoDBConfigProperty | IResolvable`. 

Since we are creating the this with `AppSync.DataSource` ourselves we should know that it will always be `DynamoDBConfigProperty` and not `IResolvable`.

A more fault tolerant option would be to just print a warning for any table names that could not be resolved for whatever reason. The customer would then add this information to the mapping themselves in their Gen 2 project.

packages/amplify-graphql-api-construct/src/amplify-graphql-api.ts
```diff
@@ -211,6 +211,8 @@ export class AmplifyGraphqlApi extends Construct {
       transformParameters: {
         ...defaultTranslationBehavior,
         ...(translationBehavior ?? {}),
+        // migrating from construct -> Gen2 is not supported
+        enableGen2Migration: false,
```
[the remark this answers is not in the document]
dpilch: 1. This is done purposely to prevent `enableGen2Migration` from appearing in the construct docs. I'll add a comment.
2. This parameter will only be used in a Gen 1 project. This flag primarily does two things. 1. add the table mapping to outputs. 2. set all tables removal policy to retain. These steps are necessary to prep the tables to be imported into to a gen 2 project. When the tables are then imported into gen 2 we no longer need this functionality in the gen 2 project so it will always be set to false.

phani-srikar (dismissed):

packages/amplify-graphql-model-transformer/src/graphql-model-transformer.ts
palpatim: It sounds like a "this shouldn't happen" check to satisfy the compiler then--thanks for the explanation. Printing warnings is probably the wrong way to go here, since we already have so many and we don't actually expect this to occur anyway, so let's keep it as an error.

packages/amplify-graphql-api-construct/src/amplify-graphql-api.ts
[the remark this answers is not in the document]
palpatim: Thanks for clarifying. I mis-thought about this as "enable Gen2 _behavior_" rather than "enable Gen2 _migration_". The purpose and use of the flag makes sense to me now.

packages/amplify-graphql-model-transformer/src/graphql-model-transformer.ts
```diff
@@ -324,25 +324,25 @@ export class ModelTransformer extends TransformerModelBase implements Transforme
...
+        }
+        // TODO: probably need a link to docs for this
+        console.warn(
+          `Could not resolve table name for ${modelName}. DataSourceMappingOutput is incomplete. Please manually add ${modelName} to the mapping for your migration.`,
+        );
+      });
     });
     if (context.transformParameters.enableGen2Migration && context.transformParameters.enableTransformerCfnOutputs) {
       const { scope } = context.stackManager;
+      // TODO: decide final naming before merge to main
```
palpatim, line 345: Note: it's actually the `exportName` that we care about, since that is how customers grab it in their **backend.ts** files.

palpatim (dismissed):

palpatim (approved):

phani-srikar (approved):

PR #2670, opened 2024-06-28
fix: add translation behavior to disable gen 1 patterns

dpilch:
#### Description of changes

Add translation behavior to disable gen 1 patterns:
* Use of `@manyToMany`
* Use of `@searchable`
* Use of `@predictions`
* Use of `fields` argument on `@hasOne`, `@hasMany`, and `@belongsTo`.
* Use of `@hasOne`, `@hasMany`, and `@belongsTo` on required fields.

* Add `allowGen1Patterns` which defaults to `true`.

#### Issue #, if available

N/A

* Unit testing

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Relevant documentation is changed or added (and PR referenced) https://github.com/aws-amplify/docs/pull/7793

### DIFF
--- a/packages/amplify-graphql-api-construct/API.md
+++ b/packages/amplify-graphql-api-construct/API.md
@@ -314,6 +314,7 @@ export interface OptimisticConflictResolutionStrategy extends ConflictResolution
 // @public
 export interface PartialTranslationBehavior {
     readonly allowDestructiveGraphqlSchemaUpdates?: boolean;
+    // @internal
     readonly allowGen1Patterns?: boolean;
     readonly disableResolverDeduping?: boolean;
     readonly enableAutoIndexQueryNames?: boolean;
@@ -439,6 +440,7 @@ export interface TimeToLiveSpecification {
 // @public
 export interface TranslationBehavior {
     readonly allowDestructiveGraphqlSchemaUpdates: boolean;
+    // @internal
     readonly allowGen1Patterns: boolean;
     readonly disableResolverDeduping: boolean;
     readonly enableAutoIndexQueryNames: boolean;
--- a/packages/amplify-graphql-api-construct/API.md
+++ b/packages/amplify-graphql-api-construct/API.md
@@ -314,6 +314,7 @@ export interface OptimisticConflictResolutionStrategy extends ConflictResolution
 // @public
 export interface PartialTranslationBehavior {
     readonly allowDestructiveGraphqlSchemaUpdates?: boolean;
+    readonly allowGen1Patterns?: boolean;
     readonly disableResolverDeduping?: boolean;
     readonly enableAutoIndexQueryNames?: boolean;
     readonly enableSearchNodeToNodeEncryption?: boolean;
--- a/packages/amplify-graphql-api-construct/API.md
+++ b/packages/amplify-graphql-api-construct/API.md
@@ -315,7 +315,7 @@ export interface OptimisticConflictResolutionStrategy extends ConflictResolution
 export interface PartialTranslationBehavior {
     readonly allowDestructiveGraphqlSchemaUpdates?: boolean;
     // @internal
-    readonly allowGen1Patterns?: boolean;
+    readonly _allowGen1Patterns?: boolean;
     readonly disableResolverDeduping?: boolean;
     readonly enableAutoIndexQueryNames?: boolean;
     readonly enableSearchNodeToNodeEncryption?: boolean;
@@ -441,7 +441,7 @@ export interface TimeToLiveSpecification {
 export interface TranslationBehavior {
     readonly allowDestructiveGraphqlSchemaUpdates: boolean;
     // @internal
-    readonly allowGen1Patterns: boolean;
+    readonly _allowGen1Patterns: boolean;
     readonly disableResolverDeduping: boolean;
     readonly enableAutoIndexQueryNames: boolean;
     // (undocumented)
--- a/packages/amplify-graphql-api-construct/API.md
+++ b/packages/amplify-graphql-api-construct/API.md
@@ -438,6 +438,7 @@ export interface TimeToLiveSpecification {
 // @public
 export interface TranslationBehavior {
     readonly allowDestructiveGraphqlSchemaUpdates: boolean;
+    readonly allowGen1Patterns: boolean;
     readonly disableResolverDeduping: boolean;
     readonly enableAutoIndexQueryNames: boolean;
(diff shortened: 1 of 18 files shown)


packages/amplify-graphql-relational-transformer/src/graphql-belongs-to-transformer.ts
```diff
@@ -158,6 +159,14 @@ export class BelongsToTransformer extends TransformerPluginBase {
 
 const validate = (config: BelongsToDirectiveConfiguration, ctx: TransformerContextProvider): void => {
   const { field, object } = config;
+  if (!ctx.transformParameters.allowGen1Patterns) {
+    if (field.type.kind === Kind.NON_NULL_TYPE) {
+      throw new InvalidDirectiveError(`@${BelongsToDirective.name} cannot be used on required fields.`);
```
phani-srikar, line 164: nit: Pointing customer to the faulty field by including its name along with parent type name can help them debug faster.

packages/amplify-graphql-relational-transformer/src/graphql-belongs-to-transformer.ts
```diff
@@ -158,6 +159,14 @@ export class BelongsToTransformer extends TransformerPluginBase {
...
+    }
+    if (config.fields) {
+      throw new InvalidDirectiveError(`fields argument on @${BelongsToDirective.name} is deprecated. Use references instead.`);
```
phani-srikar, line 167: nit: same suggestion as above about more informative error message. Please apply the same for other Errors added in this PR.

packages/amplify-graphql-api-construct/src/types.ts
```diff
@@ -496,6 +496,13 @@ export interface TranslationBehavior {
    * @experimental
    */
   readonly replaceTableUponGsiUpdate: boolean;
+
+  /**
+   * When disabled usage of Gen 1 patterns will result in an error thrown.
+   *
```
phani-srikar, line 502: Are we planning to describe what these Gen-1 patterns are in a documentation update? If not, I would just call them out here. If a complete list is more verbose, we can include couple of examples to give a better idea in terms of where it applies.

phani-srikar (changes requested):

packages/amplify-graphql-api-construct/src/types.ts
dpilch: This isn't fully intended to be used by customers. This will be set to false in `defineData` and won't be configurable by customers. Customers could set this when using the construct directly, but there isn't really a use case for this because it will only restrict the features available in the construct.

packages/amplify-graphql-api-construct/src/__tests__/__functional__/disable-gen1-patterns.test.ts
```diff
@@ -0,0 +1,353 @@
...
+        new AmplifyGraphqlApi(stack, 'TestApi', {
+          definition: AmplifyGraphqlDefinition.fromString(schema),
+          authorizationModes: {
+            apiKeyConfig: { expires: cdk.Duration.days(7) },
+          },
+        }),
+    ).not.toThrow();
+  });
+
+  test('does not allow @manyToMany', () => {
```
palpatim, line 49: nit here & throughout: test titles should reflect the conditions better, as in
```suggestion
  test('does not allow @manyToMany when allowGen1Patterns is false', () => {
```

You could accomplish the same thing by organizing nested describe blocks:
```typescript
describe('allowGen1Patterns', () => {
  test('defaults to allow', () => {...});
  describe('flag is true', () => {/* all the cases */});
  describe('flag is false', () => {/* all the cases */});
});

File: packages/amplify-graphql-api-construct/src/types.ts (hunk 3 above)
Comment (palpatim) [reply packages/amplify-graphql-api-construct/src/types.ts:502]: 1. Agree with the intent, but I notice that we don't flag this as an internal API. Please add a doc comment noting it so customers understand that we reserve the right to change this API without a breaking change
2. For our own future selves, can you add one or two examples, like "e.g., @manyToMany, fields-style references, @predictions, etc", so we know what "Gen 1 patterns" means in 6 months? :)

File: packages/amplify-graphql-relational-transformer/src/graphql-has-many-transformer.ts [hunk 5]
```diff
@@ -158,6 +159,14 @@ export class HasManyTransformer extends TransformerPluginBase {
 
 const validate = (config: HasManyDirectiveConfiguration, ctx: TransformerContextProvider): void => {
   const { field } = config;
+  if (!ctx.transformParameters.allowGen1Patterns) {
+    if (field.type.kind === Kind.NON_NULL_TYPE) {
+      throw new InvalidDirectiveError(`@${HasManyDirective.name} cannot be used on required fields.`);
+    }
+    if (config.fields) {
+      throw new InvalidDirectiveError(`fields argument on @${HasManyDirective.name} is deprecated. Use references instead.`);
```
palpatim: Here and throughout all error messages: It's not deprecated, it's outright forbidden in Gen2 projects.

packages/amplify-graphql-transformer/API.md
```diff
@@ -25,7 +25,7 @@ import { UserDefinedSlot } from '@aws-amplify/graphql-transformer-core';
 export const constructTransform: (config: TransformConfig) => GraphQLTransform;
 
 // @public (undocumented)
-export const constructTransformerChain: (options?: TransformerFactoryArgs) => TransformerPluginProvider[];
+export const constructTransformerChain: (options?: TransformerFactoryArgs, allowGen1Patterns?: boolean) => TransformerPluginProvider[];
```
palpatim, line 28: Can we put `allowGen1Patterns` into `TransformerFactoryArgs` rather than tacking on a dangling argument?

packages/amplify-graphql-api-construct/API.md
```diff
@@ -314,6 +314,8 @@ export interface OptimisticConflictResolutionStrategy extends ConflictResolution
 // @public
 export interface PartialTranslationBehavior {
     readonly allowDestructiveGraphqlSchemaUpdates?: boolean;
+    // @internal
+    readonly _allowGen1Patterns?: boolean;
```
dpilch, line 318: JSII requires params marked with `@internal` to be prefixed with an underscore.

phani-srikar (approved):

palpatim (approved):

PR #2694, opened 2024-07-08
fix: add tags to dynamodb tables

sundersc:
#### Description of changes

Amplify Gen 2 applies some default tags to make it easy for customers to use resource explorer or "MyApplications" to browse their app's resources. The tags are applied on the CFN level and typically (automatically) applied onto the underlying AWS resources by CFN. Because we use a custom resource for DynamoDB tables, these tags aren't automatically applied.

This PR fixes that issue by copying the tags from the resource provision lambda and assign it to the tables.

No.

#### Issue #, if available
NA.

- New E2E tests
- Manual test

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-graphql-api-construct-tests/src/__tests__/backends/amplify-table/simple-todo/updateIndex/app.ts
+++ b/packages/amplify-graphql-api-construct-tests/src/__tests__/backends/amplify-table/simple-todo/updateIndex/app.ts
@@ -1,6 +1,6 @@
 #!/usr/bin/env node
 import 'source-map-support/register';
-import { App, Stack, Duration } from 'aws-cdk-lib';
+import { App, Stack, Duration, Tags } from 'aws-cdk-lib';
 // @ts-ignore
 import { AmplifyGraphqlApi, AmplifyGraphqlDefinition } from '@aws-amplify/graphql-api-construct';
 import { BillingMode, StreamViewType } from 'aws-cdk-lib/aws-dynamodb';
@@ -51,3 +51,9 @@ todoTable.setGlobalSecondaryIndexProvisionedThroughput('byName2', {
 todoTable.pointInTimeRecoveryEnabled = true;
 todoTable.sseSpecification = { sseEnabled: false };
 todoTable.streamSpecification = { streamViewType: StreamViewType.KEYS_ONLY };
+
+Tags.of(stack).add('created-by', 'amplify');
+Tags.of(stack).add('amplify:deployment-type', 'pipeline');
+Tags.of(stack).add('amplify:deployment-branch', 'main');
+Tags.of(stack).add('amplify:appId', '123456');
+Tags.of(stack).add('amplify:friendly-name', 'amplifyData');
--- a/packages/amplify-graphql-api-construct-tests/src/__tests__/backends/amplify-table/simple-todo/updateIndex/app.ts
+++ b/packages/amplify-graphql-api-construct-tests/src/__tests__/backends/amplify-table/simple-todo/updateIndex/app.ts
@@ -52,8 +52,8 @@ todoTable.pointInTimeRecoveryEnabled = true;
 todoTable.sseSpecification = { sseEnabled: false };
 todoTable.streamSpecification = { streamViewType: StreamViewType.KEYS_ONLY };
 
-Tags.of(stack).add('created-by', 'amplify');
-Tags.of(stack).add('amplify:deployment-type', 'pipeline');
-Tags.of(stack).add('amplify:deployment-branch', 'main');
-Tags.of(stack).add('amplify:appId', '123456');
-Tags.of(stack).add('amplify:friendly-name', 'amplifyData');
+Tags.of(stack).add('created-by', 'amplify-updated');
+Tags.of(stack).add('amplify:deployment-type', 'pipeline-updated');
+Tags.of(stack).add('amplify:deployment-branch', 'main-updated');
+Tags.of(stack).add('amplify:appId', '123456-updated');
+Tags.of(stack).add('amplify:friendly-name', 'amplifyData-updated');
--- a/packages/amplify-graphql-model-transformer/src/resources/amplify-dynamodb-table/amplify-table-manager-lambda/amplify-table-manager-handler.ts
+++ b/packages/amplify-graphql-model-transformer/src/resources/amplify-dynamodb-table/amplify-table-manager-lambda/amplify-table-manager-handler.ts
@@ -16,6 +16,10 @@ import {
   UpdateTableCommandInput,
   UpdateTimeToLiveCommandInput,
 } from '@aws-sdk/client-dynamodb';
+import {
+  Lambda,
+  ListTagsCommand,
+} from '@aws-sdk/client-lambda';
 import { OnEventResponse } from '../amplify-table-manager-lambda-types';
 import * as cfnResponse from './cfn-response';
 import { startExecution } from './outbound';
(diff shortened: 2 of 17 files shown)


packages/amplify-graphql-api-construct-tests/src/__tests__/backends/amplify-table/simple-todo/updateIndex/app.ts
```diff
@@ -51,3 +51,9 @@ todoTable.setGlobalSecondaryIndexProvisionedThroughput('byName2', {
 todoTable.pointInTimeRecoveryEnabled = true;
 todoTable.sseSpecification = { sseEnabled: false };
 todoTable.streamSpecification = { streamViewType: StreamViewType.KEYS_ONLY };
+
+Tags.of(stack).add('created-by', 'amplify');
+Tags.of(stack).add('amplify:deployment-type', 'pipeline');
+Tags.of(stack).add('amplify:deployment-branch', 'main');
+Tags.of(stack).add('amplify:appId', '123456');
+Tags.of(stack).add('amplify:friendly-name', 'amplifyData');
```
palpatim, line 59: I believe you're mimicking tags that would be added by the backend? That's reasonable, but it does lead to some confusion in the test body itself, where it looks like deployments change from "sandbox" to "pipeline" for some reason. Instead, can we make the tag values clearly artificial (e.g., "original" and "modified") to make it clear you're testing that tags update as expected?

packages/amplify-graphql-model-transformer/src/resources/amplify-dynamodb-table/amplify-table-manager-lambda/amplify-table-manager-handler.ts
```diff
@@ -739,6 +799,32 @@ export const getSseUpdate = (currentState: TableDescription, endState: CustomDDB
...
 };
 
+/**
+ * Compares the currentState with the tags on the resource provider lambda to determine if the tags are updated
+ * @param currentTags current tags on the table
+ * @param newTags new tags on the lambda
+ * @returns Boolean indicating if the tags are updated
+ */
+export const requiresTagsUpdate = (currentTags: DynamoDBTag[], newTags?: DynamoDBTag[]): boolean => {
+  if (!newTags || newTags.length === 0) {
```
palpatim, line 809: This means we won't support deleting all tags. Is that expected?

packages/amplify-graphql-model-transformer/src/resources/amplify-dynamodb-table/amplify-table-manager-lambda/amplify-table-manager-handler.ts
```diff
@@ -178,6 +215,28 @@ const processOnEvent = async (event: AWSCDKAsyncCustomResource.OnEventRequest):
         }
       }
 
+      // Determine if table needs tags update.
+      // For Gen2 deployments, the tags do not change between deployments. This check is to ensure that
+      // the tags are applied to tables created earlier (before adding tagging support).
+      // TODO: Handle tags update for CDK construct deployments
```
palpatim, line 221: Can't comment directly on the relevant code above, but we should update our GSI tests to ensure that tags propagate correctly during GSI updates that entirely replace the table.

packages/amplify-graphql-model-transformer/src/resources/amplify-dynamodb-table/amplify-table-manager-lambda/amplify-table-manager-handler.ts
```diff
@@ -865,11 +951,16 @@ type CreateTableResponse = {
...
  */
-export const extractTableInputFromEvent = (
+export const extractTableInputFromEvent = async (
   event: AWSCDKAsyncCustomResource.OnEventRequest | AWSCDKAsyncCustomResource.IsCompleteRequest,
-): CustomDDB.Input => {
+  context: TableManagerContext,
+): Promise<CustomDDB.Input> => {
   // isolate the resource properties from the event and remove the service token
-  const resourceProperties = { ...event.ResourceProperties } as Record<string, any> & { ServiceToken?: string };
+  const tags = await getLambdaTags(context.invokedFunctionArn);
```
palpatim, line 959: Can you confirm that the lifecycle of the CDK deployment is such that the Lambda tags are fully deployed by the time this handler is invoked? In other words, we're not going to hit some kind of weird race condition in future where we see tags have not yet been propagated to the function?

I'm assuming this is safe, since Tags are applied during synth, and I would expect them to be applied to the Lambda resource just like other attributes, but since Tags are standalone, we want to make sure we don't hit either lifecycle ordering or eventual consistency problems down the road.

dpilch: I wasn't able to find any documentation on this. I think it would come down to how the custom resource provider is implemented by cloud formation. I will check with someone on cloud formation to see if they know.

dpilch: CFN states the tags will be set before the event on the lambda is processed.

packages/amplify-graphql-model-transformer/src/resources/amplify-dynamodb-table/amplify-table-manager-lambda/amplify-table-manager-handler.ts
dpilch: I don't see any reason to disallow this case. I'll update this.

palpatim (dismissed):

packages/amplify-graphql-model-transformer/src/resources/amplify-dynamodb-table/amplify-table-manager-lambda/amplify-table-manager-handler.ts
```diff
@@ -806,7 +806,7 @@ export const getSseUpdate = (currentState: TableDescription, endState: CustomDDB
  * @returns Boolean indicating if the tags are updated
  */
 export const requiresTagsUpdate = (currentTags: DynamoDBTag[], newTags?: DynamoDBTag[]): boolean => {
-  if (!newTags || newTags.length === 0) {
```
sundersc: Calling the UpdateTags with an empty array throws an runtime error. The PR is intended to fix Gen2 workflow which always contains at least one tag. If we need to handle CDK deployments as well, we need to handle it separately, this logic won't work.

sundersc: I have added a logic to remove the existing tags. It required additional IAM permission `UntagResource` and new API call.

packages/amplify-graphql-api-construct-tests/src/__tests__/backends/amplify-table/simple-todo/updateIndex/app.ts
sundersc: Updated.

packages/amplify-graphql-model-transformer/src/resources/amplify-dynamodb-table/amplify-table-manager-lambda/amplify-table-manager-handler.ts
sundersc: Removing tags requires a different API call. I have updated the logic to remove tags.

palpatim (approved):

phani-srikar (approved):

PR #2703, opened 2024-07-12
test: migration e2es

dpilch:
#### Description of changes

Add migration e2es for three schemas (basic, references, many-to-many). The tests all follow the same pattern.

1. Create a gen 1 project.
2. Get the table mapping from gen 1 project
3. Create a gen 2 project with the table mapping.
4. Make queries/mutations on both gen 1 and gen 2.
5. Delete gen 1 project.
6. Query gen 2 project to ensure imported table still exists.

In a following PR these tests will be augmented to include modifying the table from the gen 2 project after the gen 1 project is deleted.

Additional E2E tests for import validation will come in a later PR.

N/A

#### Issue #, if available

N/A

N/A

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-e2e-tests/schemas/many-to-many.graphql
```diff
@@ -0,0 +1,12 @@
+type Post @model @auth(rules: [{ allow: public }]) {
```
palpatim, line 1: Can we disambiguate the file names between `many-to-many` and `many_to_many`? Preference would be to just reuse them, but it looks like you've got auth rules in one and not the other, so at least a clarification on what the schemas are for would be helpful.

packages/amplify-e2e-tests/schemas/simple_model_public_auth.graphql
```diff
@@ -0,0 +1,4 @@
+type Todo @model @auth(rules: [{ allow: public }]) {
+  id: ID!
+  content: String
+}
```
palpatim, line 4: is this going to fail linting/prettier check b/c of the missing line at EOF?

packages/amplify-graphql-api-construct-tests/src/__tests__/backends/migration/base/app.ts
```diff
@@ -0,0 +1,35 @@
...
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const packageJson = require('../package.json');
+
+const app = new App();
+const stack = new Stack(app, packageJson.name.replace(/_/g, '-'), {
+  env: { region: process.env.CLI_REGION || 'us-west-2' },
+});
+
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const importedAmplifyDynamoDBTableMap: Record<string, string> = require('../table-map.json');
```
palpatim, line 15: Could we extend the `StackConfig` structure used by `packages/amplify-graphql-api-construct-tests/src/__tests__/backends/configurable-stack/app.ts` with this map instead of continuing to propagate new backend configs? We shouldn't make heroic efforts to do so, but it'd be nice to reduce the number of custom-built stacks we've got floating around our E2E tests

packages/amplify-graphql-api-construct-tests/src/__tests__/backends/migration/many-to-many/app.ts
```diff
@@ -0,0 +1,65 @@
+#!/usr/bin/env node
```
palpatim, line 1: as above, is there an opportunity to use `packages/amplify-graphql-api-construct-tests/src/__tests__/backends/configurable-stack/app.ts` instead of a custom stack?

packages/amplify-graphql-api-construct-tests/src/__tests__/migration/base-migration.test.ts
```diff
@@ -0,0 +1,157 @@
...
+    }
+    try {
+      await cdkDestroy(gen2ProjRoot, '--all');
+    } catch (_) {
+      /* No-op */
+    }
+
+    try {
+      // Tables are set to retain when migrating from gen 1 to gen 2
+      await deleteDDBTables(Object.values(dataSourceMapping));
```
palpatim, line 41: If they're set to retain, why are we attempting to delete them? Should we be asserting that the delete fails? In fact, should we have an E2E that asserts that deletion protection is enabled for a migrating table?

packages/amplify-graphql-api-construct-tests/src/__tests__/migration/base-migration.test.ts
```diff
@@ -0,0 +1,157 @@
...
+    } catch (_) {
+      /* No-op */
+    }
+
+    deleteProjectDir(gen1ProjRoot);
+    deleteProjectDir(gen2ProjRoot);
+  });
+
+  test('base migration', async () => {
```
palpatim, line 50: nit: can you clarify in the test title what a "base migration" means?

packages/amplify-graphql-api-construct-tests/src/__tests__/migration/base-migration.test.ts
```diff
@@ -0,0 +1,157 @@
...
+      /* GraphQL */ `
+        mutation CREATE_TODO {
+          createTodo(input: { content: "todo desc" }) {
+            id
+            content
+          }
+        }
+      `,
+    );
+    expect(gen1Result.statusCode).toEqual(200);
```
palpatim, line 75: Since AppSync returns a 200 even if the result has GraphQL errors, let's add a comment that we'll assert the actual results of the create by inspecting the `list` results below. Presumably this assertion is just to make sure we don't get a 400 b/c of a malformed request?

packages/amplify-graphql-api-construct-tests/src/__tests__/migration/migration-validation.test.ts
palpatim, line 1: Is this supposed to be removed?

packages/amplify-graphql-api-construct-tests/src/commands.ts
```diff
@@ -1,7 +1,20 @@
 import * as path from 'path';
 import * as fs from 'fs';
 import { copySync, moveSync, readFileSync, writeFileSync } from 'fs-extra';
-import { getScriptRunnerPath, sleep, nspawn as spawn } from 'amplify-category-api-e2e-core';
+import {
```
palpatim, line 4: very nit: sort imports

packages/amplify-graphql-api-construct-tests/src/commands.ts
```diff
@@ -160,3 +173,60 @@ export const updateCDKAppWithTemplate = (cwd: string, templatePath: string): voi
...
+  DataSourceMappingOutput: string;
+}> => {
+  await initJSProjectWithProfile(projRoot, { name });
+  await addApiWithoutSchema(projRoot, { transformerVersion: 2 });
+  await updateApiSchema(projRoot, name, schema);
+  await amplifyPush(projRoot);
+
+  // TODO: can't use feature flag until released in CLI
+  // addFeatureFlag(projRoot, 'graphqltransformer', 'enableGen2Migration', true);
+  // await amplifyPush(projRoot);
```
palpatim, line 200: I assume that adding the feature flag will require a second `amplify push`, since the flow we're testing is migration of an existing Gen1 project. Let's add a comment to make sure we remember that when the time comes to uncomment this block.

packages/amplify-graphql-api-construct-tests/src/__tests__/backends/migration/base/app.ts
dpilch: I didn't realize that we had this. I'll update to use that.

packages/amplify-graphql-api-construct-tests/src/__tests__/migration/base-migration.test.ts
dpilch: This is CFN deletion policy (not DDB deletion protection). The CFN deletion policy is set to retain on the Gen 1 and Gen 2 project.
1. The tables will still exist after the customer removes their gen 1 project.
2. The tables will still exist if the customer stops migrating and stays with gen 1.

The deletion is necessary to prevent resource leaks because the tables will not be removed after the `cdk destroy`/`amplify delete`.

The CFN deletion policy assertion is covered by the final list result result. This ensures that after the gen 1 project is deleted that the gen 2 project still has access to the DDB table. https://github.com/aws-amplify/amplify-category-api/pull/2703/files#diff-5085bba2562ae965a2878aeeae16f2cec073ff1564d2d4e459559c600fd42fd0R135-R137

I'll add some more comments so this is more clear.

packages/amplify-e2e-tests/schemas/many-to-many.graphql
dpilch: Accidental addition. Only `many-to-many.graphql` is needed.

palpatim (approved):

palpatim (approved):

sundersc (approved):

PR #2728, opened 2024-07-29
feat: allow SQL Lambda to be hotswapped in sandbox

bobbyu99:
#### Description of changes
Added a deployment identifier to differentiate between a sandbox and branch deployment. If it is a sandbox deployment, `SQLLayerVersionCustomResource` and `SQLSNSTopicARNCustomResource` will use a physical ID in the format of `resourceName-YYYY-MM-DD` to avoid multiple deployments in the same day. 

Unit test added.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-graphql-model-transformer/src/resolvers/rds/resolver.ts
```diff
@@ -196,14 +196,28 @@ export const createRdsLambda = (
...
   const key = Fn.join('', [SQLLayerVersionManifestKeyPrefix, Fn.ref('AWS::Region')]);
 
   const manifestArn = `arn:aws:s3:::${SQLLayerManifestBucket}/${key}`;
 
   const resourceName = resourceNames.sqlLayerVersionResolverCustomResource;
+
+  // if deployment type is sandbox, use id in the format: resourceName-YYYY-MM-DD to avoid multiple deployments in the same day
+  // if deployment type is branch, use id in the original format
+  let physicalIdValue: string;
+  if (context?.deploymentIdentifier?.deploymentType === 'sandbox') {
```
palpatim, line 215: nit: consider a [TS type predicate](https://www.typescriptlang.org/docs/handbook/advanced-types.html#using-type-predicates) to determine whether you have a sandbox type or not. That has the advantage of centralizing type introspection into a single function rather than scattering it throughout the project. It also allows TypeScript to infer the structure of the rest of the object even if the type structure gets hairy. See examples in `packages/amplify-graphql-transformer-core/src/utils/model-datasource-strategy-utils.ts` ([example](https://github.com/aws-amplify/amplify-category-api/blob/main/packages/amplify-graphql-transformer-core/src/utils/model-datasource-strategy-utils.ts#L76-L87))

packages/amplify-graphql-model-transformer/src/resolvers/rds/resolver.ts
```diff
@@ -196,14 +196,28 @@ export const createRdsLambda = (
...
+    physicalIdValue = `${resourceName}-${new Date().toISOString().substring(0, 10)}`;
```
palpatim, line 216: Why is "per day" the right granularity for controlling deployments in a sandbox? Why would we try to compromise on this at all? By introducing an arbitrary cutoff, we're going to still require deployments for some customers whose sandbox deployments happen to span a day boundary, while missing layer version updates for customers who do multiple deployments that don't happen to span a day boundary. Since sandboxes aren't intended for production, why can't we just use a strictly deterministic resource name, based e.g. on the sandbox name?

packages/amplify-graphql-model-transformer/src/resolvers/rds/resolver.ts
```diff
@@ -231,14 +245,28 @@ export const createLayerVersionCustomResource = (scope: Construct, resourceNames
...
+): AwsCustomResource => {
   const { SQLLayerManifestBucket, SQLLayerManifestBucketRegion, SQLSNSTopicARNManifestKeyPrefix } = ResourceConstants.RESOURCES;
 
   const key = Fn.join('', [SQLSNSTopicARNManifestKeyPrefix, Fn.ref('AWS::Region')]);
 
   const manifestArn = `arn:aws:s3:::${SQLLayerManifestBucket}/${key}`;
 
   const resourceName = resourceNames.sqlSNSTopicARNResolverCustomResource;
+
+  // if deployment type is sandbox, use id in the format: resourceName-YYYY-MM-DD to avoid multiple deployments in the same day
```
palpatim, line 261: Same point here as above: why partition by date?

packages/amplify-graphql-transformer-core/API.md
```diff
@@ -309,7 +310,7 @@ export class GraphQLTransform {
     // Warning: (ae-forgotten-export) The symbol "TransformOption" needs to be exported by the entry point index.d.ts
     //
     // (undocumented)
-    transform({ assetProvider, dataSourceStrategies, nestedStackProvider, parameterProvider, rdsLayerMapping, rdsSnsTopicMapping, schema, scope, sqlDirectiveDataSourceStrategies, synthParameters, }: TransformOption): void;
+    transform({ assetProvider, dataSourceStrategies, nestedStackProvider, parameterProvider, rdsLayerMapping, rdsSnsTopicMapping, schema, scope, sqlDirectiveDataSourceStrategies, synthParameters, deploymentIdentifier, }: TransformOption): void;
```
palpatim, line 313: Should we make `deploymentIdentifier` an attribute of `synthParameters` rather than dangling it off the end of the transform input?

packages/amplify-graphql-transformer-core/src/transformation/transform.ts
```diff
@@ -195,6 +197,7 @@ export class GraphQLTransform {
     scope,
     sqlDirectiveDataSourceStrategies,
     synthParameters,
+    deploymentIdentifier,
```
palpatim, line 200: as above, should `deploymentIdentifier` be part of `synthParameters` rather than a top-level attribute of `TransformOption`?

packages/amplify-graphql-transformer-interfaces/API.md
```diff
@@ -146,6 +146,16 @@ export interface AssetProvider {
     provide: (scope: Construct, name: string, props: AssetProps) => S3Asset;
 }
 
+// @public (undocumented)
+export interface BranchIdentifier {
```
palpatim, line 150: See note above about naming. While the internal interfaces don't strictly have to correspond to the external interfaces, in cases where the serve the same purpose with the same data, it reduces our confusion to keep the names the same.

packages/amplify-graphql-transformer/src/graphql-transformer.ts
```diff
@@ -118,6 +119,7 @@ export type ExecuteTransformConfig = TransformConfig &
     parameterProvider?: TransformParameterProvider;
     assetProvider: AssetProvider;
     synthParameters: SynthParameters;
+    deploymentIdentifier?: DeploymentIdentifier;
```
palpatim, line 122: This file only applies to Gen 1 transformer right? Is `deploymentIdentifier` required for Gen 1? (Or I might be misremembering how this file is used!)

packages/amplify-graphql-model-transformer/src/resolvers/index.ts
```diff
@@ -1,3 +1,4 @@
 export * from './generators';
 export * from './common';
+export * from './rds';
```
palpatim, line 3: Why are we now exporting this? (I'm assuming it's to make new logic in `resolver.ts` easier to access somewhere?

packages/amplify-graphql-model-transformer/src/resolvers/rds/resolver.ts
bobbyu99: I was discussing with Christ, Phani, Dane, and Josef, the major concern we have is that, if we use a fixed name like the resourceName, it will not pick up the weekly release/update that might contain security patches. Between weekly and daily, we decided to go with daily. 

I'm still trying to understand the nuances here. Assuming we have a layer version update at noon, and a user deploys once in the morning, and once in the afternoon. 
1) For the current approach using Date.now(), this user will have the update
2) For "per day" approach, this user will not have the update. Unless they deploy again the next day
3) For fixed name approach with only `${resourceName}`, they will never get this update no matter how many times they deploy (i'm not sure if i get it right here)

packages/amplify-graphql-model-transformer/src/resolvers/index.ts
bobbyu99: I exported `createLayerVersionCustomResource`, `createSNSTopicARNCustomResource` for the unit test in the same package:
`import { createLayerVersionCustomResource, createSNSTopicARNCustomResource } from '@aws-amplify/graphql-model-transformer';
`
I guess define a relative path instead of exporting is a better choice here?

palpatim: No need to change, as long as we don't need to export it to the public interface. I was just wondering why the new export :)

packages/amplify-graphql-api-construct/src/types.ts
```diff
@@ -786,8 +786,35 @@ export interface AmplifyGraphqlApiProps {
    * For more information, refer to https://docs.amplify.aws/lib/datastore/getting-started/q/platform/js/
    */
   readonly dataStoreConfiguration?: DataStoreConfiguration;
+
+  /**
+   * The deployment identifier for the current deployment.
+   */
+  readonly deploymentIdentifier?: DeploymentIdentifier;
```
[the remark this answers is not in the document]
bobbyu99: `translationBehavior` only contains boolean props and have 1-on-1 parity with feature flags. I was going with that idea initially, but I feel its not very appropriate. Is there any other place we can put `deploymentIdentifier`?

packages/amplify-graphql-transformer/src/graphql-transformer.ts
bobbyu99: This file is used in both Gen 1 and Gen 2. But `deploymentIdentifier` is not required for Gen 1

packages/amplify-graphql-api-construct/src/types.ts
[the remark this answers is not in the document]
palpatim: `translationBehavior` may not be the right place then. I still object to this being part of the API definition though — it’s not definition, it’s context. That context happens to change the way the resource name is generated, but it’s similar to how CDK uses context to control, e.g., regions, account IDs, or dependent resource IDs.

If we don’t use `translationBehavior`, I’m wondering if we can use a shell environment variable or similar to control the transformation behavior.

If that doesn’t work, then we should create a new optional struct inside the GraphQL API construct to hold these flags — something like `transformationContext` or similar. The use case is that they should only ever be used by the Gen2 CLI. We’d mark the entire struct as experimental, or internal, or something, and put big scary warnings on it that customers shouldn’t rely on the behavior directly.

packages/amplify-graphql-api-construct/src/amplify-graphql-api.ts
```diff
@@ -275,6 +284,27 @@ export class AmplifyGraphqlApi extends Construct {
...
   }
 
+  /**
+   * Gets the deployment information from the context.
+   * @returns the deployment identifier.
+   */
+  private getDeploymentIdentifier(): DeploymentIdentifier {
+    const deploymentType = this.node.tryGetContext(CDKContextKey.DEPLOYMENT_TYPE);
+
+    if (deploymentType === 'sandbox') {
```
palpatim, line 294: Consider abstracting the 'sandbox' magic string into a centralized constant.

packages/amplify-graphql-api-construct/src/amplify-graphql-api.ts
```diff
@@ -275,6 +284,27 @@ export class AmplifyGraphqlApi extends Construct {
...
+      return {
+        deploymentType: deploymentType,
+        namespace: this.node.tryGetContext(CDKContextKey.BACKEND_NAMESPACE),
```
palpatim, line 297: What is the behavior if `BACKEND_NAMESPACE` context key isn't set? Will there be a meaningful error propagated to a customer-facing context?

..
..
..

Trick question :). Even if the error is something like "backendNamespace context key not set", that doesn't tell the customer how to fix it. We should catch the condition ourselves -- if either of the context keys are set, both must be set, and we should craft an appropriate error message to help customers resolve the situation.)

packages/amplify-graphql-api-construct/src/amplify-graphql-api.ts
```diff
@@ -275,6 +284,27 @@ export class AmplifyGraphqlApi extends Construct {
...
+      } as SandboxDeploymentIdentifier;
+    } else {
+      return {
+        deploymentType: deploymentType,
+        namespace: this.node.tryGetContext(CDKContextKey.BACKEND_NAMESPACE),
+        name: this.node.tryGetContext(CDKContextKey.BACKEND_NAME),
```
palpatim, line 303: What if one or both of the context key isn't set? Maybe it's always set for a Gen 2 CLI invocation, but customers can also do a straight-up `cdk synth` or `cdk deploy` without setting context.

packages/amplify-graphql-model-transformer/API.md
```diff
@@ -5,7 +5,10 @@
 ```ts
 
 import { AppSyncDataSourceType } from '@aws-amplify/graphql-transformer-interfaces';
+import { AwsCustomResource } from 'aws-cdk-lib/custom-resources';
```
palpatim, line 8: I'm surprised at the number of new exports from this package -- are these all only to enable unit testing of new helpers? If so, we should adjust our package-level `index.ts` to ensure that we only vend truly public exports from the package. If not, what other packages are using the new exports? (I'm assuming these aren't intended for customer consumption?)

packages/amplify-graphql-model-transformer/src/__tests__/custom-resources.test.ts
```diff
@@ -0,0 +1,74 @@
...
+  const snsTopicARNCustomResourceType = 'Custom::SQLSNSTopicARNCustomResource';
+
+  let app: App;
+  let stack: Stack;
+  beforeEach(() => {
+    app = new App();
+    stack = new Stack(app, 'TestStack');
+  });
+
+  it('should create a layer version custom resource with correct physical ID format for sandbox deployment', () => {
```
palpatim, line 31: Do we have equivalent tests for branch deployments?

packages/amplify-graphql-model-transformer/src/resolvers/rds/resolver.ts
```diff
@@ -216,7 +230,7 @@ export const createLayerVersionCustomResource = (scope: Construct, resourceNames
       },
       // Make the physical ID change each time we do a deployment, so we always check for the latest version. This means we will never have
       // a strictly no-op deployment, but the SQL Lambda configuration won't change unless the actual layer value changes
```
palpatim, line 232: nit: move this comment above where you're creating they physical ID so we're explaining the motivation in context.

packages/amplify-graphql-model-transformer/src/resolvers/rds/resolver.ts
```diff
@@ -251,7 +279,7 @@ export const createSNSTopicARNCustomResource = (scope: Construct, resourceNames:
...
```
palpatim, line 281: As above, move this comment to where you're creating the physical ID

packages/amplify-graphql-api-construct/src/amplify-graphql-api.ts
```diff
@@ -275,6 +294,62 @@ export class AmplifyGraphqlApi extends Construct {
...
+  /**
+   * Gets the deployment information from the context.
+   * @returns the deployment identifier.
+   */
+  private getDeploymentIdentifier(): DeploymentIdentifier {
+    const deploymentType = this.node.tryGetContext(CDKContextKey.DEPLOYMENT_TYPE);
+    const expectedDeploymentTypes = [this.SANDBOX_DEPLOYMENT_TYPE, this.BRANCH_DEPLOYMENT_TYPE];
+
+    // Throw an error if the deployment type is neither sandbox nor branch.
+    if (!expectedDeploymentTypes.includes(deploymentType)) {
```
edwardfoyle, line 306: If the GQL construct is used outside of the context of an Amplify backend (ie as a construct in a native CDK project), then these context props will not be present

packages/amplify-graphql-transformer-interfaces/src/transformer-context/synth-parameters.ts
```diff
@@ -7,4 +7,18 @@ export type SynthParameters = {
   identityPoolId?: string;
   adminRoles?: string[];
   enableIamAccess?: boolean;
+  deploymentIdentifier?: DeploymentIdentifier;
 };
+
+export interface SandboxDeploymentIdentifier {
```
edwardfoyle, line 13: branch and sandbox deployments both have a name and namespace:
https://github.com/aws-amplify/amplify-backend/blob/39175772301cd9db58fd07ff6ed3e2c0e20fb876/packages/plugin-types/API.md?plain=1#L64-L74

In the case of sandbox, the name is the value of `package.json#name` and the namespace is either the local username or the value of `--identifier` supplied to the CLI

packages/amplify-graphql-api-construct/src/amplify-graphql-api.ts
bobbyu99: To clarify, you are referring to the class variables `this.SANDBOX_DEPLOYMENT_TYPE, this.BRANCH_DEPLOYMENT_TYPE` right?

packages/amplify-graphql-api-construct/src/amplify-graphql-api.ts
```diff
@@ -275,6 +294,62 @@ export class AmplifyGraphqlApi extends Construct {
...
+        `${CDKContextKey.DEPLOYMENT_TYPE} must be one of (${expectedDeploymentTypes.join(', ')}). ` +
+        'Please set the deployment type by running:\n' +
+        'cdk --context amplify-backend-type=sandbox\n' +
+        'or\n' +
+        'cdk --context amplify-backend-type=branch'
+      );
+    }
+
+    if (deploymentType === this.SANDBOX_DEPLOYMENT_TYPE) {
+      const namespace = this.node.tryGetContext(CDKContextKey.BACKEND_NAMESPACE);
```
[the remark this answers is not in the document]
bobbyu99: Sorry im not quite getting the question here. I'm getting the deployment type, the name and namespace to construct a DeploymentIdentifier to be used later. 

But as you mention in the other comment above that sandbox also has both name and namespace, i dont need a `if ... else ...`. Regardless of sandbox or branch deployments, i gotta make sure that both name and namespace are present.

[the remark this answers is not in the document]
edwardfoyle: It doesn't look like you need to pass in a whole DeploymentIdentifier though. The only place it's being read (I think) is in `isSandboxDeployment` which only looks at the deployment type, not the other fields. So maybe you could just pass in `deploymentType` and not care about reading any of the other values

packages/amplify-graphql-api-construct/src/amplify-graphql-api.ts
edwardfoyle: Correct, none of these context values would be present in a native CDK project. The construct should probably treat absence of these values the same as a "branch" deployment

packages/amplify-graphql-api-construct/src/amplify-graphql-api.ts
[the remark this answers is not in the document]
bobbyu99: I see what you mean. Yeah `isSandboxDeployment` only needs `deploymentType`. As I was discussing with my team earlier, we decided to include name and namespace as we think they could be useful information in the future.

packages/amplify-graphql-api-construct/src/amplify-graphql-api.ts
bobbyu99: Let me rephrase what you are suggesting here. Are you saying that in the case of a native CDK project, where `this.SANDBOX_DEPLOYMENT_TYPE, this.BRANCH_DEPLOYMENT_TYPE` are absent, `getDeploymentIdentifier()` should treat this case as a branch deployment and return a `BranchDeploymentIdentifier`?

I see another workaround for the absence of `this.SANDBOX_DEPLOYMENT_TYPE, this.BRANCH_DEPLOYMENT_TYPE` is to make them local inside `getDeploymentIdentifier()`. Will this solve the problem?

palpatim: You should only emit the warning if a deployment type is actually present:

```
if (deploymentType && !expectedDeploymentTypes.includes(deploymentType)) {
```

packages/amplify-graphql-api-construct/src/amplify-graphql-api.ts
```diff
@@ -302,52 +297,35 @@ export class AmplifyGraphqlApi extends Construct {
...
-        'or\n' +
-        'cdk --context amplify-backend-type=branch'
+      console.warn(
+        `${CDKContextKey.DEPLOYMENT_TYPE} is recommended to be one of (${expectedDeploymentTypes.join(', ')}).\n` +
+          'Consider setting the deployment type by running:\n' +
+          'cdk --context amplify-backend-type=sandbox\n' +
+          'or\n' +
+          'cdk --context amplify-backend-type=branch\n',
       );
     }
```
palpatim, line 309: You should go ahead and return if we encounter this condition, since we can't safely proceed with an unknown deployment type.

packages/amplify-graphql-api-construct/src/amplify-graphql-api.ts
```diff
@@ -275,6 +289,45 @@ export class AmplifyGraphqlApi extends Construct {
...
+      console.warn(
+        `${CDKContextKey.DEPLOYMENT_TYPE} is recommended to be one of (${expectedDeploymentTypes.join(', ')}).\n` +
+          'Consider setting the deployment type by running:\n' +
+          'cdk --context amplify-backend-type=sandbox\n' +
+          'or\n' +
+          'cdk --context amplify-backend-type=branch\n',
+      );
+    }
+
+    // Check namespace and name. Throw an error, if ONLY one of them is set, .
```
palpatim, line 311: punctuation nit :)

```suggestion
    // Check namespace and name. Throw an error if ONLY one of them is set.
```

packages/amplify-graphql-api-construct/src/amplify-graphql-api.ts
palpatim: You should also return if we don't have a deployment type at all, right?

bobbyu99: > You should go ahead and return if we encounter this condition, since we can't safely proceed with an unknown deployment type.

If we return here on an unknown deployment type, we would not be checking namespace and name right?

packages/amplify-graphql-api-construct/src/amplify-graphql-api.ts
sobolk: As far as I recall data construct was supposed to be usable outside of amplify in raw CDK apps.

Should this logic be 
```
if (deploymentType) {
  if(deploymentType is invalid) {
   // If we're in Amplify Gen2 app, customer doesn't control these keys, fail fast.
   throw Error();
  }
} else {
   // you are in raw CDK app, none of Amplify CDK context keys is present, which is fine, default to something reasonable.
}
```

packages/amplify-graphql-api-construct/src/amplify-graphql-api.ts
bobbyu99: > You should also return if we don't have a deployment type at all, right?

Yeah i guess in that case returning an empty deploymentIdentifier would be appropriate

packages/amplify-graphql-api-construct/src/amplify-graphql-api.ts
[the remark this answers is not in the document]
edwardfoyle: I'm not going to block on it, but my recommendation is YAGNI. It is easy enough to add it in the future that I would punt until necessary.

packages/amplify-graphql-api-construct/src/amplify-graphql-api.ts
```diff
@@ -275,6 +289,45 @@ export class AmplifyGraphqlApi extends Construct {
...
+    const expectedDeploymentTypes = [this.SANDBOX_DEPLOYMENT_TYPE, this.BRANCH_DEPLOYMENT_TYPE];
+
+    // Log a warning if the deployment type is neither sandbox nor branch.
+    if (!expectedDeploymentTypes.includes(deploymentType)) {
+      console.warn(
+        `${CDKContextKey.DEPLOYMENT_TYPE} is recommended to be one of (${expectedDeploymentTypes.join(', ')}).\n` +
+          'Consider setting the deployment type by running:\n' +
+          'cdk --context amplify-backend-type=sandbox\n' +
+          'or\n' +
+          'cdk --context amplify-backend-type=branch\n',
```
edwardfoyle, line 307: I think this warning is going to cause more confusion than clarity. We don't intend for customers to interact directly with these context values so they aren't going to know what this mean. Also, saying "deployment type is recommended to be... consider setting..." is a lot of weasel words and immediately raises the question "what are the implications of setting or not setting?". If I saw this warning I would just be confused and probably ignore it

edwardfoyle: We also already have [this mechanism](https://github.com/aws-amplify/amplify-backend/blob/d8b43d252c7d5553bf719dd889571807098adf13/packages/backend-data/src/factory.ts#L264-L272) for changing GSI update behavior in sandbox vs non-sandbox. Perhaps we could follow a similar pattern to inject necessary context into the data construct rather than the construct trying to fetch it.

palpatim (dismissed):

packages/amplify-graphql-api-construct/src/amplify-graphql-api.ts
```diff
@@ -275,6 +294,89 @@ export class AmplifyGraphqlApi extends Construct {
...
+    if (sandboxDeploymentIdentifier) {
+      return sandboxDeploymentIdentifier;
+    }
+
+    const branchDeploymentIdentifier = this.makeBranchDeploymentIdentifier(deploymentType);
+    if (branchDeploymentIdentifier) {
+      return branchDeploymentIdentifier;
+    }
+
+    console.warn(`Encountering an unknown deployment type: ${deploymentType}.\n`);
```
palpatim, line 317: nit (for future PR is fine): Change verbiage, remove redundant newline: 
```suggestion
    console.warn(`Encountered an unknown deployment type: ${deploymentType}.`);
```

packages/amplify-graphql-api-construct/src/amplify-graphql-api.ts
palpatim: > We also already have [this mechanism](https://github.com/aws-amplify/amplify-backend/blob/d8b43d252c7d5553bf719dd889571807098adf13/packages/backend-data/src/factory.ts#L264-L272) for changing GSI update behavior in sandbox vs non-sandbox. Perhaps we could follow a similar pattern to inject necessary context into the data construct rather than the construct trying to fetch it?

We'd actually prefer to go the opposite direction, and use sandbox detection to drive the GSI update behavior, rather than make CLI flags privy to more of the transformer internals. We'd have to use aspects to hook onto a specific part of the target stack (e.g., the code you link has to know about the existence of `AmplifyDynamoDbTableWrapper`), and so the CLI now has to be tightly coupled to some part of the underlying stack. Context seems to me the right way for specifying environmental information, and to let the consuming stack do whatever it needs to.

edwardfoyle: I think the "most correct" thing to do is to expose this behavior and the GSI behavior as a prop on the DataConstruct. Then the `backend-data` package can decide how to instantiate the construct based on the Amplify context it knows about and CDK consumers can do the same. This solution isn't eliminating tight coupling, it's just moving it. You are depending on the name and values of the CDK context set by `amplify-backend` without an explicit interface guarding it.

palpatim: Fair point about "moving the coupling without an interface."

Issue #2730, opened 2024-07-30
DeploymentAwaiter fails to update in a project with serverless containers enabled


ashphy:
### How did you install the Amplify CLI?

npm

### If applicable, what version of Node.js are you using?

v20.13.1

### Amplify CLI Version

12.12.4

### What operating system are you using?

macOS Sonoma 14.5

### Did you make any manual changes to the cloud resources managed by Amplify? Please describe the changes made.

No manual changes made

### Describe the bug

In a project with [Serverless containers](https://docs.amplify.aws/gen1/react/tools/cli/usage/containers/) enabled, updating a backend deployed with Amplify CLI v12.12.3 or earlier to Amplify CLI v12.12.4 and then running `amplify push` results in a failure to update DeploymentAwaiter.

Context:
- DeploymentAwaiter is deployed using a CloudFormation Custom Resource.
- Previously ran on Node 16, changed to Node 18 in #2529.
- [Node 18 runtime image does not include aws-sdk v2](https://aws.amazon.com/jp/blogs/compute/node-js-18-x-runtime-now-available-in-aws-lambda/), requiring updated source code aws-amplify/amplify-cli#13757.

Issue:
- The runtime is updated to Node 18, but the Lambda source code is not updated during `amplify push`, causing the following error:

```
INIT_START Runtime Version: nodejs:18.v31	Runtime Version ARN: arn:aws:lambda:us-west-2::runtime:XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
2024-07-30T12:16:16.780Z	undefined	ERROR	Uncaught Exception 	{
    "errorType": "Runtime.ImportModuleError",
    "errorMessage": "Error: Cannot find module 'aws-sdk'\nRequire stack:\n- /var/task/outbound.js\n- /var/task/cfn-response.js\n- /var/task/framework.js\n- /var/runtime/index.mjs",
    "stack": [
        "Runtime.ImportModuleError: Error: Cannot find module 'aws-sdk'",
        "Require stack:",
        "- /var/task/outbound.js",
        "- /var/task/cfn-response.js",
        "- /var/task/framework.js",
        "- /var/runtime/index.mjs",
        "    at _loadUserApp (file:///var/runtime/index.mjs:1087:17)",
        "    at async UserFunction.js.module.exports.load (file:///var/runtime/index.mjs:1119:21)",
        "    at async start (file:///var/runtime/index.mjs:1282:23)",
[...]

PR #2732, opened 2024-07-31
fix: update awaiter s3 key

dpilch:
#### Description of changes

With the CDK version upgrade the runtime version of the custom resource lambda changed from 16 to 18. With the runtime change the lambda code needed be changed to account for breaking changes between 16 and 18. However, the S3 key for the zip file containing the lambda code was not changed, even though the hash of code has changed. During the CFN deploy if the S3 key for the lambda code is unchanged the lambda code will not be updated. 

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-lambda-function-code.html

> Changes to a deployment package in Amazon S3 or a container image in ECR are not detected automatically during stack updates. To update the function code, change the object key or version in the template.

This resulted in the runtime version being upgraded to 18, but the code was not updated (and stayed with the runtime version 16 compatible implementation). The lambda would then throw errors when executed.

This change updates the lambda code s3 key to append the runtime version. This will cause the next deploy to update the lambda code because the s3 key has changed.

Change in CLI repo: https://github.com/aws-amplify/amplify-cli/pull/13872

Awaiter function code S3 key updated to include node runtime version.

#### Issue #, if available

https://github.com/aws-amplify/amplify-category-api/issues/2730

Manual testing. This issue only presents itself when upgrading from an earlier version of the CLI with an existing awaiter function deployed. We do not have an E2E testing framework that allows multiple versions of the CLI to be used. Adding this testing would delay the release of this fix. The tests can be added at a later time.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-category-api/src/provider-utils/awscloudformation/base-api-stack.ts
+++ b/packages/amplify-category-api/src/provider-utils/awscloudformation/base-api-stack.ts
@@ -15,7 +15,7 @@ import { NETWORK_STACK_LOGICAL_ID } from '../../category-constants';
 import Container from './docker-compose/ecs-objects/container';
 import { GitHubSourceActionInfo, PipelineWithAwaiter } from './pipeline-with-awaiter';
 
-const PIPELINE_AWAITER_ZIP = 'custom-resource-pipeline-awaiter.zip';
+const PIPELINE_AWAITER_ZIP = 'custom-resource-pipeline-awaiter-18.zip';
 
 export enum DEPLOYMENT_MECHANISM {
   /**
--- a/packages/amplify-category-api/src/__tests__/provider-utils/awscloudformation/__snapshots__/ecs-apigw-stack.test.ts.snap
+++ b/packages/amplify-category-api/src/__tests__/provider-utils/awscloudformation/__snapshots__/ecs-apigw-stack.test.ts.snap
@@ -204,7 +204,7 @@ Object {
       "Type": "String",
     },
     "awaiterS3Key": Object {
-      "Default": "custom-resource-pipeline-awaiter.zip",
+      "Default": "custom-resource-pipeline-awaiter-18.zip",
       "Type": "String",
     },
     "deploymentBucketName": Object {


packages/amplify-category-api/src/provider-utils/awscloudformation/base-api-stack.ts
```diff
@@ -15,7 +15,7 @@ import { NETWORK_STACK_LOGICAL_ID } from '../../category-constants';
 import Container from './docker-compose/ecs-objects/container';
 import { GitHubSourceActionInfo, PipelineWithAwaiter } from './pipeline-with-awaiter';
 
-const PIPELINE_AWAITER_ZIP = 'custom-resource-pipeline-awaiter.zip';
+const PIPELINE_AWAITER_ZIP = 'custom-resource-pipeline-awaiter-18.zip';
```
sdstolworthy, line 18: Does this need to be parameterized? Under what conditions would this number need to change in the future? What is 18 referring to?

dpilch: This name would need to change anytime [this file](https://github.com/aws-amplify/amplify-cli/blob/dev/packages/amplify-provider-awscloudformation/resources/custom-resource-pipeline-awaiter.zip) is modified in the CLI repo. 18 is referring the node runtime version. The zip file was updated to be compatible with Node 18.

I think parametrization would be possible but I think would require large changes to the API category and the CLI.

I chose the option to simply update the name (which will update the S3 key) because it would be the fastest method to release a fix for this while still being effective.

phani-srikar: This is the name of the key as stored in S3 and this naming convention is shared b/w CLI and API plugins [reference](https://github.com/aws-amplify/amplify-cli/blob/5ec5d4d6678a345346aad2f4a4fe5ba5b518b08c/packages/amplify-provider-awscloudformation/src/resource-package/resource-packager.ts#L60) with the assumption on CLI side that the cloudformation provider would upload the lambda code in the deployment bucket using that key and we use it as part of the code pipeline state machine in our REST API deployments.

sdstolworthy: Node 18 is EOL in 9 months. How do we prevent missing this 9 months from now? https://github.com/nodejs/Release

phani-srikar: This key doesn't require a change for every Lambda runtime update but when the underlying code is not compatible with older versions which coincidentally the case from 16.x -> 18.x runtime since lambda changed it's default support of AWS SDK version from 2.x to 3.x as part of this runtime update [reference](https://aws.amazon.com/blogs/compute/node-js-18-x-runtime-now-available-in-aws-lambda/). We should have an E2E test added to the migration tests that CLI has so that we can catch such cases where a runtime update requires code change. We could do this as a follow-up after the fix is verified to work for this case.

palpatim (approved):

phani-srikar (approved):

PR #2735, opened 2024-07-31
chore: update CDK postgres E2Es to use Data API

fuelvin1:
#### Description of changes

- Adds new methods for alternate flow using the Aurora serverless V2 Postgres clusters as part of setup and teardown.
- Uses Data API to provision test database and tables.

Update `sql-models-2` test to use postgres and verified it passes.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-data-construct/.jsii
```diff
@@ -3731,6 +3731,6 @@
     }
   },
   "types": {},
-  "version": "1.9.4",
-  "fingerprint": "yar8i9UOGPXL5F0IgqsPFOL68J/RpxMvEnFpAli1gME="
+  "version": "1.9.3",
```
sundersc, line 3734: Looks you may need to rebase with main branch.

packages/amplify-graphql-api-construct-tests/rds-test-config.json
phani-srikar, line 1: I believe we don't need this file?

packages/amplify-graphql-api-construct-tests/src/__tests__/relationships/postgres-uuid-pk/uuid-pk-ddbprimary-sqlrelated.test.ts
```diff
@@ -23,8 +23,9 @@ describe('PostgreSQL tables with UUID primary keys', () => {
   const region = process.env.CLI_REGION ?? 'us-west-2';
   const baseProjFolderName = path.basename(__filename, '.test.ts');
 
+  // sufficient password length that meets the requirements for RDS cluster/instance
```
phani-srikar, line 26: nit: here and other places - don't think this comment is relevant here, since we're generating DB Name.

packages/amplify-e2e-core/src/utils/rds.ts
```diff
@@ -126,6 +134,125 @@ export const createRDSInstance = async (
...
+export const createRDSCluster = async (
+  config: RDSConfig,
+): Promise<{
+  clusterArn: string;
+  endpoint: string;
+  port: number;
+  dbName: string;
+  password: string;
+  managedSecretArn: string;
+  dbInstance: DBInstance;
```
palpatim, line 151: Consider a named structure (e.g., a typealias or interface) that encompasses these fields, to centralize the definition and make future maintenance easier.

packages/amplify-e2e-core/src/utils/rds.ts
```diff
@@ -126,6 +134,125 @@ export const createRDSInstance = async (
...
+}> => {
+  const rdsClient = new RDSClient({ region: config.region });
+  const initialDBName = 'defaultdb';
+
+  const params: CreateDBClusterMessage = {
+    /** input parameters */
+    EnableHttpEndpoint: true,
+    Engine: 'aurora-postgresql',
```
palpatim, line 159: Since we know we'll support mysql soon, let's make this an argument to the `createRDSCluster` function rather than hardcoding it. We might not want our tests to directly use the Engine value of `aurora-postgresql`, but we have a number of other specifiers for DB type / DB engine that we could use. (If this wasn't a requirement we knew was coming soon, I'd agree with your hardcoded approach--[YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it)!)

packages/amplify-e2e-core/src/utils/rds.ts
```diff
@@ -126,6 +134,125 @@ export const createRDSInstance = async (
...
+    );
+
+    if (availableInstanceResponse.state !== 'SUCCESS') {
+      throw new Error('Error in creating a new RDS instance inside the cluster.');
+    }
+
+    let password = config.password;
+    let masterUserSecret;
+
+    if (!config.password) {
```
palpatim, line 227: It looks like this method is only ever called from test invocations that use Data API. If that is the case, we should be using the RDS cluster master password ARN, and never returning the password.

If I'm mistaken, consider refactoring this function into 2 different functions for each use case, to minimize nested if/thens and make the return values purpose-built for the use case.

packages/amplify-e2e-core/src/utils/rds.ts
```diff
@@ -185,6 +312,85 @@ export const setupRDSInstanceAndData = async (
...
+    throw new Error('Failed to store db connection config in secrets manager');
+  }
+
+  const client = new RDSDataClient({ region: config.region });
+
+  // create a new test database with given name
+  const createDBInput: ExecuteStatementCommandInput = {
+    resourceArn: dbCluster.clusterArn,
+    secretArn,
+    sql: `create database ${config.dbname};`,
```
palpatim, line 355: Rework this to include dbname as a parameter, to avoid potential SQL injection if dbname is ever compromised. (Plus, it's a good habit to get into to not interpolate into SQL if you can avoid it). In this case, the risk is minimal since we generate the dbname, but if in future we decided "hey, let's let dbname be specified by an environment variable", then we'd be open to executing SQL with untrusted input.

packages/amplify-e2e-core/src/utils/rds.ts
```diff
@@ -185,6 +312,85 @@ export const setupRDSInstanceAndData = async (
...
+  });
+
+  return {
+    endpoint: dbCluster.endpoint,
+    port: dbCluster.port,
+    dbName: dbCluster.dbName,
+    dbInstance: dbCluster.dbInstance,
+    password: dbCluster.password,
+    secretArn: secretArn,
+    managedSecretArn: dbCluster?.managedSecretArn,
```
palpatim, line 390: What is the difference between secretArn and managedSecretArn? How can we make that difference easier to understand just by looking at the names?

packages/amplify-e2e-core/src/utils/rds.ts
```diff
@@ -221,6 +427,39 @@ export const deleteDBInstance = async (identifier: string, region: string): Prom
...
+    DBClusterIdentifier: identifier,
+    SkipFinalSnapshot: true,
+  };
+  console.log(`Deleting cluster: ${identifier}`);
+  const command = new DeleteDBClusterCommand(params);
+  try {
+    await client.send(command);
+  } catch (error) {
+    console.log(error);
+    throw new Error(`Error in deleting RDS cluster: ${identifier}`);
```
palpatim, line 453: Should we include the body of `error`, rather than just logging it? The error message is useful for troubleshooting, and will be surfaced to the test runner rather than buried in logs (IIRC).

packages/amplify-e2e-core/src/utils/rds.ts
```diff
@@ -185,6 +312,85 @@ export const setupRDSInstanceAndData = async (
...
+ */
+export const setupRDSClusterAndData = async (
+  config: RDSConfig,
+  queries?: string[],
+): Promise<{
+  endpoint: string;
+  port: number;
+  dbName: string;
+  dbInstance: DBInstance;
+  password: string;
```
palpatim, line 330: As above, we shouldn't need to return password for Data API tests

phani-srikar (approved):

packages/amplify-e2e-core/src/utils/rds.ts
```diff
@@ -185,6 +301,69 @@ export const setupRDSInstanceAndData = async (
...
+  const client = new RDSDataClient({ region: config.region });
+
+  // create a new test database with given name
+  const sanitizedDbName = config.dbname.replace(/[^a-zA-Z0-9_]/g, '');
+
+  const createDBInput: ExecuteStatementCommandInput = {
+    resourceArn: dbCluster.clusterArn,
+    secretArn: dbCluster.secretArn,
+    sql: `create database ${sanitizedDbName}`,
+    database: dbCluster.dbName,
```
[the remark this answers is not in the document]
fuelvin1: Yes, that's correct

packages/amplify-e2e-core/src/utils/rds.ts
```diff
@@ -217,10 +396,43 @@ export const deleteDBInstance = async (identifier: string, region: string): Prom
...
+ * @param identifier RDS cluster identifier to delete
+ * @param region RDS cluster region
+ */
+export const deleteDBCluster = async (identifier: string, region: string): Promise<void> => {
+  // First the instance deletion is triggered
+  const instanceID = createInstanceIdentifier(identifier);
+  console.log(`Deleting instance: ${instanceID}`);
+  await deleteDBInstance(instanceID, region);
+
+  // Now delete the cluster
```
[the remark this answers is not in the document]
fuelvin1: I can confirm it doesn't throw an error, as far as I know.

palpatim (approved):

phani-srikar (approved):

packages/amplify-e2e-core/src/utils/rds.ts
```diff
@@ -118,14 +143,105 @@ export const createRDSInstance = async (
...
+    EnableHttpEndpoint: true,
+    Engine: getRDSEngineType(config.engine),
+    DatabaseName: initialDBName,
+    DBClusterIdentifier: config.identifier,
+    MasterUsername: config.username,
+    // use RDS managed password, then retrieve the password and store in all other credential store options
+    ManageMasterUserPassword: true,
+    ServerlessV2ScalingConfiguration: {
+      MinCapacity: 4,
+      MaxCapacity: 10,
```
sundersc, line 174: As this is only a test resource, can we lower these values?

sundersc (approved):

PR #2741, opened 2024-08-02
Upgrade fast-xml-parser

dpilch:
#### Description of changes

The package.json `resolutions` is incorrectly resolving the dependabot alerts. The `resolutions` will resolve the install with local development in the monorepo. However when packages are installed by customers the resolution will not be used because the package is not aware of the monorepo resolutions.

This was ok with the Gen 1 CLI because the CLI was shipped as a binary. This allowed the CLI to apply resolutions of its own. With Gen 2 the NPM dep chain will be respected when customers install the packages.

When customers are installing Gen 2 the `aws-amplify/graphql-schema-generator` will pull in the listed fast-xml-parser version (not what is set in `resolutions`).

---

Properly applying the fix required several changes:
1. Remove fast-xml-parser resolution

2. Set the correct version of Amplify backend packages. 

The amplify-backend packages released a major version bump with GA release. We are still using the pre-GA versions. The exports we are using for these packages have not changed, but we would not get any updates with the current semver version range.

The older versions of amplify-backend will continue to use the previous versions of fast-xml-parser.

3. Upgrade Jest

The new amplify-backend packages are importing `node:url`. The previous version of `ts-jest` did not understand this import. Jest has been upgraded to v29. Changes were necessary for `useFakeTimers` and `snapshotFormat`.

There is a performance degradation with newer versions of Jest when run with node < 20 (https://github.com/jestjs/jest/issues/11956). We currently support Node 18. Node 18 will be EOL in 8 months, so we should look into upgrading to Node 20.

4. Upgrade AWS SDK versions. This will pull in the latest fast-xml-parser version.

---

With these changes there are still some traces of the previous versions of fast-xml-parser.

Requires a amplify-backend release (blocked by this change). Customers installing Gen 2 will will install the latest version of amplify-backend, which will have the latest version of fast-xml-parser. This can be resolved in this repo after the amplify-backend release.
```
├─┬ @aws-amplify/data-construct@1.9.4 -> ./packages/amplify-data-construct
│ └─┬ @aws-amplify/platform-core@1.0.2
│   └─┬ @aws-sdk/client-sts@3.614.0
│     └─┬ @aws-sdk/core@3.614.0
│       └── fast-xml-parser@4.2.5 deduped
```

An older version of the Amplify runtime libraries are used for our E2E tests. These tests need to be upgraded to use the latest runtime version.
```
├─┬ amplify-category-api-e2e-tests@3.22.8 -> ./packages/amplify-e2e-tests
[...]

PR #2747, opened 2024-08-07
chore: upgrade jest to 29

dpilch:
#### Description of changes

Upgrade Jest to 29. This is necessary to upgrade amplify-backend packages to GA versions.
* Update Jest config to account for breaking changes in order to keep same functionality.
* Lower some coverage thresholds that are reported differently with new version of jest.
* Move jest.setTimeout out of beforeAll https://github.com/jestjs/jest/issues/9359
* Increase codebuild e2e test instance sizes. Set max workers to 5.
    * Jest has a memory issue when used with node version <= 18. https://github.com/jestjs/jest/issues/11956. We will need to upgrade to Node 20 soon.
    * I saw each test suite ranging between low 2.x GB and low 3.x GB. With 5 test suites per instance, tests are now around 80% memory consumption. It might be possible to set max workers to 6, but the instances would potentially be less stable.
    * Multi test instances are using `BUILD_GENERAL1_LARGE` and single test instances are using `BUILD_GENERAL1_MEDIUM`.
    * [Instance resources](https://docs.aws.amazon.com/codebuild/latest/userguide/build-env-ref-compute-types.html)
    * [Instance cost](https://aws.amazon.com/codebuild/pricing/)

N/A

#### Issue #, if available

https://github.com/aws-amplify/amplify-category-api/pull/2741

* Unit test
* E2E test

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-e2e-core/src/utils/retrier.ts
```diff
@@ -4,7 +4,7 @@ import { sleep } from './sleep';
 const defaultSettings: RetrySettings = {
   times: Infinity,
   delayMS: 1000 * 10, // 10 seconds
-  timeoutMS: 1000 * 60 * 20, // 20 minutes
+  timeoutMS: 1000 * 60 * 25, // 25 minutes
```
stocaaro, line 7: What did you encounter that caused the bump from 20 to 25 minutes on the timeout?

I would guess this means our e2e tests slowed down by ~25%, which might be enough to be worth an investigation. If you have already done the investigation, whats driving the increase and can we do better?

stocaaro (approved):
LGTM.

I noted the timeout increase, which raises questions that are probably answerable, more than need changes.

packages/amplify-e2e-core/src/utils/retrier.ts
dpilch: There is a performance degradation with newer versions of Jest when run with node < 20 (https://github.com/jestjs/jest/issues/11956). We currently support Node 18. Node 18 will be EOL in 8 months, so we should look into upgrading to Node 20.

codebuild_specs/run_cdk_tests.yml
```diff
@@ -6,7 +6,6 @@ env:
     AMPLIFY_PATH: /root/.npm-global/lib/node_modules/@aws-amplify/cli-internal/bin/amplify
     CI: true
     CODEBUILD: true
-    NODE_OPTIONS: --max-old-space-size=8096
```
dpilch, line 9: These were ignored.

scripts/split-e2e-tests.ts
```diff
@@ -296,14 +296,16 @@ const splitTests = (
...
+      tmp.env.variables = tmp.env.variables ?? {};
       tmp.env.variables.TEST_SUITE = j.tests.join('|');
       tmp.env.variables.CLI_REGION = j.region;
       if (j.useParentAccount) {
         tmp.env.variables.USE_PARENT_ACCOUNT = 1;
       }
       if (j.runSolo) {
-        tmp.env['compute-type'] = 'BUILD_GENERAL1_SMALL';
+        tmp.env['compute-type'] = 'BUILD_GENERAL1_MEDIUM';
+        // BUILD_GENERAL1_SMALL has 7GB of memory. 6656 = 6.5GB. Leave 0.5GB for the OS and other processes.
```
palpatim, line 307: Is this comment relevant since we're using MEDIUM? Or does the comment incorrectly specify SMALL when it should be specifying MEDIUM?

scripts/split-e2e-tests.ts
```diff
@@ -379,7 +381,11 @@ const main = (): void => {
         identifier: 'run_e2e_tests',
         buildspec: 'codebuild_specs/run_e2e_tests.yml',
         env: {
-          'compute-type': 'BUILD_GENERAL1_MEDIUM',
+          'compute-type': 'BUILD_GENERAL1_LARGE',
+          variables: {
+            // BUILD_GENERAL1_LARGE has 15GB of memory. 14848MB = 14.5GB. Leave 0.5GB for the OS and other processes.
```
palpatim, line 386: Is 0.5GB enough memory for everything else? I assume you're basing this on some heuristic, manual testing, etc?

dpilch: NPM docs suggest leaving 0.5GB https://nodejs.org/api/cli.html#--max-old-space-sizesize-in-megabytes

scripts/split-e2e-tests.ts
dpilch: Should be medium. I'll update.

palpatim (approved):

phani-srikar (approved):

PR #2748, opened 2024-08-07
 chore: set correct version for amplify-backend deps

dpilch:
#### Description of changes

The amplify-backend packages released a major version bump with GA release. We are still using the pre-GA versions. The exports we are using for these packages have not changed, but we would not get any updates with the current semver version range.

There was an error with the dep chain resolution with the previous amplify-backend packages. JSII needs to bundle all deps in the change that are not a JSII package. Previously `@aws-amplify/platform-core` did not properly list the transitive deps.
https://github.com/aws-amplify/amplify-category-api/blob/4c5fc2aeebfc727a2ef6d02bee93543f732ec996/yarn.lock#L347-L352
This resulted in the verification script to believe that these deps should not be included in the JSII bundled deps. The changes to `noHoist`, `bundled`, and `dependencies` is to account for transitives now being listed correctly. The dep chain is calculated by the `scripts/verify-construct-dependencies.ts` script. https://github.com/aws-amplify/amplify-category-api/blob/e35ecd5c95b519dc743a00273632b3e20f91b4e4/yarn.lock#L366-L377

---

I also needed to explicitly add `jsii-rosetta` as a dev dep because of the following error. It's not clear why changing the amplify-backend deps resulted in needing this. One theory is that one of the newly included bundled deps require this package explicitly listed.

This issue states that `jsii-rosetta` should be the same version as `jsii`. https://github.com/aws/jsii/issues/4501

```
$ jsii-pacmak
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
lerna ERR! yarn run package stderr:
node:internal/modules/cjs/loader:1143
  throw err;
  ^

Error: Cannot find module 'jsii-rosetta'
Require stack:
- /Users/dppilche/amplify/amplify-category-api/node_modules/jsii-pacmak/bin/jsii-pacmak.js
- /Users/dppilche/amplify/amplify-category-api/node_modules/jsii-pacmak/bin/jsii-pacmak
    at Module._resolveFilename (node:internal/modules/cjs/loader:1140:15)
    at Module._load (node:internal/modules/cjs/loader:981:27)
    at Module.require (node:internal/modules/cjs/loader:1231:19)
    at require (node:internal/modules/helpers:177:18)
    at Object.<anonymous> (/Users/dppilche/amplify/amplify-category-api/node_modules/jsii-pacmak/bin/jsii-pacmak.js:5:24)
    at Module._compile (node:internal/modules/cjs/loader:1364:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1422:10)
    at Module.load (node:internal/modules/cjs/loader:1203:32)
    at Module._load (node:internal/modules/cjs/loader:1019:12)
    at Module.require (node:internal/modules/cjs/loader:1231:19) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [
    '/Users/dppilche/amplify/amplify-category-api/node_modules/jsii-pacmak/bin/jsii-pacmak.js',
    '/Users/dppilche/amplify/amplify-category-api/node_modules/jsii-pacmak/bin/jsii-pacmak'
  ]
}

```

---

The final change was to pin the version of `@microsoft/api-extractor` due to a change of the default behavior. https://github.com/microsoft/rushstack/issues/4842. It's not clear why we are only encountering the issue now rather than when the change was first released. Pinning the version is the same action taken by amplify-backend. https://github.com/aws-amplify/amplify-backend/pull/1780

* Unit tests
* E2E tests

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-graphql-api-construct-tests/src/__tests__/sql-iam-access.test.ts
```diff
@@ -58,13 +58,13 @@ describe('CDK SQL Iam Access', () => {
...
     const name = await initCDKProject(projRoot, templatePath, {
-      additionalDependencies: ['@aws-amplify/auth-construct-alpha@^0.5.6'],
+      additionalDependencies: ['@aws-amplify/auth-construct@^1.0.0'],
       cdkContext: {
         'enable-iam-authorization-mode': 'false',
       },
     });
     const nameWithIam = await initCDKProject(projRootWithIam, templatePath, {
-      additionalDependencies: ['@aws-amplify/auth-construct-alpha@^0.5.6'],
+      additionalDependencies: ['@aws-amplify/auth-construct@^1.0.0'],
```
phani-srikar, line 67: It would help to define this dependency as a constant that can be used across tests and save us from making changes in multiple places in the future.

palpatim (approved):
Approved pending convo on data client bundle size

phani-srikar (approved):

PR #2768, opened 2024-08-13
fix: change gen 1 patterns to warning message instead of error

dpilch:
#### Description of changes

* change gen 1 pattern to warning message instead of error
* remove `_allowGen1Patterns` from the translation behavior config and always disallow gen 1 patterns in the construct. Gen 1 patterns will always be allowed in Gen 1.
    * We marked this as internal and stated that breaking changes could happen without a major version so we are ok to change this.

* remove `_allowGen1Patterns` from the translation behavior config and always disallow gen 1 patterns in the construct.

#### Issue #, if available

N/A

N/A

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-graphql-predictions-transformer/src/graphql-predictions-transformer.ts
```diff
@@ -153,6 +153,12 @@ export class PredictionsTransformer extends TransformerPluginBase {
   };
 
   generateResolvers = (context: TransformerContextProvider): void => {
+    // This validation can't occur in validate because the api has not been initialized until generateResolvers
+    if (!context.transformParameters.allowGen1Patterns) {
+      cdk.Annotations.of(context.api).addWarning(
```
dpilch, line 158: Annotations also has `addDeprecation`. I chose to use `addWarning` here because `addDeprecation` appears to be geared towards deprecating an entire parameter. We are removing some functionality from a parameter.

palpatim (approved):

palpatim (approved):

phani-srikar (approved):

PR #2778, opened 2024-08-15
chore: add sql local testing setup

fuelvin1:
#### Description of changes

- Implemented Sql local testing functionality using a pre-provisioned cluster's identifier
- Generated and sanitized database name in the centralized database controller, instead of in each test

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-e2e-core/src/utils/rds.ts
```diff
@@ -5,11 +5,16 @@ import {
   CreateDBInstanceCommandInput,
   DBInstance,
   DeleteDBInstanceCommand,
+  DescribeDBClustersCommand,
   waitUntilDBInstanceAvailable,
   CreateDBClusterMessage,
   waitUntilDBClusterAvailable,
   DeleteDBClusterCommand,
   DeleteDBClusterCommandInput,
+  $Command,
```
palpatim, line 14: Where is this used?

packages/amplify-e2e-core/src/utils/rds.ts
```diff
@@ -241,6 +253,57 @@ export const createRDSCluster = async (config: RDSConfig): Promise<ClusterInfo>
...
 
+/**
+ * Accesses the local RDS Aurora serverless V2 cluster with one DB instance using the given input configuration.
+ * @param config Configuration of the database cluster.
+ * @returns EndPoint address, port and database name of the accessed RDS cluster.
+ */
+export const useRDSCluster = async (config: RDSConfig): Promise<ClusterInfo> => {
+  try {
+    // Send requests to get the cluster and instance config info
+    const databaseName = 'defaultdb';
```
palpatim, line 264: Is a hardcoded db name the right way to go here? What happens if multiple tests (e.g., one from a local machine and one from CICD) run in the same account & same region? Wouldn't they overwrite each other?

packages/amplify-e2e-core/src/utils/rds.ts
```diff
@@ -5,11 +5,16 @@ import {
...
+  DescribeDBClustersCommand,
   waitUntilDBInstanceAvailable,
   CreateDBClusterMessage,
   waitUntilDBClusterAvailable,
   DeleteDBClusterCommand,
   DeleteDBClusterCommandInput,
+  $Command,
+  DescribeDBInstancesCommand,
+  DescribeDBClustersCommandOutput,
+  DescribeDBInstancesCommandOutput,
```
palpatim, line 17: nit here and throughout: Please sort the imports :) It makes it easier to keep track of what is already imported

packages/amplify-e2e-core/src/utils/rds.ts
```diff
@@ -35,11 +45,13 @@ const DEFAULT_SECURITY_GROUP = 'default';
 const IPIFY_URL = 'https://api.ipify.org/';
 const AWSCHECKIP_URL = 'https://checkip.amazonaws.com/';
 
+let clusterInfo: ClusterInfo;
```
palpatim, line 48: I'm not wild about this being a global -- it means that subsequent invocations of this method could mutate the value even though other processes may be relying on it. Why do we need to preserve this state here?

packages/amplify-e2e-core/src/utils/rds.ts
```diff
@@ -308,24 +371,29 @@ export const setupRDSInstanceAndData = async (
...
+    dbCluster = await useRDSCluster(config);
+    console.log(`Using RDS ${config.engine} DB cluster with identifier ${config.identifier}`);
+  }
 
   if (!dbCluster.secretArn) {
     throw new Error('Failed to store db connection config in secrets manager');
   }
 
-  const client = new RDSDataClient({ region: config.region });
+  clusterInfo = dbCluster;
```
palpatim, line 389: See note above about mutating global state

packages/amplify-e2e-core/src/utils/rds.ts
```diff
@@ -241,6 +253,57 @@ export const createRDSCluster = async (config: RDSConfig): Promise<ClusterInfo>
   }
 };
 
+/**
+ * Accesses the local RDS Aurora serverless V2 cluster with one DB instance using the given input configuration.
+ * @param config Configuration of the database cluster.
+ * @returns EndPoint address, port and database name of the accessed RDS cluster.
+ */
+export const useRDSCluster = async (config: RDSConfig): Promise<ClusterInfo> => {
```
palpatim, line 261: I'm confused about this method -- you named it "use" but it appears to only be gathering information about the cluster, not establishing connections or anything. What is the actual purpose, and what name would more accurately reflect its intent? Similarly, the comment above "accesses the local cluster..." isn't helpful. Why is it accessing the cluster--to validate? To retrieve? To establish a long-term connection?

packages/amplify-e2e-core/src/utils/rds.ts
```diff
@@ -308,24 +371,29 @@ export const setupRDSInstanceAndData = async (
  * @returns Endpoint address, port and database name of the created RDS cluster.
  */
 
-export const setupRDSClusterAndData = async (config: RDSConfig, queries?: string[]): Promise<ClusterInfo> => {
-  console.log(`Creating RDS ${config.engine} DB cluster with identifier ${config.identifier}`);
+export const setupRDSClusterAndData = async (useLocalCluster: boolean, config: RDSConfig, queries?: string[]): Promise<ClusterInfo> => {
```
palpatim, line 374: Continuing the above discussion of `clusterInfo` as global state -- this method will set the `clusterInfo` global as a side effect, but it doesn't need to -- the method actually returns cluster info. Consider removing the global and making callers responsible for retaining the value so they can pass it to subsequent calls that need it.

packages/amplify-e2e-core/src/utils/rds.ts
```diff
@@ -344,25 +412,55 @@ export const setupRDSClusterAndData = async (config: RDSConfig, queries?: string
...
+      secretArn: dbCluster.secretArn,
+      sql: `drop database ${config.dbname} with (FORCE)`,
+      database: dbCluster.dbName,
+    };
+    const executeStatementResponse = await client.send(new ExecuteStatementCommand(executeStatementInput));
+    console.log('Run query response: ' + JSON.stringify(executeStatementResponse));
+  } catch (err) {
+    throw new Error(`Error in running queries in test database: ${err}`);
+  }*/
+
```
palpatim, line 436: Unless there's a pretty compelling reason to retain this (e.g., you're going to reinstate it soon), please remove dead code so it doesn't clutter up the codebase.

packages/amplify-e2e-core/src/utils/rds.ts
```diff
@@ -422,7 +520,7 @@ export const deleteDBCluster = async (identifier: string, region: string): Promi
     await client.send(command);
   } catch (error) {
     console.log(error);
-    throw new Error(`Error in deleting RDS cluster ${identifier}: ${error.response.json}`);
+    throw new Error(`Error in deleting RDS cluster ${identifier}: ${error.json}`);
```
palpatim, line 523: How do we know the incoming error has a `json` property? If it doesn't will we lose meaningful troubleshooting context?

scripts/e2e-test-local-cluster-config.json
```diff
@@ -0,0 +1,16 @@
+{
```
palpatim, line 1: 1. Consider renaming this to `scripts/e2e-test-local-cluster-config.json.SAMPLE` or similar, so other callers don't use it by accident and connect to a non-existent database. It would then be a local tester's responsibility to copy the sample file to the correct name.
2. If you haven't already got a task for it, please update one of the READMEs in this repo and our runbook with instructions for local testing. :)

scripts/e2e-test-local-cluster-config.json
```diff
@@ -0,0 +1,16 @@
...
+    {
+      "dbConfig": {
+        "identifier": ""
+      }
+    }
+  ],
+  "us-west-2": [
+    {
+      "dbConfig": {
+        "identifier": "YMmwoSDbAQ"
```
palpatim, line 12: Clear this value to reduce confusion. Alternately, set it to a clearly bogus value to indicate that it must be replaced. Whatever you decide, please make sure to do the same for the us-east-1 value above

packages/amplify-e2e-core/src/utils/rds.ts
fuelvin1: Done - got rid of the hardcoded db name.

scripts/e2e-test-local-cluster-config.json
phani-srikar: Replace with sample values like Tim suggested and can we rename this file to `e2e-test-local-cluster-config.sample.json` and also add a `.gitignore` entry for `e2e-test-local-cluster-config.json` so developers don't commit it accidentally?

packages/amplify-e2e-core/src/utils/rds.ts
phani-srikar: You can just drop the database or I noticed that we don't drop the databases in CI flow (for reasons you mentioned earlier), so probably just keep the same behavior for now and we can add that later.

packages/amplify-e2e-core/src/utils/rds.ts
```diff
@@ -241,6 +253,57 @@ export const createRDSCluster = async (config: RDSConfig): Promise<ClusterInfo>
...
+ * Accesses the local RDS Aurora serverless V2 cluster with one DB instance using the given input configuration.
+ * @param config Configuration of the database cluster.
+ * @returns EndPoint address, port and database name of the accessed RDS cluster.
+ */
+export const useRDSCluster = async (config: RDSConfig): Promise<ClusterInfo> => {
+  try {
+    // Send requests to get the cluster and instance config info
+    const databaseName = 'defaultdb';
+    const identifier = config.identifier;
+    const instanceIdentifier = createInstanceIdentifier(config.identifier);
```
phani-srikar, line 266: We cannot assume that the instances developer have in their cluster adhere to this naming. You can get the instance attached to cluster from your describe cluster command.

packages/amplify-graphql-api-construct-tests/src/sql-datatabase-controller.ts
```diff
@@ -51,20 +63,79 @@ export interface SqlDatabaseDetails {
...
+      this.options.dbname = generateDBName().replace(/[^a-zA-Z0-9_]/g, '');
+    }
   }
 
+  setNewConfig = async (): Promise<RDSConfig> => {
+    // Access the specific config from the local cluster configs JSON file
+    const repoRoot = path.join(__dirname, '..', '..', '..');
+    const localClusterPath = path.join(repoRoot, 'scripts', 'e2e-test-local-cluster-config.json');
+    const localClustersObject = JSON.parse(fs.readFileSync(localClusterPath, 'utf-8'));
+    const cluster = localClustersObject[this.options.region][0];
```
phani-srikar, line 88: We need to check if this is empty list before referencing the first element.

packages/amplify-graphql-api-construct-tests/src/sql-datatabase-controller.ts
```diff
@@ -51,20 +63,79 @@ export interface SqlDatabaseDetails {
...
+
+    // Get the config identifier and connection URI
+    const identifier = cluster.dbConfig.identifier;
+
+    if (identifier === '' || identifier === null) {
+      return null;
+    }
+
+    const connectionUri = `/${identifier}/test`;
```
phani-srikar, line 97: nit: rename to `connectionUriSsmPath`

packages/amplify-graphql-api-construct-tests/src/sql-datatabase-controller.ts
```diff
@@ -260,6 +440,7 @@ export class SqlDatatabaseController {
    */
   getConnectionUri = (engine: SqlEngine, username: string, password: string, hostname: string, port: number, dbName: string): string => {
     const protocol = engine === 'postgres' ? 'postgresql' : 'mysql';
-    return `${protocol}://${username}:${password}@${hostname}:${port}/${dbName}`;
+    // ELVIN - TEST
```
phani-srikar, line 443: nit: remove this

packages/amplify-graphql-api-construct-tests/src/sql-datatabase-controller.ts
```diff
@@ -81,110 +152,219 @@ export class SqlDatatabaseController {
...
       const secretsManagerClient = new SecretsManagerClient({ region: this.options.region });
       const secretManagerCommand = new GetSecretValueCommand({
         SecretId: secretArn,
       });
       const secretsManagerResponse = await secretsManagerClient.send(secretManagerCommand);
-      const { password: managedPassword } = JSON.parse(secretsManagerResponse.SecretString);
-      if (!managedPassword) {
-        throw new Error('Unable to get RDS cluster master user password');
+      const secretArnWithCustomKey = secretsManagerResponse.ARN; // arn:aws:secretsmanager:us-west-2:637423428135:secret:yjLvhaVbVT-secret-custom-key-BTqGiz
+      const keyArn = 'arn:aws:kms:us-west-2:637423428135:key/a84fe5fe-ee01-44bf-b94e-57604a109c77'; // ELVIN - GET THE MASTER CREDENTIALS KMS KEY SOMEHOW (like "arn:aws:kms:us-west-2:637423428135:key/62ec591f-7188-49db-8819-87c15727908f")
```
phani-srikar, line 270: Let's scope this down to only handle DB managed secrets for now. Remove these hard-coded values.

phani-srikar (changes requested):

scripts/e2e-test-local-cluster-config.json
```diff
@@ -0,0 +1,18 @@
+{
+  "us-east-1": [
```
phani-srikar, line 2: This file can be removed since you already included a templated sample.

packages/amplify-e2e-core/src/utils/rds.ts
```diff
@@ -241,6 +244,81 @@ export const createRDSCluster = async (config: RDSConfig): Promise<ClusterInfo>
   }
 };
 
+/**
+ * Accesses the local RDS Aurora serverless V2 cluster with one DB instance using the given input configuration.
```
phani-srikar, line 248:
```suggestion
 * Setup the test database and data in the pre-existing RDS Aurora serverless V2 cluster with one writer DB instance. Get the necessary configuration settings of the cluster and instance.
```

packages/amplify-e2e-core/src/utils/rds.ts
```diff
@@ -241,6 +244,81 @@ export const createRDSCluster = async (config: RDSConfig): Promise<ClusterInfo>
...
+ * @param config Configuration of the database cluster.
+ * @returns EndPoint address, port and database name of the accessed RDS cluster.
```
phani-srikar, line 250: nit: Update these doc comments to match updated method definition. You can just say returns the cluster configuration information.

packages/amplify-e2e-core/src/utils/rds.ts
```diff
@@ -309,23 +387,19 @@ export const setupRDSInstanceAndData = async (
  */
 
 export const setupRDSClusterAndData = async (config: RDSConfig, queries?: string[]): Promise<ClusterInfo> => {
-  console.log(`Creating RDS ${config.engine} DB cluster with identifier ${config.identifier}`);
-
   const dbCluster = await createRDSCluster(config);
+  console.log(`Creating RDS ${config.engine} DB cluster with identifier ${config.identifier}`);
```
phani-srikar, line 391: nit: might want to log this before we create the cluster.

packages/amplify-e2e-core/src/utils/rds.ts
```diff
@@ -309,23 +387,19 @@ export const setupRDSInstanceAndData = async (
...
   const client = new RDSDataClient({ region: config.region });
 
-  // create a new test database with given name
-  const sanitizedDbName = config.dbname.replace(/[^a-zA-Z0-9_]/g, '');
-
   const createDBInput: ExecuteStatementCommandInput = {
     resourceArn: dbCluster.clusterArn,
     secretArn: dbCluster.secretArn,
-    sql: `create database ${sanitizedDbName}`,
+    sql: `create database ${config.dbname}`,
```
phani-srikar, line 402: nit: add a note here that is already sanitized when we generate the random database name.

packages/amplify-e2e-core/src/utils/rds.ts
```diff
@@ -395,7 +469,7 @@ export const deleteDBInstance = async (identifier: string, region: string): Prom
     // );
   } catch (error) {
     console.log(error);
-    throw new Error(`Error in deleting RDS instance: ${error.response.json}`);
+    throw new Error(`Error in deleting RDS instance: ${error.json}`);
```
phani-srikar, line 472: here and below: the `.json` might not exist, use `JSON.Stringify(err)` instead.

phani-srikar (approved):

packages/amplify-e2e-core/src/utils/rds.ts
```diff
@@ -241,6 +244,83 @@ export const createRDSCluster = async (config: RDSConfig): Promise<ClusterInfo>
...
+    const describeInstanceResponse = await client.send(describeInstanceCommand);
+    if (!describeInstanceResponse || describeInstanceResponse?.DBInstances?.length === 0) {
+      throw Error('Specified cluster instance info cannot be fetched');
+    }
+
+    const clusterArn = dbClusterObj.DBClusterArn;
+    const secretArn = dbClusterObj.MasterUserSecret.SecretArn;
+    const defaultDbName = dbClusterObj.DatabaseName;
+    const dataClient = new RDSDataClient({ region: config.region });
+    const createDBInput: ExecuteStatementCommandInput = {
```
palpatim, line 279: As mentioned in previous PRs: we need to sanitize the db name before interpolating it into the SQL. I believe you've currently got the sanitization going on in the test code, but it more appropriately belongs here, especially since you're returning the dbName as part of your return value anyway.

packages/amplify-e2e-core/src/utils/rds.ts
palpatim: As noted above, this isn't sufficient, since we export the "create database" method so that other callers use it, and there's nothing enforcing those callers to sanitize the name. We need to be in the habit of sanitizing input at the point of usage, even if we believe it's sanitized elsewhere.

packages/amplify-graphql-api-construct-tests/src/sql-datatabase-controller.ts
```diff
@@ -51,20 +54,34 @@ export interface SqlDatabaseDetails {
...
     if (options.engine === 'postgres' && isDataAPISupported(options.region)) {
       this.useDataAPI = true;
+      this.enableLocalTesting = !isCI() && getClusterIdentifier(options.region, options.engine) !== undefined;
     } else {
       this.useDataAPI = false;
     }
+
+    // If database name not manually set, provide and sanitize the config dbname
+    if (!options.dbname || options.dbname.length == 0 || this.enableLocalTesting) {
+      this.options.dbname = generateDBName().replace(/[^a-zA-Z0-9_]/g, '');
```
palpatim, line 70: as noted elsewhere, we can certainly keep this sanitization here, but we _also_ need to sanitize values before we interpolate them into SQL strings

packages/amplify-graphql-api-construct-tests/src/sql-datatabase-controller.ts
```diff
@@ -114,7 +131,7 @@ export class SqlDatatabaseController {
     };
     console.log(`Stored db connection config in Secrets manager: ${JSON.stringify(dbConnectionConfigSecretsManagerCustomKey)}`);
 
-    const pathPrefix = `/${this.options.identifier}/test`;
+    const pathPrefix = `/${this.options.identifier}/${this.options.dbname}/test`;
```
palpatim, line 134: make sure to use the dbname value coming back from the setup command.

phani-srikar (approved):

palpatim (approved):

PR #2787, opened 2024-08-19
fix: encode password to handle special chars

phani-srikar:
#### Description of changes

This fixes the failures we observed with SQL tests when the randomly generated/RDS managed database password has special characters and is not encoded.

Ran the canary with changes and verified that tests now pass.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-graphql-api-construct-tests/src/sql-datatabase-controller.ts
+++ b/packages/amplify-graphql-api-construct-tests/src/sql-datatabase-controller.ts
@@ -260,6 +260,6 @@ export class SqlDatatabaseController {
    */
   getConnectionUri = (engine: SqlEngine, username: string, password: string, hostname: string, port: number, dbName: string): string => {
     const protocol = engine === 'postgres' ? 'postgresql' : 'mysql';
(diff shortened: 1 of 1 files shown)


packages/amplify-graphql-api-construct-tests/src/sql-datatabase-controller.ts
```diff
@@ -260,6 +260,6 @@ export class SqlDatatabaseController {
    */
   getConnectionUri = (engine: SqlEngine, username: string, password: string, hostname: string, port: number, dbName: string): string => {
     const protocol = engine === 'postgres' ? 'postgresql' : 'mysql';
-    return `${protocol}://${username}:${password}@${hostname}:${port}/${dbName}`;
+    return `${protocol}://${username}:${encodeURIComponent(password)}@${hostname}:${port}/${dbName}`;
```
palpatim, line 263: We should encode username as well, since technically mysql usernames can include spaces.

palpatim (approved):

AaronZyLee (approved):

PR #2796, opened 2024-08-21
chore: remove SQL related packages, update publish scripts

phani-srikar:
#### Description of changes

- Removes the schema generator and `@sql` transformer packages since they are only available in production via CDK construct or Gen2 workflows. These are already published from `main`.
- Add post publish step to attach `latest` tag to the published version with `api-plugin-stable` dist-tag for the `@aws-amplify/amplify-category-api` plugin to Gen-1 CLI.
- Some cleanup to configs.

CI checks. Had [full E2Es pass](https://tiny.amazon.com/flob31hg/IsenLink) on my earlier commit and commits after that update release scripts etc. The E2E run on latest commit is in progress with 2 failures that I'm running in 1st re-try.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


AaronZyLee (dismissed):
LGTM. Just one question on the yarn.lock change. Are the newly added packages such as aws-sdk and smithy a part of the schema generator dependencies?

codebuild_specs/scripts/post_publish.sh
```diff
@@ -0,0 +1,49 @@
+#!/bin/bash -e
+
+REGISTRY="https://registry.npmjs.org"
+PACKAGE_NAME="amplify-category-api"
+NEW_TAG="latest"
+RETRIES=30
+DELAY=60 # 1 minute delay between retries
```
phani-srikar, line 7: This waits for overall 30 mins for the released version to be available on NPM. With my testing with a tag release, it succeeded within 2 retires.

package.json
```diff
@@ -29,17 +29,16 @@
     "link-win": "node ./scripts/link-bin.js node_modules/amplify-cli-internal/bin/amplify amplify-dev",
     "lint": "eslint . --ext .js,.jsx,.ts,.tsx",
     "package": "lerna run package",
-    "postpublish:release": "git fetch . release:main && git push origin main",
-    "postpublish:api-plugin-stable": "git push origin api-plugin-stable",
+    "postpublish:release-api-plugin-stable": "./codebuild_specs/scripts/post_publish.sh",
```
phani-srikar, line 32: I've verified that lerna detects these post-publish scripts and runs them after their `publish` equivalents by adding a `postpublish:tag` temporarily. To be extra sure that it doesn't accidentally run `postpublish:release` and do destructive things to `main`, removed that script which is not needed.

.prettierignore
```diff
@@ -15,5 +15,4 @@ packages/amplify-e2e-tests/amplify-e2e-reports
 packages/amplify-graphql-model-transformer/publish-notification-lambda/lib/
 packages/amplify-graphql-model-transformer/rds-lambda/
 packages/amplify-graphql-model-transformer/rds-patching-lambda/lib/
```
palpatim, line 17: We wouldn't need any of these for Gen 1 right?

package.json
```diff
@@ -29,17 +29,16 @@
...
+    "publish-to-verdaccio": "lerna publish --dist-tag=api-plugin-stable --yes --no-commit-hooks --no-push --exact --conventional-commits --no-git-tag-version --no-verify-access",
     "publish:beta": "lerna publish --exact --dist-tag=beta --preid=beta --conventional-commits --conventional-prerelease --message 'chore(release): Publish [ci skip]' --no-verify-access --yes",
     "publish:main": "lerna publish --canary --force-publish --preid=alpha --exact --include-merged-tags --conventional-prerelease --no-verify-access --yes",
     "publish:release": "lerna publish --conventional-commits --conventional-graduate --exact --yes --message 'chore(release): Publish [ci skip]' --no-verify-access",
-    "publish:api-plugin-stable": "lerna publish --dist-tag=api-plugin-stable --conventional-commits --conventional-graduate --exact --yes --message 'chore(release): Publish [ci skip]' --no-verify-access",
+    "publish:release-api-plugin-stable": "lerna publish --dist-tag=api-plugin-stable --conventional-commits --conventional-graduate --exact --yes --message 'chore(release): Publish [ci skip]' --no-verify-access",
```
palpatim, line 41: Nit/suggestion: can we update the message to include the version number? (Note double quotes to preserve the NPM_TAG env var. I'm assuming that's the correct var to output the new release version.)

```suggestion
    "publish:release-api-plugin-stable": "lerna publish --dist-tag=api-plugin-stable --conventional-commits --conventional-graduate --exact --yes --message \"chore(release): Publish release $NPM_TAG\" --no-verify-access",
```

packages/amplify-category-api/package.json
```diff
@@ -30,7 +30,7 @@
   },
   "dependencies": {
     "@aws-amplify/graphql-auth-transformer": "3.6.5",
-    "@aws-amplify/graphql-schema-generator": "0.9.4",
+    "@aws-amplify/graphql-schema-generator": "^0.9.4",
```
palpatim, line 33: I assume we're retaining this for backwards compat with our Gen 2 E2E tests? If so, I further assume we'll remove this dependency here and throughout once we migrate all those tests to CDK rather than CLI?

scripts/split-e2e-tests.ts
```diff
@@ -148,8 +148,6 @@ const RUN_SOLO: (string | RegExp)[] = [
 const RUN_IN_ALL_REGIONS = [
   // DDB tests
   'src/__tests__/api_canary.test.ts',
-  // CDK tests
-  'src/__tests__/base-cdk.test.ts',
 ];
 
 const RUN_IN_NON_OPT_IN_REGIONS: (string | RegExp)[] = [
```
palpatim, line 153: Should we remove the rds & sql tests below?

package.json
phani-srikar: We do not have the version number available in env variables as lerna computes the new version to publish based on conventional commit history, as part of this command. We could do a `lerna version --no-push` before we publish to get the versions that will eventually be published, but that command updates local `package.json` files with new version and we would need to revert those before we run this publish command.

palpatim: Bummer, thanks for the explanation.

.prettierignore
phani-srikar: We don't need them in Gen-1, but prefer to keep them until we migrate the set of E2Es we run on `main` that use the Gen-1 plugin.

scripts/split-e2e-tests.ts
phani-srikar: Would prefer keeping them until we migrate our Gen-1 plugin based E2Es to use CDK construct on `main`.

packages/amplify-category-api/package.json
phani-srikar: Yes we will remove these dependencies and SQL E2Es once we migrate our tests to use CDK construct.

palpatim (approved):

dpilch (approved):

PR #2799, opened 2024-08-22
test: use pre provisioned cluster

fuelvin1:
#### Description of changes

Update the SQL E2E test setup to be able to use the pre-provisioned cluster. Use `sql-pg-canary` test to run in parent account and `us-east-1` region that has the pre-provisioned cluster.

Ran the updated test

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-e2e-core/src/utils/rds.ts
+++ b/packages/amplify-e2e-core/src/utils/rds.ts
@@ -278,14 +278,15 @@ export const setupDataInExistingCluster = async (
     }
 
     const clusterArn = dbClusterObj.DBClusterArn;
-    const masterSecretArn = dbClusterObj?.MasterUserSecret?.SecretArn || secretArn;
+    // use the provided database user secret or fallback to using master user secret
+    const dbUserSecretArn = secretArn || dbClusterObj?.MasterUserSecret?.SecretArn;
     const defaultDbName = dbClusterObj.DatabaseName;
     const dataClient = new RDSDataClient({ region: config.region });
     const sanitizedDbName = config.dbname?.replace(/[^a-zA-Z0-9_]/g, '');
 
     const createDBInput: ExecuteStatementCommandInput = {
       resourceArn: clusterArn,
-      secretArn: masterSecretArn,
+      secretArn: dbUserSecretArn,
       sql: `create database ${sanitizedDbName}`,
       database: defaultDbName,
     };
@@ -303,7 +304,7 @@ export const setupDataInExistingCluster = async (
       try {
         const executeStatementInput: ExecuteStatementCommandInput = {
           resourceArn: clusterArn,
-          secretArn: masterSecretArn,
+          secretArn: dbUserSecretArn,
           sql: query,
           database: sanitizedDbName,
         };
@@ -320,7 +321,7 @@ export const setupDataInExistingCluster = async (
       port: dbClusterObj.Port,
       dbName: sanitizedDbName,
       dbInstance: describeInstanceResponse.DBInstances[0],
-      secretArn: masterSecretArn,
+      secretArn: dbUserSecretArn,
       username: dbClusterObj.MasterUsername,
     };
   } catch (error) {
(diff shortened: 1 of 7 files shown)


packages/amplify-e2e-core/src/utils/rds.ts
```diff
@@ -273,14 +278,14 @@ export const setupDataInExistingCluster = async (identifier: string, config: RDS
     }
 
     const clusterArn = dbClusterObj.DBClusterArn;
-    const secretArn = dbClusterObj.MasterUserSecret.SecretArn;
+    const masterSecretArn = dbClusterObj?.MasterUserSecret?.SecretArn || secretArn;
```
palpatim, line 281: Typically you'd use `secretArn` if provided, and fall back to the master only if it wasn't present. As written, callers have no way to use users with a reduced scope if a Master User secret exists in the cluster config. 

Either way, you've no guarantee that the resolved value is a master secret, right? If you end up using the incoming argument, all you know is that it's a secret, not necessarily a DB master secret.

packages/amplify-graphql-api-construct-tests/src/utils/sql-pre-provisioned-cluster.ts
```diff
@@ -0,0 +1,35 @@
...
+  const s3Client = new S3Client({ region });
+  const stsClient = new STSClient({ region });
+
+  try {
+    const callerIdentity = await stsClient.send(new GetCallerIdentityCommand({}));
+    const accountId = callerIdentity?.Account;
+    if (!accountId) {
+      throw new Error('cannot get the current account Id');
+    }
+    // const clusterManifestPrefix = process.env.RDS_CLUSTER_MANIFEST_BUCKET_PREFIX;
```
palpatim, line 21: nit: Remove unless there's a compelling reason to keep it

packages/amplify-graphql-api-construct-tests/src/utils/sql-pre-provisioned-cluster.ts
```diff
@@ -0,0 +1,35 @@
...
+    const command = new GetObjectCommand({
+      Bucket: `${clusterManifestPrefix}${accountId}`,
+      Key: `${engine}/${region}`,
+    });
+    const response = await s3Client.send(command);
+    const clusterInfo: PreProvisionedClusterInfo = JSON.parse(await response.Body.transformToString());
+    console.log(JSON.stringify(clusterInfo));
+    return clusterInfo;
+  } catch (err) {
+    console.error(err);
```
palpatim, line 32: Can you confirm that returning an `undefined` value from this method will cause the test to create a new cluster? In other words, all the failure modes, from "no manifest for the specified engine/region combo" to missing/invalid account IDs, to errors connecting to S3, will result in a test-specific cluster being created and subsequently torn down?

packages/amplify-graphql-api-construct-tests/src/utils/sql-pre-provisioned-cluster.ts
```diff
@@ -0,0 +1,35 @@
...
+  const stsClient = new STSClient({ region });
+
+  try {
+    const callerIdentity = await stsClient.send(new GetCallerIdentityCommand({}));
+    const accountId = callerIdentity?.Account;
+    if (!accountId) {
+      throw new Error('cannot get the current account Id');
+    }
+    // const clusterManifestPrefix = process.env.RDS_CLUSTER_MANIFEST_BUCKET_PREFIX;
+    const clusterManifestPrefix = 'rdsmanifestbucket';
```
phani-srikar, line 22: nit: remove this and uncomment the above line.

packages/amplify-graphql-api-construct-tests/src/utils/sql-pre-provisioned-cluster.ts
fuelvin1: Yes, those cases should all cause the test to create a new cluster. I also just added the case where the cluster manifest is missing.

palpatim (dismissed):

phani-srikar (approved):

phani-srikar (changes requested):
Update yarn lock

palpatim (approved):

phani-srikar (approved):

PR #2812, opened 2024-08-27
allow specifying JS runtime when creating resolver functions within transformer

atierian:
### Allows specifying JS runtime for resolver functions
This change is only applicable to the transformer internals -- it does not expose a way for customers to specify the runtime of a resolver function.

Note: The changes made to accomplish this are the minimal changes necessary, not necessarily the best way. A more appropriate way of supporting JS runtime would be adding overloads or additional optional properties specifying `code` / `codeS3Location`. This is likely what we should do when adding support for handling `a.handler.custom({})` in the transformer. In the meantime, I favored minimal risk over ideal approach for this change.

* Conditionally added support for `Code` based AppSync Functions: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-appsync-functionconfiguration.html#cfn-appsync-functionconfiguration-code
* Conditionally added support for `Code` based AppSync Resolvers: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-appsync-resolver.html#cfn-appsync-resolver-code

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-graphql-transformer-core/API.md
+++ b/packages/amplify-graphql-transformer-core/API.md
@@ -13,6 +13,7 @@ import { AssetProvider } from '@aws-amplify/graphql-transformer-interfaces';
 import { AuthorizationConfig } from 'aws-cdk-lib/aws-appsync';
 import { AuthorizationType } from 'aws-cdk-lib/aws-appsync';
 import { CfnApiKey } from 'aws-cdk-lib/aws-appsync';
+import { CfnFunctionConfiguration } from 'aws-cdk-lib/aws-appsync';
 import { CfnGraphQLSchema } from 'aws-cdk-lib/aws-appsync';
 import { CfnParameter } from 'aws-cdk-lib';
 import { CfnResource } from 'aws-cdk-lib';
@@ -768,9 +769,9 @@ export abstract class TransformerPluginBase implements TransformerPluginProvider
 
 // @public (undocumented)
 export class TransformerResolver implements TransformerResolverProvider {
-    constructor(typeName: string, fieldName: string, resolverLogicalId: string, requestMappingTemplate: MappingTemplateProvider, responseMappingTemplate: MappingTemplateProvider, requestSlots: string[], responseSlots: string[], datasource?: DataSourceProvider | undefined);
+    constructor(typeName: string, fieldName: string, resolverLogicalId: string, requestMappingTemplate: MappingTemplateProvider, responseMappingTemplate: MappingTemplateProvider, requestSlots: string[], responseSlots: string[], datasource?: DataSourceProvider | undefined, runtime?: CfnFunctionConfiguration.AppSyncRuntimeProperty | undefined);
     // (undocumented)
-    addToSlot: (slotName: string, requestMappingTemplate?: MappingTemplateProvider, responseMappingTemplate?: MappingTemplateProvider, dataSource?: DataSourceProvider) => void;
+    addToSlot: (slotName: string, requestMappingTemplate?: MappingTemplateProvider, responseMappingTemplate?: MappingTemplateProvider, dataSource?: DataSourceProvider, runtime?: CfnFunctionConfiguration.AppSyncRuntimeProperty) => void;
     // Warning: (ae-forgotten-export) The symbol "Slot" needs to be exported by the entry point index.d.ts
     //
     // (undocumented)
--- a/packages/amplify-graphql-transformer-core/src/transformer-context/resolver.ts
+++ b/packages/amplify-graphql-transformer-core/src/transformer-context/resolver.ts
@@ -23,6 +23,7 @@ type Slot = {
   requestMappingTemplate?: MappingTemplateProvider;
   responseMappingTemplate?: MappingTemplateProvider;
   dataSource?: DataSourceProvider;
+  runtime?: CfnFunctionConfiguration.AppSyncRuntimeProperty;
 };
 
 // Name of the None Data source used for pipeline resolver
@@ -142,6 +143,7 @@ export class TransformerResolver implements TransformerResolverProvider {
     private requestSlots: string[],
     private responseSlots: string[],
     private datasource?: DataSourceProvider,
+    private runtime?: CfnFunctionConfiguration.AppSyncRuntimeProperty,
   ) {
     if (!typeName) {
       throw new InvalidDirectiveError('typeName is required');
@@ -179,6 +181,7 @@ export class TransformerResolver implements TransformerResolverProvider {
     requestMappingTemplate?: MappingTemplateProvider,
     responseMappingTemplate?: MappingTemplateProvider,
     dataSource?: DataSourceProvider,
+    runtime?: CfnFunctionConfiguration.AppSyncRuntimeProperty,
   ): void => {
     if (!this.slotNames.has(slotName)) {
       throw new Error(`Resolver is missing slot ${slotName}`);
@@ -197,6 +200,7 @@ export class TransformerResolver implements TransformerResolverProvider {
         requestMappingTemplate,
         responseMappingTemplate,
         dataSource,
+        runtime,
       });
     }
     this.slotMap.set(slotName, slotEntry);
@@ -272,7 +276,10 @@ export class TransformerResolver implements TransformerResolverProvider {
       this.responseMappingTemplate,
       this.datasource?.name || NONE_DATA_SOURCE_NAME,
       scope,
+      this.runtime,
     );
+
+    const stashStatement = this.stashStatementGenerator(this.runtime);
 
     let dataSourceType = 'NONE';
     let dataSource = '';
@@ -282,7 +289,7 @@ export class TransformerResolver implements TransformerResolverProvider {
         case 'AMAZON_DYNAMODB':
           if (this.datasource.ds.dynamoDbConfig && !isResolvableObject(this.datasource.ds.dynamoDbConfig)) {
             const tableName = this.datasource.ds.dynamoDbConfig?.tableName;
-            dataSource = `$util.qr($ctx.stash.put("tableName", "${tableName}"))`;
+            dataSource = stashStatement('tableName', `"${tableName}"`);
             if (
               this.datasource.ds.dynamoDbConfig?.deltaSyncConfig &&
               !isResolvableObject(this.datasource.ds.dynamoDbConfig?.deltaSyncConfig)
@@ -302,7 +309,7 @@ export class TransformerResolver implements TransformerResolverProvider {
                   return SyncUtils.syncDataSourceConfig().DeltaSyncTableTTL.toString();
                 },
               });
-              dataSource += `\n$util.qr($ctx.stash.put("deltaSyncTableTtl", ${deltaSyncTableTtl}))`;
+              dataSource += '\n' + stashStatement('deltaSyncTableTtl', `"${deltaSyncTableTtl}"`);
             }
           }
 
@@ -334,19 +341,19 @@ export class TransformerResolver implements TransformerResolverProvider {
         case 'AMAZON_ELASTICSEARCH':
           if (this.datasource.ds.elasticsearchConfig && !isResolvableObject(this.datasource.ds.elasticsearchConfig)) {
             const endpoint = this.datasource.ds.elasticsearchConfig?.endpoint;
-            dataSource = `$util.qr($ctx.stash.put("endpoint", "${endpoint}"))`;
+            dataSource = stashStatement('endpoint', `"${endpoint}"`);
           }
           break;
         case 'AWS_LAMBDA':
           if (this.datasource.ds.lambdaConfig && !isResolvableObject(this.datasource.ds.lambdaConfig)) {
             const lambdaFunctionArn = this.datasource.ds.lambdaConfig?.lambdaFunctionArn;
-            dataSource = `$util.qr($ctx.stash.put("lambdaFunctionArn", "${lambdaFunctionArn}"))`;
+            dataSource = stashStatement('lambdaFunctionArn', `"${lambdaFunctionArn}"`);
           }
           break;
         case 'HTTP':
           if (this.datasource.ds.httpConfig && !isResolvableObject(this.datasource.ds.httpConfig)) {
             const endpoint = this.datasource.ds.httpConfig?.endpoint;
-            dataSource = `$util.qr($ctx.stash.put("endpoint", "${endpoint}"))`;
+            dataSource = stashStatement('endpoint', `"${endpoint}"`);
           }
           break;
         case 'RELATIONAL_DATABASE':
@@ -356,7 +363,7 @@ export class TransformerResolver implements TransformerResolverProvider {
             !isResolvableObject(this.datasource.ds.relationalDatabaseConfig?.rdsHttpEndpointConfig)
           ) {
             const databaseName = this.datasource.ds.relationalDatabaseConfig?.rdsHttpEndpointConfig!.databaseName;
-            dataSource = `$util.qr($ctx.stash.metadata.put("databaseName", "${databaseName}"))`;
+            dataSource = stashStatement('databaseName', `"${databaseName}"`);
           }
           break;
         default:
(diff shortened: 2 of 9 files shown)


packages/amplify-graphql-transformer-core/src/transformer-context/resolver.ts
```diff
@@ -356,56 +363,85 @@ export class TransformerResolver implements TransformerResolverProvider {
...
     }
     const adminRoles = context.synthParameters.adminRoles ?? [];
+    const adminRolesStatement = stashStatement('adminRoles', `${JSON.stringify(adminRoles)}`);
     initResolver += dedent`\n
-      $util.qr($ctx.stash.put("adminRoles", ${JSON.stringify(adminRoles)}))
+      ${adminRolesStatement}
     `;
-    initResolver += '\n$util.toJson({})';
+
+    if (this.runtime?.name === 'APPSYNC_JS') {
```
palpatim, line 415: Here and throughout: can we do something other than a magic string comparisons for runtime evaluation? E.g., centralize a constant, add a type predicate, etc

packages/amplify-graphql-transformer-core/src/transformer-context/resolver.ts
```diff
@@ -466,4 +503,18 @@ export class TransformerResolver implements TransformerResolverProvider {
       });
     }
   }
+
+  private stashStatementGenerator(
+    runtime?: CfnFunctionConfiguration.AppSyncRuntimeProperty,
+  ): (name: string, value?: string, object?: string) => string {
+    const jsStashStatement = (name: string, value?: string, object?: string): string => {
```
palpatim, line 510: 1. Are we ever going to need the type signature of a `stash` statement elsewhere, suggesting that we should create a typealias and/or structured input for it?
2. naming nit: these are generator functions, not statements :)

packages/amplify-graphql-transformer-core/src/transformer-context/resolver.ts
```diff
@@ -282,7 +289,7 @@ export class TransformerResolver implements TransformerResolverProvider {
         case 'AMAZON_DYNAMODB':
           if (this.datasource.ds.dynamoDbConfig && !isResolvableObject(this.datasource.ds.dynamoDbConfig)) {
             const tableName = this.datasource.ds.dynamoDbConfig?.tableName;
-            dataSource = `$util.qr($ctx.stash.put("tableName", "${tableName}"))`;
+            dataSource = stashStatement('tableName', `"${tableName}"`);
```
palpatim, line 292: It seems brittle to make the caller responsible for remembering to include double quotes, and I'm worried about the error that will happen when we revisit this in a few months, have forgotten the details, and do a copypasta error leading to a hard-to-find bug. Can we adjust the stashStatement API to make it clear how to pass a value:

```typescript
stashStatement('tableName', { quote: tableName })
...
stashStatement('conditions', { literal: '[]' })
```

or

```ts
stashString('tableName', tableName)
...
stashExpression('tableName', '[]')
```

or some other clearer way of passing values?

(I know these kinds of sharp syntax edges are all over the codebase--believe me I know!--but I'd love for us to be the change we want to see in the code.)

packages/amplify-graphql-transformer-core/src/utils/function-runtime.ts
```diff
@@ -0,0 +1,45 @@
...
+    runtime?: CfnFunctionConfiguration.AppSyncRuntimeProperty;
+    api: GraphQLApi;
+  },
+): RuntimeSpecificFunctionProps => {
+  const { requestMappingTemplate, responseMappingTemplate, runtime, api } = props;
+
+  const requestTemplateLocation = requestMappingTemplate.bind(scope, api.assetProvider);
+  const responseTemplateLocation = responseMappingTemplate.bind(scope, api.assetProvider);
+
+  if (runtime?.name === 'APPSYNC_JS') {
```
palpatim, line 30: This gives me a bit of pause. I assume that your note about making minimal changes is why we've elected to special case the JS handling by assuming inline code, rather than allowing templates like the VTL template provider does? If so, I agree it's probably the right thing to do for this exercise.

packages/amplify-graphql-transformer-interfaces/src/graphql-api-provider.ts
```diff
@@ -128,6 +128,7 @@ export type MappingTemplateProvider = InlineMappingTemplateProvider | S3MappingT
 
 export interface GraphQLAPIProvider extends IConstruct {
   readonly apiId: string;
+  readonly graphqlUrl: string;
```
palpatim, line 131: I don't actually see this accessed or mutated anywhere in this PR. Am I missing its usage, or will it be consumed in a followup?

packages/amplify-graphql-transformer-core/src/transformer-context/resolver.ts
atierian: Good call, addressed in [113d921](https://github.com/aws-amplify/amplify-category-api/pull/2812/commits/113d921c23ec8d9262af4a05dca7705b881dcb82)

packages/amplify-graphql-transformer-core/src/transformer-context/resolver.ts
atierian: Great feedback, thanks. Extracting to types was borderline necessary already but definitely necessary after introducing `stashString` and `stashExpression`. Addressed in [113d921](https://github.com/aws-amplify/amplify-category-api/pull/2812/commits/113d921c23ec8d9262af4a05dca7705b881dcb82)

packages/amplify-graphql-transformer-core/src/transformer-context/resolver.ts
atierian: Yes, definitely a sharp edge. I figured it was no worse than before, but that's not following the boy scout rule very well. I like the `stashString` / `stashExpression` suggestion.

Addressed in [113d921](https://github.com/aws-amplify/amplify-category-api/pull/2812/commits/113d921c23ec8d9262af4a05dca7705b881dcb82).

packages/amplify-graphql-transformer-core/src/utils/function-runtime.ts
atierian: Your assumption is correct. I would've preferred to support `codeS3Location` as well here (i.e. do it right), but that got messy quickly because of VTL based assumptions throughout the codebase that templates are split into individual requests and responses -- something that is not the case with `code` (JS function resolvers).

Where both `responseMappingTemplateS3Location` and `requestMappingTemplateS3Location` are necessary for the VTL runtime, a single `codeS3Location` is required for JS runtime.

I'm hoping it's not as much work as it appeared, but I opted to ~kick the can down the road~ do this incrementally to reduce the risk of regressions in other transformers.

packages/amplify-graphql-transformer-interfaces/src/graphql-api-provider.ts
atierian: This change shouldn't have been included here. Removed it in [65451d6](https://github.com/aws-amplify/amplify-category-api/pull/2812/commits/65451d6fb963abc4c3232cbcc77b65b4d697015c) and will include it in a subsequent relevant PR. Thanks for calling that out.

packages/amplify-graphql-transformer-core/src/utils/function-runtime.ts
palpatim: > I opted to kick the can down the road ~~do this incrementally~~

Quality save. :)

packages/amplify-graphql-transformer-core/src/transformer-context/resolver.ts
```diff
@@ -504,17 +503,72 @@ export class TransformerResolver implements TransformerResolverProvider {
...
-    runtime?: CfnFunctionConfiguration.AppSyncRuntimeProperty,
-  ): (name: string, value?: string, object?: string) => string {
-    const jsStashStatement = (name: string, value?: string, object?: string): string => {
+  /**
+   * Generates a function to create stash statements based on the runtime.
+   *
+   * @param {CfnFunctionConfiguration.AppSyncRuntimeProperty} runtime - The AppSync runtime configuration.
+   * @returns {StashStatementGenerator} An object with methods to generate stash statements.
+   */
+  private generateStashStatementGenerator(runtime?: CfnFunctionConfiguration.AppSyncRuntimeProperty): StashStatementGenerator {
```
palpatim, line 512: Do any of our use cases accept double-quoted strings? If so, we'll need to escape the incoming values before passing them to the actual stasher (For JS, that'd be something like `value.replace(/"/g, '\\"')`; for VTL, it'd be something like `value.replace(/"/g, "${esc.q}")`, but we need to make sure that renders correctly in the resolver!

packages/amplify-graphql-transformer-core/src/transformer-context/resolver.ts
```diff
@@ -504,17 +503,72 @@ export class TransformerResolver implements TransformerResolverProvider {
...
+
+type StashStatementGeneratorFunction = (props: StashStatementGeneratorProps) => string;
+
+/**
+ *  Stash statement generator methods.
+ */
+type StashStatementGenerator = {
+  /**
+   * Generates a stash statement for string values.
+   * This method ensures that the value is properly quoted as a string.
```
palpatim, line 559: per comment above:
```suggestion
   * This method ensures that the value is properly quoted as a string. Double-quotes in the incoming string will be escaped according to the rules of the runtime.
```

packages/amplify-graphql-transformer-core/src/transformer-context/resolver.ts
atierian: Nothing within the `TransformerResolver` has double quoted strings. As you mentioned, we may have needs for that capability elsewhere. That's one of the reasons I opted to keep `generateStashStatementGenerator` private to the class, so as not to give the impression that this is an all-powerful / generic utility that can be used anywhere.

Do you see any concerns for the scope of stash statements generated within `TransformerResolver`? If not, I'd prefer to keep this as is for now and tackle generic handling as needed / separately.

palpatim (dismissed):

packages/amplify-graphql-transformer-core/API.md
```diff
@@ -768,9 +769,9 @@ export abstract class TransformerPluginBase implements TransformerPluginProvider
 
 // @public (undocumented)
 export class TransformerResolver implements TransformerResolverProvider {
-    constructor(typeName: string, fieldName: string, resolverLogicalId: string, requestMappingTemplate: MappingTemplateProvider, responseMappingTemplate: MappingTemplateProvider, requestSlots: string[], responseSlots: string[], datasource?: DataSourceProvider | undefined);
+    constructor(typeName: string, fieldName: string, resolverLogicalId: string, requestMappingTemplate: MappingTemplateProvider, responseMappingTemplate: MappingTemplateProvider, requestSlots: string[], responseSlots: string[], datasource?: DataSourceProvider | undefined, runtime?: CfnFunctionConfiguration.AppSyncRuntimeProperty | undefined);
```
phani-srikar, line 772: since the `runtime` prop itself is optional, can it just be `CfnFunctionConfiguration.AppSyncRuntimeProperty`?

packages/amplify-graphql-transformer-core/src/transformer-context/resolver.ts
```diff
@@ -272,8 +277,11 @@ export class TransformerResolver implements TransformerResolverProvider {
       this.responseMappingTemplate,
       this.datasource?.name || NONE_DATA_SOURCE_NAME,
       scope,
+      this.runtime,
     );
 
+    const { stashString, stashExpression } = this.generateStashStatementGenerator(this.runtime);
```
phani-srikar, line 283: nit: having two `generate`s in same name is confusing. Would rename this to `createStashStatementGenerator`.

packages/amplify-graphql-transformer-core/src/transformer-context/resolver.ts
```diff
@@ -356,56 +364,83 @@ export class TransformerResolver implements TransformerResolverProvider {
...
+      ${stashString({ name: 'dataSourceType', value: dataSourceType, object: 'metadata' })}
+      ${stashString({ name: 'apiId', value: api.apiId, object: 'metadata' })}
+      ${stashExpression({ name: 'connectionAttributes', value: '{}' })}
       ${dataSource}
     `;
     const account = Stack.of(context.stackManager.scope).account;
     const authRole = context.synthParameters.authenticatedUserRoleName;
     if (authRole) {
+      const authRoleArn = `arn:aws:sts::${account}:assumed-role/${authRole}/CognitoIdentityCredentials`;
+      const authRoleStatement = stashString({ name: 'authRole', value: authRoleArn });
```
phani-srikar, line 388: Thanks for introducing this stashing abstraction!

packages/amplify-graphql-transformer-core/src/transformer-context/resolver.ts
```diff
@@ -356,56 +364,83 @@ export class TransformerResolver implements TransformerResolverProvider {
...
+        export const request = (ctx) => {
+          ${initResolver}
+          return {};
+        }
+      `;
+    } else {
+      initResolver += '\n$util.toJson({})';
+    }
+
+    const initResponseResolver = isJsResolverFnRuntime(this.runtime)
```
phani-srikar, line 426: I'm not a fan of introducing more such conditionals in existing packages based on runtime. Something to think about long term is to have runtime specific generators defined. That would consolidate all the runtime specific generations/overrides.

packages/amplify-graphql-transformer-core/src/utils/function-runtime.ts
```diff
@@ -0,0 +1,72 @@
...
+  const { requestMappingTemplate, responseMappingTemplate, runtime, api } = props;
+
+  const requestTemplateLocation = requestMappingTemplate.bind(scope, api.assetProvider);
+  const responseTemplateLocation = responseMappingTemplate.bind(scope, api.assetProvider);
+
+  if (isJsResolverFnRuntime(runtime)) {
+    return {
+      runtime,
+      code: requestTemplateLocation + '\n\n' + responseTemplateLocation,
+    };
```
phani-srikar, line 61: can you justify how the `request/responseMappingTemplateS3Location` which are set below for VTL runtime are not relevant to JS runtime? Do we only support inline code for JS, if so why?

packages/amplify-graphql-transformer-core/API.md
atierian: I would think so, but this is autogenerated via `yarn extract-api`; I have no control over it.
Here's the [constructor definition](https://github.com/aws-amplify/amplify-category-api/pull/2812/files#diff-b3129bb53f056565ba11ca27f5846ebb888f0f8bdd643238e2621cd44cd365bdR147) it's based on. You can see the same behavior for the previous property `datasource?`

packages/amplify-graphql-transformer-core/src/transformer-context/resolver.ts
atierian: Totally agree. The alternative meant a much larger (and imo riskier) code change that should be done, but done in isolation.

packages/amplify-graphql-transformer-core/src/utils/function-runtime.ts
atierian: Tim had a similar question. I think my response is relevant here as well.
https://github.com/aws-amplify/amplify-category-api/pull/2812#discussion_r1735295301

Let me know if I can clarify anything better.

packages/amplify-graphql-transformer-core/src/transformer-context/resolver.ts
atierian: Good call. Addressed in [35481e8](https://github.com/aws-amplify/amplify-category-api/pull/2812/commits/35481e8689c9ba316c20adf2b32461d17486193f)

palpatim (approved):

phani-srikar (approved):

PR #2832, opened 2024-09-03
chore: add amplify backend validation in CI

bobbyu99:
#### Description of changes

Added amplify backend validation

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes


shared-scripts.sh
```diff
@@ -132,6 +132,72 @@ function _lint {
...
+
+  # Unset container credentials environment variables since some of the tests in packages/cli/src/command_middleware.test.ts 
+  # expect not to fetch the credentials. This is to avoid the tests from failing.
+  echo "Unsetting container credentials environment variables"
+  unset AWS_CONTAINER_CREDENTIALS_RELATIVE_URI
+  unset AWS_CONTAINER_CREDENTIALS_FULL_URI
+
+  # 1. Install NVM and set up
+  echo "Installing NVM and setting Node.js version"
+  curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/master/install.sh | bash
```
palpatim, line 147: nit: the `-o-` formulation is not a standard construction. Please use `-o -` to make it clear that you're passing separate flags.
```suggestion
  curl -o - https://raw.githubusercontent.com/nvm-sh/nvm/master/install.sh | bash
```

shared-scripts.sh
```diff
@@ -132,6 +132,72 @@ function _lint {
...
+  echo "Verify Amplify Backend Compatibility"
+  loadCacheFromBuildJob
+
+  # Unset container credentials environment variables since some of the tests in packages/cli/src/command_middleware.test.ts 
+  # expect not to fetch the credentials. This is to avoid the tests from failing.
+  echo "Unsetting container credentials environment variables"
+  unset AWS_CONTAINER_CREDENTIALS_RELATIVE_URI
+  unset AWS_CONTAINER_CREDENTIALS_FULL_URI
+
+  # 1. Install NVM and set up
```
palpatim, line 145: Since we're explicitly overriding the container's node version, add a quick note in this comment about why.

shared-scripts.sh
```diff
@@ -132,6 +132,72 @@ function _lint {
...
+  echo "Installing NVM and setting Node.js version"
+  curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/master/install.sh | bash
+  export NVM_DIR="$HOME/.nvm"
+  [ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"
+  nvm install 18.20.4
+  nvm use 18.20.4
+  # Echo to verify the Node.js version
```
palpatim, line 152: The comment doesn't really add any explanatory value here :) 
```suggestion
```

shared-scripts.sh
```diff
@@ -132,6 +132,72 @@ function _lint {
...
+  echo "Node.js version in use:"
+  node -v
+  # Increase buffer size to avoid error when git operations return large response on CI
```
palpatim, line 155: This is a good example of a comment: it explains why we're doing this thing that's not obvious from just reading the code itself. It could be improved by explaining why that particular magic number is the right one. That will help us understand why that value is in use, whether it's safe to modify later, etc.

shared-scripts.sh
```diff
@@ -132,6 +132,72 @@ function _lint {
...
+  echo "Emulating Validate Shell"
+  # Move back to root and clone amplify-backend repo
+  cd ..
+  git clone https://github.com/aws-amplify/amplify-backend.git
+  cd amplify-backend
+  # Create a new local branch for testing 
+  git checkout -B validate-amplify-backend
+  # Update the packages and ensure the correct versions are being used
+  npm update
+  # Verify that the package-lock.json contains the updated version with localhost tarballs
```
palpatim, line 184: You're asserting that it contains the packages, but not that they're local tarballs -- the package.json could still be pulling in actual NPM published versions and still pass the grep. Should you be looking for `graphql-api-construct.*tgz` and `data-construct.*tgz` or similar?

shared-scripts.sh
```diff
@@ -132,6 +132,72 @@ function _lint {
...
+  source ./shared-scripts.sh && _publishLocalWorkspace
+  setNpmRegistryUrlToLocal
+  # Verify that the NPM registry has been set to the local Verdaccio server
+  npm config get registry
+
+  # 3. Validate Shell (Emulating the "validate" shell)
+  echo "Emulating Validate Shell"
+  # Move back to root and clone amplify-backend repo
+  cd ..
+  git clone https://github.com/aws-amplify/amplify-backend.git
```
palpatim, line 178: Consider limiting clone depth since you don't need the whole commit history. 

```suggestion
  git clone --depth 1 https://github.com/aws-amplify/amplify-backend.git
```

palpatim (approved):

phani-srikar (approved):

PR #2839, opened 2024-09-04
chore: use current branch for triggering release flow

phani-srikar:
#### Description of changes

Update triggering codebuild projects from current branch instead of using fixed branch names.
Add `api-plugin-stable` as known reserved tag so we prevent developers from using it during tag release workflows.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/scripts/cloud-release.sh
+++ b/scripts/cloud-release.sh
@@ -7,7 +7,12 @@ export DEPRECATE_PROJECT_NAME=amplify-category-api-deprecate-workflow
 
 function triggerRelease {
   branch_name=$(git branch --show-current)
-  triggerProjectBatch $RELEASE_ACCOUNT_PROD $RELEASE_ROLE_NAME "${RELEASE_PROFILE_NAME}Prod" $RELEASE_PROJECT_NAME $branch_name
+  if isReleaseBranch "$branch_name"; then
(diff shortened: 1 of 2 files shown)


scripts/cloud-release.sh
```diff
@@ -6,7 +6,8 @@ export RELEASE_PROJECT_NAME=amplify-category-api-release-workflow
 export DEPRECATE_PROJECT_NAME=amplify-category-api-deprecate-workflow
 
 function triggerRelease {
-  triggerProjectBatch $RELEASE_ACCOUNT_PROD $RELEASE_ROLE_NAME "${RELEASE_PROFILE_NAME}Prod" $RELEASE_PROJECT_NAME "release"
+  branch_name=$(git branch --show-current)
```
palpatim, line 9: Since this will only be used for prod releases for Gen2 or LTS, can we add some validation here that it's one of our expected branch names. That will prevent us from accidentally triggering a release on a PR branch, for example.

phani-srikar: Added a check to limit this to release branches. The reason not to be too strict here is to save us from making changes in multiple places (or forgetting to) when we add more release branches, since we already have publish scripts for only specific branches defined in our outer `package.json`.

AaronZyLee (approved):

PR #2864, opened 2024-09-11
feat(function-transformer): add support for event invocation types (async lambda)

atierian:
## Changes already reviewed in 
- https://github.com/aws-amplify/amplify-category-api/pull/2666

---

## Description of changes
Adds support for [AppSync asynchronous Lambda function invocations](https://aws.amazon.com/about-aws/whats-new/2024/05/aws-appsync-events-asynchronous-lambda-function-invocations/) via the `@function` directive.

### API Changes
```diff
directive @function(
  name: String!,
  region: String,
  accountId: String,
+  invocationType: InvocationType = RequestResponse
)

+ enum InvocationType { 
+   RequestResponse
+   Event
+ }
```

#### `appendError` vs `error`
The response mapping template for `Event` invocation types uses `util.appendError` rather than `util.error` used by `RequestResponse` (default) invocation types.

`util.error` has always been the behavior of the `@function` directive. I was unable to find any justification for the decision in the original PR:
- https://github.com/aws-amplify/amplify-cli/pull/1346

`util.error` immediately exits preventing any further execution in the response (also request but n/a here) mapping template **and** subsequent resolvers in the same pipeline. This is reasonable in the context of single function resolvers, but is unnecessarily constraining in a multi-function pipeline -- using `util.appendError` could¹ allow the subsequent Lambda function to make a handling decision for the previous error based on business logic.

Changing this behavior for existing `RequestResponse` (default) `@function` directives would be a breaking change. However, we have an opportunity to move in a more flexible direction for the newly introduced `Event` based `@function` directives. We're using `util.appendError` in the response mapping template of `Event` `@function` directives for two reasons:
1. **It allows us to return the `EventInvocationResponse` regardless of whether the Lambda function execution failed or not.**
```json
{ 
  "data": { 
    "myMutation": { 
      "success": true
    }
  }
}
```
```json
{ 
  "data": { 
    "myMutation": { 
      "success": false
    }
  },
  "errors": [
    {  }
  ]
}
```
For comparison, if `Event` based `@function` directives used `util.error`, this would be the failure response:
```json
{ 
  "data": { 
    "myMutation": null
  },
  "errors": [
    {  }
  ]
}
```

2. **It allows execution of subsequent resolvers in the pipeline.**

¹ This would require passing `ctx.error` in the Lambda invocation payload; this is **not** done today and is not included in this PR.

While we're on the topic, I'm going to document the behavior of `util.error` and `util.appendError` in pipeline resolvers for future reference.

- `util.error` immediately exits, returning the error and only the error to the caller. Neither subsequent lines in the request/response mapping template nor subsequent resolvers in a pipeline are executed after `util.error`
- `util.appendError` continues execution and returns the appended error in the `errors` block of the request.
- Appending an error using `util.appendError`, then calling `util.error` results in the appended error being **excluded** in the `errors` block -- only the error passed to `util.error` is returned in the `errors` block.
 
### Response Types
Using `@function(name: 'name', invocationType: Event)` requires that the response type of that field is `EventInvocationResponse`, which must be explicitly defined in the schema as:
```gql
type EventInvocationResponse {
  success: Boolean!
}
```
This applies when the `Event` invocation type `@function` directive is the final `@function` directive defined on that field. For example:
```gql
type Mutation {
  asyncStuff(msg: String): EventInvocationResponse @function(name: "stuff", invocationType: Event)
}
```
```gql
type Mutation {
  asyncStuff(msg: String): EventInvocationResponse  
  @function(name: "stuff")
  @function(name: "other-stuff", invocationType: Event)
}
```
But **not** when a 'RequestResponse' invocation is the final `@function` directive. This is fine:
```gql
type Mutation {
  asyncStuff(msg: String): String
  @function(name: "stuff", invocationType: Event)
  @function(name: "other-stuff")
}
```

### Constraints
`Event` invocation type `@function` directives can only be used on `Mutation` or `Query` fields. This is is more constraining than `RequestResponse` (default) invocation types.

- Snapshot testing.
- Manually deployed and testing GraphQL API.

Note: This PR is into the `feature/async-lambda` branch where a tagged release will be cut. A full E2E test run will happen prior to merging this to `main`.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-graphql-directives/src/directives/function.ts
```diff
@@ -1,12 +1,22 @@
 import { Directive } from './directive';
 
 const name = 'function';
+// export type FunctionInvocationType = 'RequestResponse' | 'Event'
```
palpatim, line 4: nit: remove or reinstate and make `FunctionDirectiveDefaults.invocationType` a FunctionInvocationType rather than a string. (Future PR is fine)

packages/amplify-graphql-function-transformer/src/__tests__/amplify-graphql-function-transformer.test.ts
```diff
@@ -369,3 +371,318 @@ test('includes auth info in stash', () => {
...
+          asyncStuff(msg: String): EventInvocationResponse
+          @function(name: "asyncstuff2-\${env}")
+          @function(name: "asyncstuff-\${env}", invocationType: Event)
+        }
+        type EventInvocationResponse {
+          success: Boolean!
+        }
+      `;
+      const out = testTransform({ schema, transformers: [new FunctionTransformer()] });
+      expect(out).toBeDefined();
```
palpatim, line 456: Here and other success tests: Do we need to assert the execution order & return value of the resolver? Or like, a snapshot? :)

packages/amplify-graphql-function-transformer/src/graphql-function-transformer.ts
```diff
@@ -178,6 +193,51 @@ export class FunctionTransformer extends TransformerPluginBase {
   };
 }
 
+/**
+ * The response mapping template for 'RequestResponse' (default) and 'Event' invocation types differ.
+ * Use this to generate the appropriate response mapping template.
+ * @param config the {@link FunctionDirectiveConfiguration} for `@function` definition.
+ * @returns the response mapping template used by AppSync to handle responses from the Lambda function invocation.
+ */
+const responseMappingTemplate = (config: FunctionDirectiveConfiguration): string => {
```
palpatim, line 202: nit: split each template generator into a separate method. (future PR is ok)

packages/amplify-graphql-function-transformer/src/graphql-function-transformer.ts
```diff
@@ -178,6 +193,51 @@ export class FunctionTransformer extends TransformerPluginBase {
...
+      #set( $success = true )
+      #if( $ctx.error )
+        $util.appendError($ctx.error.message, $ctx.error.type)
+        #set( $success = false )
+      #end
+      #set( $response = {
+        "success": $success
+      } )
+      $util.toJson($response)
+    */
```
palpatim, line 214: nit: remove (future PR is OK)

packages/amplify-graphql-function-transformer/src/graphql-function-transformer.ts
```diff
@@ -178,6 +193,51 @@ export class FunctionTransformer extends TransformerPluginBase {
...
+      ]),
+    );
+  }
+
+  /*
+    #if( $ctx.error )
+      $util.error($ctx.error.message, $ctx.error.type)
+    #end
+    $util.toJson($ctx.result)
+  */
```
palpatim, line 232: nit: remove

packages/amplify-graphql-function-transformer/src/graphql-function-transformer.ts
```diff
@@ -178,6 +193,51 @@ export class FunctionTransformer extends TransformerPluginBase {
...
+  return printBlock('Handle error or return result')(
```
palpatim, line 233: Should we add detail like "for a synchronous function invocation"? Might make it easier to read through the generated VTL.

packages/amplify-graphql-function-transformer/src/graphql-function-transformer.ts
```diff
@@ -178,6 +193,51 @@ export class FunctionTransformer extends TransformerPluginBase {
...
+      #if( $ctx.error )
+        $util.appendError($ctx.error.message, $ctx.error.type)
+        #set( $success = false )
+      #end
+      #set( $response = {
+        "success": $success
+      } )
+      $util.toJson($response)
+    */
+    return printBlock('Handle error or return result')(
```
palpatim, line 215: Should we add detail like "for an asynchronous function invocation"? Might make it easier to read through the generated VTL.

packages/amplify-graphql-function-transformer/src/graphql-function-transformer.ts
```diff
@@ -195,3 +255,256 @@ const lambdaArnKey = (name: string, region?: string, accountId?: string): string
...
+
+/**
+ * Validates that a `@function` directive is used in a supported way.
+ * @param config the {@link FunctionDirectiveConfiguration} to validate.
+ * @param definition the {@link FieldDefinitionNode} on which the `@function` directive is used.
+ * @param ctx the {@link TransformerContextProvider}
+ */
+const validate = (config: FunctionDirectiveConfiguration, definition: FieldDefinitionNode, ctx: TransformerContextProvider): void => {
+  if (config.invocationType === 'Event') {
+    // For any occurance  of 'invocationType: Event' validate:
```
palpatim, line 269: nit (future PR is OK)
```suggestion
    // For any occurence  of 'invocationType: Event' validate:
```

packages/amplify-graphql-function-transformer/src/graphql-function-transformer.ts
```diff
@@ -195,3 +255,256 @@ const lambdaArnKey = (name: string, region?: string, accountId?: string): string
...
+    // This really shouldn't happen because this code path should not execute without there being at least one `@function` directive
+    // defined on this field. But we're accessing the last element below, so better safe than sorry.
+    throw new InvalidDirectiveError(
+      'Unable to locate @function directives for field.' +
+        'This is an unexpected error that suggests there is a bug in the amplify-graphql-function-transformer.\n' +
+        'Please open a GitHub issue referencing this error message with your schema definition.',
+    );
+  }
+
+  const ultimateFunctionDirective = functionDirectivesForField[functionDirectivesForField.length - 1];
```
palpatim, line 324: I mentally read `ultimateFunctionDirective` in the "SUNDAY! SUNDAY! SUNDAY!" voice.

packages/amplify-graphql-function-transformer/src/graphql-function-transformer.ts
```diff
@@ -195,3 +255,256 @@ const lambdaArnKey = (name: string, region?: string, accountId?: string): string
...
+};
+
+/**
+ * Validates that the parent type is valid for 'invocationType: Event'
+ * - parent type === 'Query' | 'Mutation'
+ * This is done to ensure we don't accidentally add Event invocation type support to other types, e.g. `@model` or subscriptions.
+ * @param config the {@link FunctionDirectiveConfiguration} for 'invocationType: Event' definitions.
+ */
+const validateIsSupportedEventInvocationParentType = (config: FunctionDirectiveConfiguration): void => {
+  const isValidParentType = config.resolverTypeName === 'Query' || config.resolverTypeName === 'Mutation';
```
palpatim, line 438: nit: graphql-utils.ts has `isMutationType` and `isQueryType` utils rather than propagating string comparisons. Disclaimer: the util uses string comparisons.

packages/amplify-graphql-function-transformer/src/graphql-function-transformer.ts
phani-srikar: +1 to splitting this up into dedicated methods based on invocation type.

packages/amplify-graphql-function-transformer/src/graphql-function-transformer.ts
```diff
@@ -195,3 +255,256 @@ const lambdaArnKey = (name: string, region?: string, accountId?: string): string
...
+
+  if (!isValidParentType) {
+    throw new InvalidDirectiveError("@function definition with 'invocationType: Event' must be defined on Query or Mutation field.");
+  }
+};
+
+// #endregion Validation helpers
+
+// #region Debug Description Helpers
```
phani-srikar, line 447: nit: what value are these comments adding in addition to the method specific docs? Is this your IDE setting or something?

phani-srikar (dismissed):

packages/amplify-graphql-function-transformer/src/graphql-function-transformer.ts
atierian: yep! https://code.visualstudio.com/docs/editor/codebasics#_folding

packages/amplify-graphql-directives/src/directives/function.ts
atierian: I can't believe that slipped through. Thanks -- addressed in https://github.com/aws-amplify/amplify-category-api/pull/2868

packages/amplify-graphql-function-transformer/src/__tests__/amplify-graphql-function-transformer.test.ts
atierian: Great point. Addressed in https://github.com/aws-amplify/amplify-category-api/pull/2868

packages/amplify-graphql-function-transformer/src/graphql-function-transformer.ts
atierian: Done. Addressed in https://github.com/aws-amplify/amplify-category-api/pull/2868

packages/amplify-graphql-function-transformer/src/graphql-function-transformer.ts
atierian: I still stand by my [original comment here](https://github.com/aws-amplify/amplify-category-api/pull/2666#discussion_r1661158663), but definitely don't feel strongly enough to push for keeping it. Addressed in https://github.com/aws-amplify/amplify-category-api/pull/2868

packages/amplify-graphql-function-transformer/src/graphql-function-transformer.ts
atierian: Addressed in https://github.com/aws-amplify/amplify-category-api/pull/2868

packages/amplify-graphql-function-transformer/src/graphql-function-transformer.ts
atierian: Good call. Addressed in https://github.com/aws-amplify/amplify-category-api/pull/2868

packages/amplify-graphql-function-transformer/src/graphql-function-transformer.ts
atierian: Same as above. Addressed in https://github.com/aws-amplify/amplify-category-api/pull/2868

packages/amplify-graphql-function-transformer/src/graphql-function-transformer.ts
atierian: Done. Addressed in https://github.com/aws-amplify/amplify-category-api/pull/2868

packages/amplify-graphql-function-transformer/src/graphql-function-transformer.ts
atierian: These utilities actually don't work here. 🤷 
"don't work" meaning:
```ts
config.resolverTypeName = 'Query';
isQueryType(config.resolverTypeName) // false
```

sundersc (approved):

PR #2867, opened 2024-09-11
fix: un-require required input fields with @default

p5quared:
#### Description of changes
Un-require required input fields that have `@default` applied. Also update the corresponding snapshot to document this behavior.

Prior to this change, a copy of the input object was being modified but `ctx.output` was not being updated with the new input.

n/a

#### Issue #, if available
https://github.com/aws-amplify/amplify-codegen/issues/390

Fixed a test snapshot and passed it.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-graphql-default-value-transformer/src/graphql-default-value-transformer.ts
```diff
@@ -72,10 +72,21 @@ const validateDefaultValueType = (ctx: TransformerSchemaVisitStepContextProvider
   }
 };
 
+const validateNotPrimaryKey = (field: FieldDefinitionNode): void => {
+  const isPrimaryKeyField =
+    field.directives!.find((dir) => dir.name.value === 'primaryKey') ||
+    (getBaseType(field.type) === 'ID' && field.type.kind === Kind.NON_NULL_TYPE && field.name.value === 'id');
+
```
phani-srikar, line 79: There is one more case to consider here which is when the `@default` is applied to fields part of a composite primary key. [Gen2 doc](https://docs.amplify.aws/react/build-a-backend/data/data-modeling/identifiers/#composite-identifier) or this [Gen1 doc is more explicit](https://docs.amplify.aws/gen1/react/build-a-backend/graphqlapi/data-modeling/#configure-a-primary-key) about the schema with composite primary key.

p5quared: Added a separate check scanning over `sortKeyFields`.

packages/amplify-graphql-default-value-transformer/src/graphql-default-value-transformer.ts
```diff
@@ -72,10 +73,34 @@ const validateDefaultValueType = (ctx: TransformerSchemaVisitStepContextProvider
...
+  if (isPrimaryKeyField) {
+    throw new InvalidDirectiveError('The @default directive may not be applied to primaryKey fields.');
+  }
+};
+
+const validateNotCompositeKeyMember = (config: DefaultValueDirectiveConfiguration): void => {
+  const objectDirectives = config.object.fields?.flatMap((f) => f.directives);
+  const primaryKeyDirective = objectDirectives?.find((dir) => dir?.name.value === 'primaryKey');
```
phani-srikar, line 88: nit: can we refactor this check to see if the field is a primary key into it's own method that can be re-used between this method and the one above?

PR #2870, opened 2024-09-12
added enums via ref for single models

tejas2008:
#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/packages/amplify-graphql-schema-generator/src/ts-schema-generator/generate-ts-schema.ts
+++ b/packages/amplify-graphql-schema-generator/src/ts-schema-generator/generate-ts-schema.ts
@@ -1,6 +1,7 @@
 import * as ts from 'typescript';
 import { Schema } from '../schema-representation';
 import { createImportExpression, createSchema, DataSourceGenerateConfig } from './helpers';
+import { LogFormat } from '@aws-sdk/client-lambda';
 
 const file = ts.createSourceFile('schema.ts', '', ts.ScriptTarget.ESNext, false, ts.ScriptKind.TS);
 const printer = ts.createPrinter({ newLine: ts.NewLineKind.LineFeed });
@@ -18,3 +19,7 @@ export const generateTypescriptDataSchema = (schema: Schema, config?: DataSource
   );
   return result;
 };
+
+
+
+
--- a/packages/amplify-graphql-schema-generator/src/ts-schema-generator/helpers.ts
+++ b/packages/amplify-graphql-schema-generator/src/ts-schema-generator/helpers.ts
@@ -1,7 +1,7 @@
 import ts from 'typescript';
 import { TYPESCRIPT_DATA_SCHEMA_CONSTANTS } from 'graphql-transformer-common';
 import { VpcConfig } from '@aws-amplify/graphql-transformer-interfaces';
-import { DBEngineType, Field, FieldType, Model, Schema } from '../schema-representation';
+import { DBEngineType, EnumType, Field, FieldType, Model, Schema } from '../schema-representation';
 
 const GQL_TYPESCRIPT_DATA_SCHEMA_TYPE_MAP = {
   string: 'string',
@@ -38,6 +38,7 @@ const createProperty = (field: Field): ts.Node => {
  * @returns Typescript data schema type in TS Node format
  */
 const createDataType = (type: FieldType): ts.Node => {
+
   if (type.kind === 'Scalar') {
     return ts.factory.createCallExpression(
       ts.factory.createIdentifier(`${TYPESCRIPT_DATA_SCHEMA_CONSTANTS.REFERENCE_A}.${getTypescriptDataSchemaType(type.name)}`),
@@ -47,21 +48,19 @@ const createDataType = (type: FieldType): ts.Node => {
   }
 
   if (type.kind === 'Enum') {
+
     return ts.factory.createCallExpression(
-      ts.factory.createIdentifier(`${TYPESCRIPT_DATA_SCHEMA_CONSTANTS.REFERENCE_A}.${TYPESCRIPT_DATA_SCHEMA_CONSTANTS.ENUM_METHOD}`),
+      ts.factory.createIdentifier(`${TYPESCRIPT_DATA_SCHEMA_CONSTANTS.REFERENCE_A}.${TYPESCRIPT_DATA_SCHEMA_CONSTANTS.REF_METHOD}`),
       undefined,
-      [
-        ts.factory.createArrayLiteralExpression(
-          type.values.map((value) => ts.factory.createStringLiteral(value)),
-          true,
-        ),
-      ],
+      [ts.factory.createStringLiteral(type.name)]
+      
     );
   }
 
   // We do not import any Database type as 'Custom' type.
   // In case if there is a custom type in the IR schema, we will import it as string.
   if (type.kind === 'Custom') {
+    console.log(type);
     return ts.factory.createCallExpression(
       ts.factory.createIdentifier(`${TYPESCRIPT_DATA_SCHEMA_CONSTANTS.REFERENCE_A}.${TYPESCRIPT_DATA_SCHEMA_CONSTANTS.STRING_METHOD}`),
       undefined,
@@ -90,6 +89,18 @@ const createModelDefinition = (model: Model): ts.Node => {
   });
   return ts.factory.createObjectLiteralExpression(properties as ts.ObjectLiteralElementLike[], true);
 };
+
+const createEnums = (type: EnumType): ts.Node => {
+  const typeExpression =  ts.factory.createCallExpression(ts.factory.createIdentifier(`${TYPESCRIPT_DATA_SCHEMA_CONSTANTS.REFERENCE_A}.${TYPESCRIPT_DATA_SCHEMA_CONSTANTS.ENUM_METHOD}`), undefined, 
+  [
+    ts.factory.createArrayLiteralExpression(
+      type.values.map((value) => ts.factory.createStringLiteral(value)),
+      true,
+    ),
+  ],
+  );
+  return ts.factory.createPropertyAssignment(ts.factory.createIdentifier(type.name), typeExpression);
+}
 
 const createModel = (model: Model): ts.Node => {
   const modelExpr = ts.factory.createCallExpression(
@@ -147,19 +158,44 @@ export const createSchema = (schema: Schema, config?: DataSourceGenerateConfig):
     throw new Error('No valid tables found. Make sure at least one table has a primary key.');
   }
 
+  // find enums in all models
+
+  const enums = schema
+    .getModels()
+    .map((model) => model
+        .getFields()
+        .map((field) => {
+            if(field.type.kind === 'Enum'){
+                return createEnums(field.type);
+            }
+            else if(field.type.kind === 'NonNull' && field.type.type.kind === 'Enum'){
+                return createEnums(field.type.type);
+            } 
+            else {
+              return undefined;
+            }
+  }));
+
+  const filteredEnums = enums.map((element) => {
+      element.filter((item) => item !== undefined)
+  });
+
   const models = schema
     .getModels()
     .filter((model) => model.getPrimaryKey())
     .map((model) => {
       return createModel(model);
     });
+
+  const modelWithEnums = models.concat(filteredEnums);
+  
   const tsSchema = ts.factory.createCallExpression(
     ts.factory.createPropertyAccessExpression(
       createConfigureExpression(schema, config),
-      ts.factory.createIdentifier(TYPESCRIPT_DATA_SCHEMA_CONSTANTS.SCHEMA_METHOD),
+      ts.factory.createIdentifier(TYPESCRIPT_DATA_SCHEMA_CONSTANTS.SCHEMA_METHOD), // create a.schema()
     ),
     undefined,
-    [ts.factory.createObjectLiteralExpression(models as ts.ObjectLiteralElementLike[], true)],
+    [ts.factory.createObjectLiteralExpression(modelWithEnums as ts.ObjectLiteralElementLike[], true)],
   );
   return ts.factory.createVariableStatement(
     [exportModifier],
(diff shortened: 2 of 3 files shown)


packages/amplify-graphql-schema-generator/src/ts-schema-generator/helpers.ts
```diff
@@ -147,19 +188,92 @@ export const createSchema = (schema: Schema, config?: DataSourceGenerateConfig):
...
     .getModels()
     .filter((model) => model.getPrimaryKey())
     .map((model) => {
-      return createModel(model);
+      return createModel(model, engine);
     });
+
+  const combinedEnums = nullEnums.concat(notNullEnums).flat() // making 1 D array
+
+  if (engine === 'Postgres') { // for postgresql
```
phani-srikar, line 230: For future: Instead of introducing these engine based conditionals that bloats up this method, we could move these to separate methods inside an engine specific controller.

packages/amplify-graphql-schema-generator/src/schema-representation/schema.ts
```diff
@@ -15,6 +15,7 @@ export class Schema {
   }
 
   public addModel(model: Model): void {
+    console.log(this.engine);
```
palpatim, line 18: Is this leftover from dev/debugging? Any reason a customer needs to see this?

packages/amplify-graphql-schema-generator/src/ts-schema-generator/generate-ts-schema.ts
```diff
@@ -18,3 +19,7 @@ export const generateTypescriptDataSchema = (schema: Schema, config?: DataSource
   );
   return result;
 };
+
+
+
+
```
palpatim, line 25: Run `yarn prettier --write .` from the top-level project dir to fix this and other linting errors before submitting.

packages/amplify-graphql-schema-generator/src/ts-schema-generator/helpers.ts
```diff
@@ -46,17 +48,24 @@ const createDataType = (type: FieldType): ts.Node => {
...
-      ts.factory.createIdentifier(`${TYPESCRIPT_DATA_SCHEMA_CONSTANTS.REFERENCE_A}.${TYPESCRIPT_DATA_SCHEMA_CONSTANTS.ENUM_METHOD}`),
-      undefined,
-      [
-        ts.factory.createArrayLiteralExpression(
-          type.values.map((value) => ts.factory.createStringLiteral(value)),
-          true,
-        ),
-      ],
-    );
+    if(engine === 'Postgres'){
```
palpatim, line 53: Looking at the code without context, it looks like we don't handle enum types for MySQL engine. Since we clearly do, let's add a comment explaining the logic about why we're doing this.

packages/amplify-graphql-schema-generator/src/schema-representation/schema.ts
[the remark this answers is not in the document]
tejas2008: Sure I will remove this and add a commit

packages/amplify-graphql-schema-generator/src/ts-schema-generator/helpers.ts
```diff
@@ -84,18 +94,49 @@ const getTypescriptDataSchemaType = (type: string): string => {
...
 
-const createModel = (model: Model): ts.Node => {
+/**
+ * Creates a typescript data schema type from internal SQL schema representation
+ * Example typescript data schema type output: `a.enum()`
+ * @param type SQL IR Enum type
+ * @returns Typescript data schema type in TS Node format
+ */
+const createEnums = (type: EnumType, engine: string, modelName: string, enumName: string): ts.Node => {
+  if (engine === 'Postgres'){
```
palpatim, line 111: nit: since `engine` is a string rather than a strongly typed value, let's make the engine handling in this method explicit and fail for unsupported engine types. That'll protect against typos in calling code. Alternately, use a strongly typed engine type to ensure we are operating on a list of values known at compile time.

packages/amplify-graphql-schema-generator/src/ts-schema-generator/helpers.ts
```diff
@@ -147,19 +188,92 @@ export const createSchema = (schema: Schema, config?: DataSourceGenerateConfig):
...
+  } 
+
+    
+  const modelsWithEnums = models.concat(combinedEnums); // appending enums to models for MySql
+  
   const tsSchema = ts.factory.createCallExpression(
     ts.factory.createPropertyAccessExpression(
       createConfigureExpression(schema, config),
-      ts.factory.createIdentifier(TYPESCRIPT_DATA_SCHEMA_CONSTANTS.SCHEMA_METHOD),
+      ts.factory.createIdentifier(TYPESCRIPT_DATA_SCHEMA_CONSTANTS.SCHEMA_METHOD), // create a.schema()
```
palpatim, line 273: prefer to remove dead code rather than leaving it commented
```suggestion
      ts.factory.createIdentifier(TYPESCRIPT_DATA_SCHEMA_CONSTANTS.SCHEMA_METHOD), 
```

packages/amplify-graphql-schema-generator/src/ts-schema-generator/helpers.ts
```diff
@@ -147,19 +188,92 @@ export const createSchema = (schema: Schema, config?: DataSourceGenerateConfig):
     throw new Error('No valid tables found. Make sure at least one table has a primary key.');
   }
 
+  const engine = schema.getEngine().type;
+
+  const nullEnums = schema  // find null enums in all models
```
palpatim, line 193: naming nit: I think you mean "nullable enum fields" or rather than "null" enums? (The latter implies a concrete value assignment, as opposed to a field type constraint)

packages/amplify-graphql-schema-generator/src/ts-schema-generator/helpers.ts
```diff
@@ -147,19 +188,92 @@ export const createSchema = (schema: Schema, config?: DataSourceGenerateConfig):
...
+        .map((field) => {
+          if(field.type.kind === 'Enum'){
+            return createEnums(field.type, engine, model.getName(), field.name);
+          }
+          else{
+            return undefined;
+          }
+     }));
+
+    const notNullEnums = schema  // find Non null enums in all models
```
palpatim, line 207: naming nit: I think you mean "required" enums or "non null enum types"?

packages/amplify-graphql-schema-generator/src/schema-representation/schema.ts
tejas2008: Yes its a leftover might have missed it I will remove it

packages/amplify-graphql-schema-generator/src/ts-schema-generator/helpers.ts
tejas2008: Sure I will look into it.

packages/amplify-graphql-schema-generator/src/ts-schema-generator/helpers.ts
tejas2008: I will update the naming to have more clear variable names to avoid confusion.

packages/amplify-graphql-schema-generator/src/ts-schema-generator/helpers.ts
tejas2008: I will update the naming to have more clear variable names to avoid confusion. NullEnumFields and RequiredEnumFields makes sense I will go ahead with these naming conventions to provide more readability and understanding

packages/amplify-graphql-schema-generator/src/ts-schema-generator/helpers.ts
tejas2008: It isn't a dead code its a comment, but its not very necessary since it is existing code so will probably remove it

packages/amplify-graphql-schema-generator/src/ts-schema-generator/helpers.ts
```diff
@@ -190,33 +195,35 @@ export const createSchema = (schema: Schema, config?: DataSourceGenerateConfig):
 
   const engine = schema.getEngine().type;
 
-  const nullEnums = schema  // find null enums in all models
+  const nullEnumFields = schema // find null enum fields in all models
```
palpatim, line 198: nit
```suggestion
  const nullableEnumFields = schema
```

packages/amplify-graphql-schema-generator/src/ts-schema-generator/helpers.ts
```diff
@@ -190,33 +195,35 @@ export const createSchema = (schema: Schema, config?: DataSourceGenerateConfig):
...
-          else{
+          } else {
             return undefined;
           }
-     }));
+        }),
+    );
 
-    const notNullEnums = schema  // find Non null enums in all models
+  const RequiredEnumFields = schema // find required enum fields in all models
```
palpatim, line 213: nit: don't PascalCase non-type, non-Class variables.
```suggestion
  const requiredEnumFields = schema
```

packages/amplify-graphql-schema-generator/src/ts-schema-generator/helpers.ts
```diff
@@ -225,25 +232,28 @@ export const createSchema = (schema: Schema, config?: DataSourceGenerateConfig):
       return createModel(model, engine);
     });
 
-  const combinedEnums = nullEnums.concat(notNullEnums).flat() // making 1 D array
+  const combinedEnums = nullEnumFields.concat(RequiredEnumFields).flat(); // making 1 D array
 
-  if (engine === 'Postgres') { // for postgresql
+  /** 
+  for postgresql since enums are declared globally to void declaring duplicate enums we
+  declare find unique enums and then pass it to the schema
```
palpatim, line 239: I can't tell what you mean with this explanation. Can you reword to clarify? (I mean to say, based on our discussion and reading through the code, I know what you're trying to do--it's the comment wording that is unclear.)

Also, nit: reserve the double-asterisk block comment structure (`/** ... */` for docComments (such as we put at the beginning of an exported symbol). Other comments should be delimited with a double-slash (`//`)

tejas2008: Sure I will re word this and and remove the double-asterisk block structure

packages/amplify-graphql-schema-generator/src/ts-schema-generator/helpers.ts
tejas2008: @phani-srikar I have updated the implementation to make it less bloated in the current task as well and more strongly typed.

packages/amplify-graphql-schema-generator/src/ts-schema-generator/helpers.ts
tejas2008: Added appropriate comment structure and comments to explain in brief the use case

packages/amplify-graphql-schema-generator/src/datasource-adapter/mysql-string-datasource-adapter.ts
```diff
@@ -279,12 +280,12 @@ export class MySQLStringDataSourceAdapter extends StringDataSourceAdapter {
...
 
-  private generateEnumName(tableName: string, fieldName: string) {
-    const enumNamePrefix = [tableName, fieldName].join('_');
+  private generateEnumName(tableName: string, fieldName: string): string {
+    const enumNamePrefix = toPascalCase([tableName, fieldName]);
     let enumName = enumNamePrefix;
     let counter = 0;
     while (this.enums.has(enumName)) {
-      enumName = [enumNamePrefix, counter.toString()].join('_');
+      enumName = toPascalCase([enumNamePrefix, counter.toString()]);
```
phani-srikar, line 288: Can you confirm this wouldn't be a breaking change if we change the naming of enums from `table_field` to `TableField`?

tejas2008: As per the discussion @palpatim the change could affect customers only if they reimport the schema which is highly unlikely, but we still are waiting for conformation from Josef on the same

packages/amplify-graphql-schema-generator/src/ts-schema-generator/generate-ts-schema.ts
```diff
@@ -1,6 +1,7 @@
 import * as ts from 'typescript';
 import { Schema } from '../schema-representation';
 import { createImportExpression, createSchema, DataSourceGenerateConfig } from './helpers';
+import { LogFormat } from '@aws-sdk/client-lambda';
```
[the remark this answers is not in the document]
tejas2008: Removed it

packages/amplify-graphql-schema-generator/src/ts-schema-generator/helpers.ts
tejas2008: Done

packages/amplify-graphql-schema-generator/src/ts-schema-generator/helpers.ts
tejas2008: addressed it

packages/amplify-graphql-schema-generator/src/ts-schema-generator/helpers.ts
```diff
@@ -147,19 +164,61 @@ export const createSchema = (schema: Schema, config?: DataSourceGenerateConfig):
...
+    model
+      .getFields()
+      .filter((field) => field.type.kind === 'NonNull' && field.type.type.kind === 'Enum')
+      .map((field) => {
+        if (field.type.kind === 'NonNull' && field.type.type.kind === 'Enum') {
+          return createEnums(field.type.type);
+        } else {
+          return undefined;
+        }
+      }),
```
p5quared, line 190: Is checking `field.type.kind` and `field.type.type.kind` redundant because of the filter?
```suggestion
      .map((field) => createEnums(field.type.type))
```

p5quared (approved):

PR #2897, opened 2024-09-24
refactor(conversation-transformer): use file templates for JS resolvers

atierian:
- moves string defined JS resolvers to JavaScript file templates using `[[KEY]]` marker for template substitution.
- adds `copy-js-resolver-templates` to move JS templates to `/lib` as part of `build` script.
- adds `packages/amplify-graphql-conversation-transformer/src/resolvers/*.js` to `.prettierignore`

**Why use `[[KEY]]` as the marker?**
This marker format satisfies two key goals:
1. Minimal risk of overwriting previously substituted values, e.g.  the value of `[[KEY1]]` containing `"[[KEY2]]"`.
2. Not break syntax highlighting in the JS file templates. The motivation behind moving to file templates is to improve maintainability of the resolver functions, for which syntax highlighting plays a vital role.

I'm satisfied that `[[KEY]]` is the appropriate format here. That said if you have an alternative in mind that satisfies both 1. and 2., and has other benefits, I'm happy to consider it.

N/A

#### Issue #, if available
N/A

- CI PR workflow
- locally run conversation E2E tests

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [ ] ~E2E test run linked~


packages/amplify-graphql-conversation-transformer/src/resolvers/invoke-lambda-resolver.ts
```diff
@@ -14,27 +16,42 @@ export const invokeLambdaMappingTemplate = (
...
+  let resolver = fs.readFileSync(path.join(__dirname, 'invoke-lambda-resolver-fn.js'), 'utf8');
+  Object.entries(substitutions).forEach(([key, value]) => {
+    const replaced = resolver.replace(new RegExp(`\\[\\[${key}\\]\\]`, 'g'), value);
+    resolver = replaced;
+  });
+
+  // This unfortunately needs to be an inline template because an s3 mapping template doesn't allow the CDK
+  // to substitute token values, which is necessary for this resolver function due to its reference of
+  // `ctx.api.graphqlUrl`.
+  return MappingTemplate.inlineTemplateFromString(resolver);
```
atierian, line 47: In a follow up, I'm planning on moving the `ctx.api.graphqlUrl` reference to the "Resolver code" portion of the pipeline along with the existing `ctx.stash.` `typeName` / `fieldName` / etc values.

This will allow us to make this an S3 asset and support hotswapping for schema defined values like system prompt, tools, model (LLM), and model inference configuration.

packages/amplify-graphql-conversation-transformer/src/resolvers/invoke-lambda-resolver-fn.js
```diff
@@ -0,0 +1,56 @@
+import { util } from '@aws-appsync/utils';
+
+export function request(ctx) {
```
palpatim, line 3: For future PR:
1. Please add a doc comment noting the replacement logic
2. Rename the file to something other than `.js` (e.g., `.js.template` or something), since this isn't actually executable JS as-is

packages/amplify-graphql-conversation-transformer/src/resolvers/invoke-lambda-resolver-fn.js
```diff
@@ -0,0 +1,56 @@
...
+  const { args, request, prev } = ctx;
+  [[TOOL_DEFINITIONS_LINE]]
```
palpatim, line 5: Here and throughout: Does your template substitution logic guarantee that these will be:
1. Always one line?
2. Always appropriately escaped for the expected quotes?

How can we ensure that the template always uses the expected quoting scheme so the caller passing the substitution values always escapes appropriately?

packages/amplify-graphql-conversation-transformer/src/resolvers/invoke-lambda-resolver.ts
```diff
@@ -14,27 +16,42 @@ export const invokeLambdaMappingTemplate = (
...
+    GRAPHQL_API_ENDPOINT,
+    MODEL_CONFIGURATION_LINE,
+    RESPONSE_MUTATION_NAME,
+    RESPONSE_MUTATION_INPUT_TYPE_NAME,
+    MESSAGE_MODEL_NAME,
+  };
+
+  let resolver = fs.readFileSync(path.join(__dirname, 'invoke-lambda-resolver-fn.js'), 'utf8');
+  Object.entries(substitutions).forEach(([key, value]) => {
+    const replaced = resolver.replace(new RegExp(`\\[\\[${key}\\]\\]`, 'g'), value);
```
palpatim, line 40: One day Node will give us a `String.prototype.replaceAll` method so we can forego this nonsense. :(

packages/amplify-graphql-conversation-transformer/src/resolvers/verify-session-owner-resolver-fn.js
```diff
@@ -0,0 +1,30 @@
+export function request(ctx) {
+  const { authFilter } = ctx.stash;
+
```
palpatim, line 3: Here and throughout the PR: let's add a defensive existence assertion for `authFilter`. I'm comfortable that the existing implementation sets it appropriately, but if a future update accidentally forgets it, I'd prefer to catch that early instead of performing a query without the filter in place. (It's possible that the `toDynamoDBFilterExpression` util below would fail in that case, but I'd prefer to make the check explicit.)

packages/amplify-graphql-conversation-transformer/src/resolvers/write-message-to-table-resolver-fn.js
```diff
@@ -0,0 +1,24 @@
+import { util } from '@aws-appsync/utils';
+import * as ddb from '@aws-appsync/utils/dynamodb';
+
+export function request(ctx) {
```
palpatim, line 4: as above, let's rename this and anything else that runs through substitution logic to indicate that it's not executable JS as-is.

packages/amplify-graphql-conversation-transformer/src/resolvers/write-message-to-table-resolver.ts
```diff
@@ -1,63 +1,23 @@
...
-      } else {
-        return ctx.result;
-      }
-    }
-    `;
+  const substitutions = {
+    CONVERSATION_MESSAGE_TYPE_NAME: `ConversationMessage${fieldName}`,
+  };
+  let resolver = fs.readFileSync(path.join(__dirname, 'write-message-to-table-resolver-fn.js'), 'utf8');
+  Object.entries(substitutions).forEach(([key, value]) => {
```
palpatim, line 16: nit for future PR: recommend centralizing substitution logic in a utility somewhere. That util can take care of the file renaming logic as well to help enforce the convention.

packages/amplify-graphql-conversation-transformer/src/transformer-steps/conversation-resolver-generator.ts
```diff
@@ -48,14 +48,22 @@ export class ConversationResolverGenerator {
...
-    this.createAssistantResponseSubscriptionResolver(ctx, capitalizedFieldName);
+    this.createAssistantResponseSubscriptionResolver(ctx, directive, capitalizedFieldName);
 
     const functionDataSource = this.addLambdaDataSource(ctx, functionDataSourceId, referencedFunction, capitalizedFieldName);
     const invokeLambdaFunction = invokeLambdaMappingTemplate(directive, ctx);
 
     this.setupMessageTableIndex(ctx, directive);
 
-    this.createConversationPipelineResolver(ctx, parentName, fieldName, capitalizedFieldName, functionDataSource, invokeLambdaFunction);
+    this.createConversationPipelineResolver(
```
palpatim, line 58: nit: recommend an options object rather than 7 params

packages/amplify-graphql-conversation-transformer/src/resolvers/write-message-to-table-resolver.ts
atierian: 💯  that's the plan. I felt this PR was already too large as is and want to do this in a separate change.

packages/amplify-graphql-conversation-transformer/src/transformer-steps/conversation-resolver-generator.ts
atierian: What I have in mind for https://github.com/aws-amplify/amplify-category-api/pull/2897#discussion_r1777331065 will include cleaning this up.

packages/amplify-graphql-conversation-transformer/src/resolvers/invoke-lambda-resolver-fn.js
atierian: Good call. 
I addressed 2. in b05fbac4e071750d75b5c3523546ca0aa2245ed9 because it's core to the scope of change.
I'll follow up for 1. in https://github.com/aws-amplify/amplify-category-api/pull/2897#discussion_r1777331065

packages/amplify-graphql-conversation-transformer/src/resolvers/invoke-lambda-resolver-fn.js
atierian: > 1. Always one line?

Not always one line. For example, `[[TOOL_DEFINITIONS_LINE]]` here can be:
```js
const dataTools = toolDefinitions.tools;
const toolsConfiguration = {
  dataTools,
  clientTools,
};
```

> 2. Always appropriately escaped for the expected quotes?

The only safeguard today are the existing snapshots, which is admittedly not great. I plan to add integration tests using AppSync's `EvaluateCode` API for the resolver functions in the future ([reference](https://docs.aws.amazon.com/appsync/latest/devguide/test-debug-resolvers-js.html)). That will give us confidence in future changes.

> How can we ensure that the template always uses the expected quoting scheme so the caller passing the substitution values always escapes appropriately?

Same answer as 2. Additional context -- the majority of these values are derived from GraphQL inputs (field name, structured directive `input` types) where values that would cause escaping issues would have failed upstream in GraphQL validation. The exceptions to this rule are the model identifier and system prompt, which are string inputs provided when defining a conversation route. The system prompt is already stringified to prevent issues here. The model identifier can also be stringified prior to substitution -- I'll look at this in a follow up.

packages/amplify-graphql-conversation-transformer/src/resolvers/verify-session-owner-resolver-fn.js
atierian: Unless you strongly object, I'll include this in a follow up (I have a draft PR queued up that makes some changes here). I know that this is a safe (and correct) change to make, but I'd really prefer to keep this PR non-functional.

packages/amplify-graphql-conversation-transformer/src/resolvers/write-message-to-table-resolver-fn.js
atierian: addressed in https://github.com/aws-amplify/amplify-category-api/commit/b05fbac4e071750d75b5c3523546ca0aa2245ed9

packages/amplify-graphql-conversation-transformer/src/resolvers/verify-session-owner-resolver-fn.js
atierian: Noted in the aforementioned draft PR as a reminder.
https://github.com/aws-amplify/amplify-category-api/pull/2910#issuecomment-2377392782

palpatim (dismissed):

palpatim (approved):

sundersc (approved):

PR #2899, opened 2024-09-24
feat: exposing stack property for data construct

ShadowCat567:
#### Description of changes

- exposes stack property for data construct, comes from open issue with cli team
- https://github.com/aws-amplify/amplify-backend/issues/1242

#### Issue #, if available 
N/A -- issue comes from `amplify-backend`

- wrote new tests that ensure the stack property exists and that it returns the expected value

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [ ] E2E test run linked


packages/amplify-graphql-api-construct/src/__tests__/__functional__/stack.test.ts
```diff
@@ -0,0 +1,35 @@
...
+            type Todo @model @auth(rules: [{ allow: public }]) {
+            description: String!
+            }
+        `),
+        authorizationModes: {
+            apiKeyConfig: { expires: cdk.Duration.days(7) },
+        },
+    });
+
+    expect(api.stack).toBe(cdk.Stack.of(api));
```
sobolk, line 33:
```suggestion
  it('verifies the stack property is set to the expected value', () => {
    const stack = new cdk.Stack();
    const api = new AmplifyGraphqlApi(stack, 'TestApi', {
        definition: AmplifyGraphqlDefinition.fromString(/* GraphQL */ `
            type Todo @model @auth(rules: [{ allow: public }]) {
            description: String!
            }
        `),
        authorizationModes: {
            apiKeyConfig: { expires: cdk.Duration.days(7) },
        },
    });

    expect(api.stack).toBe(stack);
```


I think the first test is redundant with this one. `expect(api.stack).toBe(stack);` already asserts that prop is defined.


An idea for additional test would be something like this. (pseudocode)

```

class CustomL3Construct extends Construct {
 data: DataConstruct;
 constructor ( scope, ..whatever else here) {
   this.data = new DataConstruct(scope);
 }

}

// and then in test

const stack = new Stack();
const customConstruct = new CustomL3Construct(stack);

assert( customConstruct.data.stack).toBe(stack);
```

I.e. assert that we traverse tree upwards from data construct to find nearest parent stack.

ShadowCat567: Would the suggested test be something that might make more sense in the `amplify-backend` tests? Since we would have an easier time traversing the stack tree in there

packages/amplify-graphql-api-construct/src/amplify-graphql-api.ts
```diff
@@ -87,6 +87,8 @@ export class AmplifyGraphqlApi extends Construct {
    */
   public readonly resources: AmplifyGraphqlApiResources;
 
+  public readonly stack: Stack;
```
palpatim, line 90: Add doc comment explaining what stack this refers to

packages/amplify-graphql-api-construct/src/__tests__/__functional__/stack.test.ts
sobolk: > Would the suggested test be something that might make more sense in the amplify-backend tests?

for auth, function and storage yes. We have not explicitly tested that parent stack might be couple of levels up. But for data such test should be here.

On the other hand we put some trust in `Stack.of` API to do the right thing, so maybe it would be too paranoid. I.e. should we test that CDK works? Feel free to ignore that suggestion.

packages/amplify-graphql-api-construct/src/amplify-graphql-api.ts
```diff
@@ -87,6 +87,9 @@ export class AmplifyGraphqlApi extends Construct {
    */
   public readonly resources: AmplifyGraphqlApiResources;
 
+  /**
+   * Reference to root stack of data construct
```
sobolk, line 91:
```suggestion
   * Reference to parent stack of data construct
```

or `nearest parent` .  "Root stack" might be something different depending on shape of construct tree.

packages/amplify-graphql-api-construct/src/__tests__/__functional__/stack.test.ts
ShadowCat567: I think it would be a good idea, making the CustomL3Construct is what I am a little unsure about, I'll look around and see if something like that has already been made

sobolk: > making the CustomL3Construct is what I am a little unsure about

It would be something that lives only in the test and can be empty construct. So that we can test stack->some_construct->data_construct nesting.

This might be helpful resource https://docs.aws.amazon.com/cdk/v2/guide/constructs.html .

sobolk (dismissed):

packages/amplify-graphql-api-construct/src/amplify-graphql-api.ts
ShadowCat567: it has been added 👍

palpatim (approved):

bobbyu99 (approved):

PR #2908, opened 2024-09-26
Fix migration e2es

dpilch:
#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes
- [ ] E2E test run linked

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "e2e": "lerna run e2e",
     "extract-api": "lerna run extract-api",
     "extract-dependency-licenses": "./scripts/extract-dependency-licenses.sh",
-    "hoist-cli-win": "rimraf node_modules/amplify-cli-internal && mkdir node_modules\\amplify-cli-internal && xcopy /e /q node_modules\\@aws-amplify\\cli-internal node_modules\\amplify-cli-internal\\",
+    "hoist-cli-win": "rimraf node_modules/amplify-cli-internal@12.13.0-gen2-migration-test-alpha.0 && mkdir node_modules\\amplify-cli-internal && xcopy /e /q node_modules\\@aws-amplify\\cli-internal node_modules\\amplify-cli-internal\\",
     "hoist-cli": "rm -rf node_modules/amplify-cli-internal && mkdir node_modules/amplify-cli-internal && cp -r node_modules/@aws-amplify/cli-internal/* node_modules/amplify-cli-internal",
     "link-dev": "cd node_modules/amplify-cli-internal && ln -s \"$(pwd)/bin/amplify\" \"$(yarn global bin)/amplify-dev\" && cd -",
     "link-win": "node ./scripts/link-bin.js node_modules/amplify-cli-internal/bin/amplify amplify-dev",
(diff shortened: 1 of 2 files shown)


packages/amplify-e2e-tests/schemas/references.graphql
```diff
@@ -1,17 +1,17 @@
 type Primary @model @auth(rules: [{ allow: public }]) {
-  id: ID! @primaryKey
```
dpilch, line 2: Just removing this to simplify the test case.

packages/amplify-graphql-model-transformer/src/resources/amplify-dynamodb-table/amplify-table-manager-lambda/amplify-table-manager-handler.ts
```diff
@@ -1560,4 +1578,66 @@ const importTable = async (tableDef: CustomDDB.Input): Promise<AWSCDKAsyncCustom
   return result;
 };
 
+export const getImportedTableComparisonProperties = (importedTable: TableDescription): TableComparisonProperties => {
```
palpatim, line 1581: 1. Does this need to be exported? It looks like it's used above, but I can't tell that it's used elsewhere (maybe it's exported for tests though)
2. This file is already pretty long & unwieldy -- can we refactor the table comparison logic and this new helper into a separate set of files to make it easier to figure out what the heck is going on in here? I'd also suggest you create field-specific helpers for any field that's not a straight up `ComparisonField: importedTable.Field` assignment, again to improve readability. The current structure makes it pretty hard to parse.

dpilch: 1. Exported for tests
2. Would you prefer this in a follow up or done in this PR?

palpatim (approved):

packages/amplify-graphql-model-transformer/src/resources/amplify-dynamodb-table/amplify-table-manager-lambda/import-table.ts
```diff
@@ -0,0 +1,293 @@
...
+};
+
+const getExpectedStreamSpecification = (createTableInput: CreateTableCommandInput): TableComparisonProperties['StreamSpecification'] => {
+  return createTableInput.StreamSpecification;
+};
+
+const getExpectedProvisionedThroughput = (
+  createTableInput: CreateTableCommandInput,
+): TableComparisonProperties['ProvisionedThroughput'] => {
+  return createTableInput.ProvisionedThroughput || { ReadCapacityUnits: 0, WriteCapacityUnits: 0 };
```
phani-srikar, line 177: Any reasoning for the defaulted throughput values here? Do these match the current defaults we use in provisioning?

packages/amplify-graphql-model-transformer/src/resources/amplify-dynamodb-table/amplify-table-manager-lambda/import-table.ts
```diff
@@ -0,0 +1,293 @@
...
+        Status: 'ENABLED',
+      }
+    : undefined;
+};
+
+const getExpectedDeletionProtectionEnabled = (
+  createTableInput: CreateTableCommandInput,
+): TableComparisonProperties['DeletionProtectionEnabled'] => {
+  return createTableInput.DeletionProtectionEnabled || false;
+};
```
phani-srikar, line 193: similar question about the default choices here.

phani-srikar (approved):

PR #2920, opened 2024-10-03
chore: migrate pg model dynamic auth e2e test in gen2 cdk

Siqi-Shan:
#### Description of changes

Migrate the Gen1 model dynamic auth E2E tests to Gen2 using CDK construct, with pre-provision of Aurora clusters and Cognito resources. Plus fix/change on resource pre-provision process and CDK setup.

Update: this PR also includes refactor of the Gen2 Postgres auto increment E2E test under same pattern.

CI checks and local pre-provision RDS clusters.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] E2E test run linked


packages/amplify-graphql-api-construct-tests/src/__tests__/backends/sql-configurable-stack/app.ts
```diff
@@ -0,0 +1,310 @@
...
+    }
+    case AUTH_MODE.AMAZON_COGNITO_USER_POOLS: {
+      const { userPool, userPoolClient } = createUserPool(dbDetails.dbConfig.dbName);
+
+      authorizationModes = {
+        defaultAuthorizationMode: 'AMAZON_COGNITO_USER_POOLS',
+        userPoolConfig: {
+          userPool,
+        },
+        apiKeyConfig: { expires: Duration.days(2) },
```
phani-srikar, line 141: Do we need the API key to be valid for 2 days? Especially since we teardown these apps post test run, I would recommend setting it to 1 hr to begin with.

packages/amplify-graphql-api-construct-tests/src/utils/sql-crudl-helper.ts
```diff
@@ -47,9 +47,9 @@ export type FieldMap = {
 export class CRUDLHelper {
   constructor(
     private readonly appSyncClient: AWSAppSyncClient<any>,
-    private readonly modelName: string,
-    private readonly modelListName: string,
-    private readonly fieldMap: FieldMap,
+    private readonly modelName?: string,
+    private readonly modelListName?: string,
+    private readonly fieldMap?: FieldMap,
```
p5quared, line 52: Why are these optional now?

packages/amplify-graphql-api-construct-tests/src/__tests__/backends/sql-configurable-stack/app.ts
Siqi-Shan: Line 75 of file `authorization-modes.ts` in `amplify-graphql-api-construct` enforces the `Duration` would be converted to days with following definition: `apiKeyExpirationDays: authMode.expires.toDays()`

Setting expiration to hours would cause CDK deployment to fail with error message:
- "Error: '1 hours' cannot be converted into a whole number of days."

We could adjust the construct time conversion if needed to give test a more tight time boundary.

packages/amplify-graphql-api-construct-tests/src/sql-datatabase-controller.ts
```diff
@@ -285,14 +288,20 @@ export class SqlDatatabaseController {
...
-    console.log(`Wrote DB details at ${filePath}`);
+    const stackConfigStr = JSON.stringify(stackConfig ?? {});
+
+    const dbDetailsfilePath = path.join(projRoot, 'db-details.json');
+    fs.writeFileSync(dbDetailsfilePath, detailsStr);
+    console.log(`Wrote DB details at ${dbDetailsfilePath}`);
+
+    const stackConfigFilePath = path.join(projRoot, 'stack-config.json');
+    fs.writeFileSync(stackConfigFilePath, stackConfigStr);
+    console.log(`Wrote stack config to ${stackConfigFilePath}`);
```
p5quared, line 304: If `stackConfigStr`  is `{}` perhaps there's no need to write anything and that could be logged instead.

p5quared, line 304: If `stackConfigStr`  is `{}` perhaps there's no need to write anything and that could be logged instead?

p5quared (approved):

packages/amplify-graphql-api-construct-tests/src/sql-datatabase-controller.ts
p5quared, line 304: Perhaps if `stackConfig` is `{}` there's no need to write anything and instead log that?

packages/amplify-graphql-api-construct-tests/src/utils/sql-crudl-helper.ts
p5quared, line 52: Why are these optional?

packages/amplify-graphql-api-construct-tests/src/__tests__/backends/sql-configurable-stack/app.ts
```diff
@@ -0,0 +1,310 @@
...
+enum AUTH_MODE {
+  API_KEY = 'API_KEY',
+  AWS_IAM = 'AWS_IAM',
+  AMAZON_COGNITO_USER_POOLS = 'AMAZON_COGNITO_USER_POOLS',
+  OPENID_CONNECT = 'OPENID_CONNECT',
+  AWS_LAMBDA = 'AWS_LAMBDA',
+}
+
+enum SCHEMA {
+  DEFAULT = /* GraphQL */ `
```
phani-srikar, line 29: Why would we represent this as an enum vs a string?

Siqi-Shan: Both options are viable here. My thought originally was that there could be multiple existing schemas available in the stack which could be used without providing schema through `stack-config` but more like an option. But that could be an overdesign here, as almost every test's schema is different, and it's impossible to store all those schemas in the CDK file.

p5quared: I think maybe a similar question is why enum versus an object with a string field `default`?

phani-srikar: Do we have to re-define these Auth modes and set a default for schema? Instead we could just fail if either of them is not defined in the test using it.

packages/amplify-graphql-api-construct-tests/src/__tests__/backends/sql-configurable-stack/app.ts
phani-srikar: > We could adjust the construct time conversion if needed to give test a more tight time boundary.

That might be the right way to go forward here.

packages/amplify-graphql-api-construct-tests/src/utils/sql-provider-helper.ts
```diff
@@ -0,0 +1,131 @@
...
+    case 'Int':
+      return 'INT';
+    case 'Float':
+      return 'FLOAT';
+    case 'Boolean':
+      return 'BOOLEAN';
+    case 'AWSDateTime':
+      return 'DATETIME';
+    default:
+      return 'VARCHAR(255)';
```
phani-srikar, line 114: nit: un-recognized type can just throw an error here, so we know how to handle those.

packages/amplify-graphql-api-construct-tests/src/utils/sql-provider-helper.ts
```diff
@@ -0,0 +1,131 @@
...
+};
+
+const getArrayOrObjectFieldType = (fieldType: TypeNode, engine: ImportedRDSType): string => {
+  switch (engine) {
+    case ImportedRDSType.MYSQL:
+      return 'JSON'; // MySQL does not support array types
+    case ImportedRDSType.POSTGRESQL:
+      return isListType(fieldType) ? 'VARCHAR[]' : 'json';
+    default:
+      return 'VARCHAR[]';
```
phani-srikar, line 96: can we just throw an error here or we are not sure what tests are impacted?

packages/amplify-graphql-api-construct-tests/src/utils/sql-provider-helper.ts
```diff
@@ -0,0 +1,131 @@
...
+        const fieldsToAdd = node.fields.filter((field) => !isRelationalField(field));
+        fieldsToAdd.forEach((field, index) => {
+          fieldStatements.push(getFieldStatement(field, index === 0, engine));
+        });
+        const sql = `CREATE TABLE ${convertToDBSpecificName(tableName, engine)} (${fieldStatements.join(', ')});`;
+        sqlStatements.push(sql);
+      },
+    },
+  };
+  visit(document, schemaVisitor);
```
phani-srikar, line 52: +1 to using the visitor pattern here.

packages/amplify-graphql-api-construct-tests/src/utils/sql-crudl-helper.ts
```diff
@@ -211,4 +211,58 @@ export class CRUDLHelper {
...
+    }
+  };
+
+  public checkListItemExistence = (result: any, primaryKeyValue: string, shouldExist = false, primaryKeyName = 'id') => {
+    expect(result).toBeDefined();
+    expect(result.items).toBeDefined();
+    expect(result.items?.filter((item: any) => item[primaryKeyName] === primaryKeyValue)?.length).toEqual(shouldExist ? 1 : 0);
+  };
+
+  public runCustomMutate = async (mutation: string, input: any): Promise<Record<string, any>> => {
```
phani-srikar, line 249: nit: `runCustomMutation`.

phani-srikar (changes requested):

packages/amplify-graphql-api-construct-tests/src/utils/sql-crudl-helper.ts
Siqi-Shan: renamed

packages/amplify-graphql-api-construct-tests/src/__tests__/backends/sql-configurable-stack/app.ts
Siqi-Shan: I think we could keep the Auth mode to sync with the stack config interface that is using the `AUTH_TYPE` from AppSync, while schema default could be discarded to avoid overdesign, since schema is already required when defining the stack config interface.

packages/amplify-graphql-api-construct-tests/src/utils/sql-provider-helper.ts
Siqi-Shan: now throw error in this case

packages/amplify-graphql-api-construct-tests/src/utils/sql-provider-helper.ts
Siqi-Shan: now throw error in this case

phani-srikar (approved):

p5quared (approved):

PR #2921, opened 2024-10-03
fix: add aws_iam to custom operations when enableIamAuthorization is enabled

palpatim:
#### Description of changes

* Add aws_iam to custom operations when enableIamAuthorization is enabled. Fixes https://github.com/aws-amplify/amplify-category-api/issues/2837
* Fix graphql type utils -- previously the return types of some type predicates were incorrectly grouped

**Merge conflicts**
* packages/amplify-graphql-transformer-core/API.md
  * Regenerated

#### Issue #, if available

https://github.com/aws-amplify/amplify-category-api/issues/2837

- Note: This PR leaves #2929 unfixed. See that issue for discussion.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] E2E test run [linked](https://us-east-1.console.aws.amazon.com/codesuite/codebuild/594813022831/projects/amplify-category-api-e2e-workflow/batch/amplify-category-api-e2e-workflow:c6890889-634b-4799-b824-044f301c894e?region=us-east-1)
    - [x] Passes
    - Test `schema_auth_15` fails because the prompt handling for that test hasn't been updated to reflect the new CLI Gen 1 prompts as in https://github.com/aws-amplify/amplify-category-api/pull/2924. Fixed in 5965fab87
    - Test `RelationalWithAuthV2Redacted.e2e.test.ts` fails b/c of known bucket naming conflict. Passes locally


renebrandel (approved):

packages/amplify-graphql-auth-transformer/src/graphql-auth-transformer.ts
```diff
@@ -345,8 +344,44 @@ export class AuthTransformer extends TransformerAuthBase implements TransformerA
...
+
+    const makeAwsIamDirective = (): DirectiveNode => {
+      return {
+        kind: 'Directive',
+        name: {
+          kind: 'Name',
+          value: 'aws_iam',
+        },
+      };
+    };
```
atierian, line 368: nit
```suggestion
    const makeAwsIamDirective = () => makeDirective('aws_iam', []);
```

atierian (commented):
From what I can tell this is adding `@aws_iam` to operations, but not to any types those operations may return. In Gen2 world, I'm thinking about this scenario:
```ts
const schema = a.schema({
  Foo: a.customType({
    bar: a.string()
  }),
  myFooQuery: a.query()
    // ..
    .returns(a.ref(Foo)),
});
```

```graphql
type Foo { 
  bar: String
} # does this have @aws_iam?

type Query {
  myFooQuery: Foo @aws_iam
}
```

Do those types already have `@aws_iam` / am I missing something in this PR / is it planned in a follow up?

sundersc (dismissed):

palpatim:
Converting this to draft while we figure out the right way of implementing automatically decorating custom return types, as in:

```graphql
type Foo {
  description: String
}
type Query {
  getFooCustom: Foo
}
```

palpatim:
Will also discuss whether we want to just release this fix as-is, to support custom operations that return scalars. We can then take on custom operations that return non-model types.

palpatim:
Decided to release this fix for scalar return values, and opened https://github.com/aws-amplify/amplify-category-api/issues/2929 to track the complex custom type issue.

packages/amplify-graphql-auth-transformer/src/graphql-auth-transformer.ts
```diff
@@ -345,8 +344,34 @@ export class AuthTransformer extends TransformerAuthBase implements TransformerA
...
+   * aws_iam auth directives.
+   */
+  addCustomOperationFieldsToAuthNonModelConfig = (ctx: TransformerTransformSchemaStepContextProvider): void => {
+    if (!ctx.transformParameters.sandboxModeEnabled && !ctx.synthParameters.enableIamAccess) {
+      return;
+    }
+
+    const hasAwsIamDirective = (field: FieldDefinitionNode): boolean => {
+      return field.directives?.some((dir) => dir.name.value === 'aws_iam');
+    };
```
phani-srikar, line 358: nit: We already have a centralized directive existence checker on Object types [defined in the common package]( https://github.com/aws-amplify/amplify-category-api/blob/9dde73c09dd4f42c666c256c8543a6b94dc4a084/packages/graphql-transformer-common/src/definition.ts#L436) for field types so it is re-usable across transformers

phani-srikar (approved):

atierian (approved):

p5quared (approved):

packages/amplify-graphql-auth-transformer/src/graphql-auth-transformer.ts
palpatim: Thanks, will address in a followup PR!

packages/amplify-graphql-auth-transformer/API.md
```diff
@@ -110,6 +110,8 @@ export class AuthTransformer extends TransformerAuthBase implements TransformerA
     // (undocumented)
     addAutoGeneratedRelationalFields: (ctx: TransformerContextProvider, def: ObjectTypeDefinitionNode, allowedFields: Set<string>, fields: readonly string[]) => void;
     // (undocumented)
+    addCustomOperationFieldsToAuthNonModelConfig: (ctx: TransformerTransformSchemaStepContextProvider) => void;
```
palpatim, line 113: note: this should be private. Will make it so in a future PR, which will remove it from the published API. Since the transformer interface isn't actually used by customers, this won't be a breaking change

svidgen (approved):

PR #2937, opened 2024-10-09
fix: s3 bucket creation error

bobbyu99:
#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes
- [ ] E2E test run linked


packages/graphql-transformers-e2e-tests/src/__tests__/RelationalWithAuthV2NonRedacted.e2e.test.ts
```diff
@@ -38,7 +38,7 @@ const cf = new CloudFormationClient(region);
 const customS3Client = new S3Client(region);
 const awsS3Client = new S3({ region: region });
 const cognitoClient = new CognitoClient({ apiVersion: '2016-04-19', region: region });
-const BUILD_TIMESTAMP = moment().format('YYYYMMDDHHmmss');
+const BUILD_TIMESTAMP = moment().format('YYYYMMDDHHmmssSSS');
```
phani-srikar, line 41: Instead of extending the existing reliability on a timestamp to disambiguate the bucket across tests which resulted in the flakiness, we could update the name of bucket to reflect the test name, so we don't run into such cases.

PR #2952, opened 2024-10-14
feat: generate ts schema from pg serial field

p5quared:
#### Description of changes

Update TS schema generator to properly annotate serial fields from Postgres datasources.

e.g.:
`myField SERIAL` -> `myField: a.integer().default()`

Explanation: 
We already populate `field.default` when importing a DB schema, but don't really do anything with it prior to this change. 

Now, we check if the field default is `DB_GENERATED`, and the value for the generated field indicates it is a `SERIAL` column. If this is the case, we annotate the TS schema field with `.default()`.

**N/A**

#### Issue #, if available

**N/A**

See test cases in 
* `packages/amplify-graphql-schema-generator/src/__tests__/generate-ts-data-schema.test.ts`
* `packages/amplify-graphql-schema-generator/src/__tests__/pg-string-datasource-adapter.test.ts`

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] [E2E test run linked](https://us-east-1.console.aws.amazon.com/codesuite/codebuild/594813022831/projects/amplify-category-api-e2e-workflow/batch/amplify-category-api-e2e-workflow:a6c67ff2-6e52-47d7-9323-72a02fa5e8d8?region=us-east-1) 


packages/amplify-graphql-schema-generator/src/__tests__/__snapshots__/generate.test.ts.snap
```diff
@@ -145,6 +145,19 @@ type Story @refersTo(name: \\"stories\\") @model {
 "
 `;
 
+exports[`generate graphqlSchemaFromSQLSchema creates graphql schema from "postgres" "serial" schema 1`] = `
```
p5quared, line 148: This snapshot is not intentional as we are not intending to support GQL generation here. It is generated automatically to test against anyway from the schemas in `__utils__`.

packages/amplify-graphql-schema-generator/src/__tests__/generate-ts-data-schema.test.ts
```diff
@@ -249,6 +249,108 @@ describe('Type name conversions', () => {
     expect(graphqlSchema).toMatchSnapshot();
   });
 
+  it.each([
```
p5quared, line 252: These cases are not exhaustive over every `Field` that could exist but cover the most unique cases.

packages/amplify-graphql-schema-generator/src/ts-schema-generator/helpers.ts
```diff
@@ -27,32 +27,46 @@ const GQL_TYPESCRIPT_DATA_SCHEMA_TYPE_MAP = {
...
  * Creates a typescript data schema type from internal SQL schema representation
  * Example typescript data schema type output: `a.string().required()`
- * @param type SQL IR field type
+ * @param field SQL IR field
  * @returns Typescript data schema type in TS Node format
  */
-const createDataType = (type: FieldType): ts.Node => {
-  if (type.kind === 'Scalar') {
+const createDataType = (field: Field): ts.Node => {
+  const sequenceRegex = /^nextval\(.+::regclass\)$/;
```
p5quared, line 41: [Regex Explain](https://regexr.com/871ku)

I went back and forth on a few approaches to this check but I eventually felt that a quick REGEX is the simplest surefire way to identify `SERIAL` fields. In the future I think this logic could be extracted into something like `if (canLetDBGenerate(field))` which evaluates rules `isSerial(field) || isUserDefinedFunction(field)` etc... But I felt that might premature for this implementation.

packages/amplify-graphql-schema-generator/src/__tests__/__snapshots__/generate-ts-data-schema.test.ts.snap
```diff
@@ -245,6 +245,31 @@ export const schema = configure({
...
+    database: {
+        identifier: \\"ID1234567890\\",
+        engine: \\"postgresql\\",
+        connectionUri: secret(\\"CONN_STR\\")
+    }
+}).schema({
+    \\"CoffeeQueue\\": a.model({
+        id: a.integer().default(),
+        name: a.string(),
+        orderNumber: a.integer().default()
```
p5quared, line 265: This is the desired behavior of this PR.

packages/amplify-graphql-schema-generator/src/__tests__/__snapshots__/pg-string-datasource-adapter.test.ts.snap
```diff
@@ -279,3 +279,38 @@ Array [
   },
 ]
 `;
+
+exports[`testPostgresStringDataSourceAdapter sets the model field correctly from a schema with a SERIAL field 1`] = `
```
p5quared, line 283: This behavior is not new but I am codifying it since generation depends on it now.

packages/amplify-graphql-schema-generator/src/ts-schema-generator/helpers.ts
p5quared: For more context, here's what a value for `column_default` looks like on a serial field when we query the DB schema:
`nextval('serial_table_number_seq'::regclass)`.

packages/amplify-graphql-schema-generator/src/__tests__/__snapshots__/generate.test.ts.snap
```diff
@@ -145,6 +145,19 @@ type Story @refersTo(name: \\"stories\\") @model {
...
+"input AMPLIFY {
+ engine: String = \\"postgres\\"  
+}
+
+
+type SerialTable @refersTo(name: \\"      serial_table\\") @model {
```
sundersc, line 154: Do we know why we have blank spaces before the table name for `refersTo` directive?

packages/amplify-graphql-schema-generator/src/ts-schema-generator/helpers.ts
```diff
@@ -27,32 +27,46 @@ const GQL_TYPESCRIPT_DATA_SCHEMA_TYPE_MAP = {
...
+  if (field.default?.kind === 'DB_GENERATED' && sequenceRegex.test(field.default.value.toString())) {
```
sundersc, line 42: We can possible move the check `sequenceRegex.test(field.default.value.toString()` to a separate method so that it can reused in the future like `isSequenceField(field)`.

packages/amplify-graphql-schema-generator/src/__tests__/__snapshots__/generate.test.ts.snap
p5quared: Whitespace in my template string with the field info csv was causing this--resolved.

packages/amplify-graphql-schema-generator/src/__tests__/pg-string-datasource-adapter.test.ts
```diff
@@ -133,6 +133,11 @@ describe('testPostgresStringDataSourceAdapter', () => {
     expect(adapter.getModels()).toMatchSnapshot();
   });
 
+  it('sets the model field correctly from a schema with a SERIAL field', () => {
+    const adapter = new PostgresStringDataSourceAdapter(schemas.postgres.serial);
+    expect(adapter.getModels()[0].getFields()).toMatchSnapshot();
```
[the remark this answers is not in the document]
p5quared: I modified this to match the other tests, e.g. just `adapter.getModels()`. I think it's clearer and it also retains info that about the PK:
```ts
 "primaryKey": Index {
      "fields": Array [
        "id",
      ],
      "name": "PRIMARY_KEY",
    },
```

sundersc (approved):

packages/amplify-graphql-schema-generator/src/ts-schema-generator/helpers.ts
palpatim: Can we find documentation that this is the canonical way to determine whether a field is `serial`? In the past we've relied on experimentally-verified "rules" that turned out to be incomplete, so I'd love to get ahead of that this time.

packages/amplify-graphql-schema-generator/src/ts-schema-generator/helpers.ts
```diff
@@ -70,14 +83,21 @@ const createDataType = (type: FieldType): ts.Node => {
...
-    ts.factory.createPropertyAccessExpression(createDataType(type.type) as ts.Expression, ts.factory.createIdentifier(modifier)),
+    ts.factory.createPropertyAccessExpression(createDataType(unwrappedField) as ts.Expression, ts.factory.createIdentifier(modifier)),
     undefined,
     undefined,
   );
 };
 
+const isSequenceField = (field: Field): boolean => {
+  const sequenceRegex = /^nextval\(.+::regclass\)$/;
+  return field.default?.kind === 'DB_GENERATED' && sequenceRegex.test(field.default.value.toString());
```
palpatim, line 98: (Copying this from main conversation comment)

Can we find documentation that this is the canonical way to determine whether a field is `serial`? In the past we've relied on experimentally-verified "rules" that turned out to be incomplete, so I'd love to get ahead of that this time.

palpatim (approved):

packages/amplify-graphql-schema-generator/src/ts-schema-generator/helpers.ts
p5quared: In summary we cannot determine if a column was declared `serial`, we _can_ identify sequences though, and from this, infer serial fields.

Ultimately `serial` doesn't really exist as a datatype we can directly introspect. In PG it is basically a macro for _"Create a column as an integer and create a sequence to back that column as a default"_. We end up with an ordinary non-null integer with a special default value.

We can tell if a column is backed by a _sequence_ since its default will be `nextval(_)`, one of the special [sequence functions](https://www.postgresql.org/docs/9.2/functions-sequence.html).

[Reference](https://dba.stackexchange.com/questions/47098/how-do-i-determine-if-a-column-is-defined-as-a-serial-data-type-instead-of-an-in).

packages/amplify-graphql-schema-generator/src/ts-schema-generator/helpers.ts
p5quared: Added main reply with my research/argument later in thread. I don't believe there is a truly canonical way to identify serial but we can identify if a column is backed by a sequence.

packages/amplify-graphql-schema-generator/src/ts-schema-generator/helpers.ts
palpatim: Thanks for the thorough explanation!

PR #2997, opened 2024-11-01
feat(conversation): sorting and performant list queries

atierian:
### 1. List Messages Ordering
The current `listConversationMessages` query uses a GSI to return messages ordered by `createdAt`. That GSI is:
- Name: `gsi-ConversationMessage.conversationId.createdAt`
- PK: `conversationId`
- SK: `createdAt`

The model generated list query does a DynamoDB Scan operation.

These messages are retrieved in ascending order because that is the only option for DynamoDB Scan operations. This is seemingly what we want -- most recent message as the final element. This ordering matches what a user interface would expect and is required by Bedrock's Converse API. It is, however, problematic for long running conversations with messages exceeding the provided `limit` argument. In these cases, the most recent messages are omitted. What we actually want is to retrieve the messages from the DynamoDB table in descending order, then reverse the elements of the array before returning. This ensures that we're including the most recent messages. Using descending order ([`ScanIndexForward`](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_Query.html#DDB-Query-request-ScanIndexForward)` = false`) forces us to use a DynamoDB query operation, which is a good thing -- yay performant reads!

### 2. List Conversations Ordering
The model generated list conversations query currently doesn't use a GSI with sort key for ordering. For clients listing conversations, we'd like the results to be ordered in descending order by the most recently active conversations. This would match expected ordering in user interfaces of customer applications and in Amplify console where customers should be able to view conversation records.

### 1. List Messages Ordering
- Updates `init` slot of model generated `listConversationMessages` query. ([reference](https://github.com/aws-amplify/amplify-category-api/pull/2997/files#diff-3365a7aa9bf7e001d5b54d4918ad0f9a4123957b40f65b93fc199abb91e58688))
  - if a `conversationId: { eq: '<id>' }` expression is included in the arguments, define a query expression to use query rather than scan.
  - sets the `sortDirection` to `'DESC'`.
- Adds `postDataLoad` slot to model generated `listConversationMessages` query to reverse items. ([reference](https://github.com/aws-amplify/amplify-category-api/pull/2997/files#diff-0c7fbf2212fd5de3e7f3358052ca3ea5745c865e6be17c99eae7af5572318384))

### 2. List Conversations Ordering
- Adds GSI to Conversation table for listing conversations in descending order by `updatedAt`. ([reference](https://github.com/aws-amplify/amplify-category-api/pull/2997/files#diff-acac0d969168e1ad99e7b9ac168ca4ab8f0599009b632a23e9608bb59b30c2adR275-R293))
- Adds `init` slot on model generated `listConversations` query: ([reference1](https://github.com/aws-amplify/amplify-category-api/pull/2997/files#diff-9dd6ff278c457b3e4405bc6dda546dee1945b6683d4c2f11c0122eaae538e7cd), [reference2](https://github.com/aws-amplify/amplify-category-api/pull/2997/files#diff-6c1e5869293bf5b8400612a826985166edbb7c018ada36befc01e1d0a5123b52))
  - defines `index` for querying by GSI.
  - defines query expression to use query rather than scan.
  - sets the `sortDirection` to `'DESC'`.
- Updates send message resolver to the `updatedAt` field on the conversation table. ([reference](https://github.com/aws-amplify/amplify-category-api/pull/2997/files#diff-6faba3512ff8d1bc579cd2e38243e7f9ff27fdd32a3f2bc2746732c72375d798))

## CDK / CloudFormation Parameters Changed
N/A

## Issue #, if available
N/A

## Description of how you validated changes
- [E2E test run](https://us-east-1.console.aws.amazon.com/codesuite/codebuild/594813022831/projects/amplify-category-api-e2e-workflow/batch/amplify-category-api-e2e-workflow:8049148f-4c03-4441-b48e-579ec9a108c7?region=us-east-1)

## Checklist
- [x] PR description included
- [x] `yarn test` passes
- [x] E2E test run linked

### DIFF
--- a/packages/amplify-graphql-conversation-transformer/src/resolvers/templates/set-updated-at-conversation-table-fn.template.js
+++ b/packages/amplify-graphql-conversation-transformer/src/resolvers/templates/set-updated-at-conversation-table-fn.template.js
@@ -1,0 +1,31 @@
+import { util } from '@aws-appsync/utils';
+
+export function request(ctx) {
+  const { authFilter } = ctx.stash;
+  const { conversationId } = [[CONVERSATION_ID_PARENT]];
+
+  const filter = JSON.parse(util.transform.toDynamoDBFilterExpression(authFilter));
+
+  return {
+    operation: 'UpdateItem',
+    key: util.dynamodb.toMapValues({ id: conversationId }),
+    update: {
+      expression: 'SET #updatedAt = :updatedAt',
+      expressionValues: util.dynamodb.toMapValues({
+        ':updatedAt': ctx.stash.defaultValues.createdAt,
+      }),
+      expressionNames: {
+        '#updatedAt': 'updatedAt',
+      },
+    },
+    condition: filter
+  };
+}
+
+export function response(ctx) {
+  if (ctx.error) {
+    util.error('Conversation not found', 'ResourceNotFound');
+  }
+
+  return ctx.result;
+}
--- a/packages/amplify-graphql-conversation-transformer/src/transformer-steps/conversation-resolver-generator.ts
+++ b/packages/amplify-graphql-conversation-transformer/src/transformer-steps/conversation-resolver-generator.ts
@@ -267,30 +267,30 @@ export class ConversationResolverGenerator {
     );
   }
 
-    /**
+  /**
    * Sets up the conversation table index
    * @param ctx - The transformer context provider
    * @param directive - The conversation directive configuration
    */
-    private setUpConversationTableIndex(ctx: TransformerContextProvider, directive: ConversationDirectiveConfiguration): void {
-      const conversationName = directive.conversation.model.name.value;
-      const conversation = directive.conversation.model;
+  private setUpConversationTableIndex(ctx: TransformerContextProvider, directive: ConversationDirectiveConfiguration): void {
+    const conversationName = directive.conversation.model.name.value;
+    const conversation = directive.conversation.model;
 
-      const conversationTable = getTable(ctx, conversation);
-      const gsiPartitionKeyName = '__typename';
-      const gsiPartitionKeyType = 'S';
-      const gsiSortKeyName = 'updatedAt';
-      const gsiSortKeyType = 'S';
+    const conversationTable = getTable(ctx, conversation);
+    const gsiPartitionKeyName = '__typename';
+    const gsiPartitionKeyType = 'S';
+    const gsiSortKeyName = 'updatedAt';
+    const gsiSortKeyType = 'S';
 
-      this.addGlobalSecondaryIndex(
-        conversationTable,
-        ctx,
-        conversationName,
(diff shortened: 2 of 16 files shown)


packages/amplify-graphql-conversation-transformer/src/resolvers/templates/set-updated-at-conversation-table-fn.template.js
atierian, line 1: This replaces the existing [verify-session-owner](https://github.com/aws-amplify/amplify-category-api/blob/815d51fcd6fd105614f879271f21a307affe8c9d/packages/amplify-graphql-conversation-transformer/src/resolvers/templates/verify-session-owner-resolver-fn.template.js) resolver function. The `authFilter` being used is unchanged and is now being used as a conditional write expression to ensure ownership.

packages/amplify-graphql-conversation-transformer/src/transformer-steps/conversation-resolver-generator.ts
```diff
@@ -253,6 +267,31 @@ export class ConversationResolverGenerator {
...
+    const conversationName = directive.conversation.model.name.value;
+    const conversation = directive.conversation.model;
+
+    const conversationTable = getTable(ctx, conversation);
+    const gsiPartitionKeyName = '__typename';
+    const gsiPartitionKeyType = 'S';
+    const gsiSortKeyName = 'updatedAt';
+    const gsiSortKeyType = 'S';
+
+    this.addGlobalSecondaryIndex(
```
palpatim, line 285: I would have expected PK=ownerId; SK=updatedAt. What drives the decision to use type name rather than owner?

packages/amplify-graphql-conversation-transformer/src/resolvers/templates/set-updated-at-conversation-table-fn.template.js
palpatim: nice!

packages/amplify-graphql-conversation-transformer/src/transformer-steps/conversation-resolver-generator.ts
atierian: That's also what I was hoping to do here -- it should* work great for queries originating from users (CUP / OIDC access token) because the resolver function knows the `owner` field value before executing the query. The prerequisite of knowing the `owner` field value would make the GSI unusable for admin (IAM) queries though. We could work around this with both GSIs and only using the owner+updatedAt index if identity context provides that info. I would like to do this, but I'm also hesitant to further bifurcate the user owner and admin query flows.

*should being the key word. I've often wondered why we don't automatically create a GSI where PK=ownerId on tables where the model is `owner` auth based. If there's a good reason, it's likely relevant here.

palpatim (approved):

sundersc (approved):

PR #3007, opened 2024-11-06
feat(conversation): require auth input for conversation directive

atierian:
## Related PR
- https://github.com/aws-amplify/amplify-api-next/pull/384

## Problem
The `@conversation` directive implicitly uses owner auth (cognito user pool) for associated models and operations because it's the only supported option. 

In the future, we may want to extend conversation routes to additional authorization strategies. When that time comes, we'll be forced to make a decision between:
1. implicit owner auth by default, opt into new authorization strategies
**--- or ---**
2. breaking change (forcing explicit authorization definition)

## Description of changes

We're not going to support additional authorization strategies by GA, but we are making the decision to go with _2. breaking change (forcing explicit authorization definition)_ already. We're doing this now while we can so that we don't have to make a difficult decision later.

- Adds `auth: ConversationAuth!` argument to `@conversation` directive.

The only supported strategy and provider for now are `owner` and `userPools`:
```gql
input ConversationAuth {
  strategy: ConversationAuthStrategy!
  provider: ConversationAuthProvider!
}
enum ConversationAuthStrategy {
  owner
}
enum ConversationAuthProvider {
  userPools
}
```

#### Why not just use `@auth`?
We could, but that would involve adding a check in the auth transformer that skipped any of it's work if the field also contained a `@conversation` directive. I'm not ok with doing that.

- [E2E test run](https://us-east-1.console.aws.amazon.com/codesuite/codebuild/594813022831/projects/amplify-category-api-e2e-workflow/batch/amplify-category-api-e2e-workflow:1ef0ecd7-5bf0-4c90-9d6e-9d4e9357685c?region=us-east-1#)

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] E2E test run linked

### DIFF
--- a/packages/amplify-graphql-api-construct-tests/src/__tests__/conversations/graphql/schema-conversation.graphql
+++ b/packages/amplify-graphql-api-construct-tests/src/__tests__/conversations/graphql/schema-conversation.graphql
@@ -8,6 +8,7 @@ type Mutation {
     @conversation(
       aiModel: "anthropic.claude-3-haiku-20240307-v1:0"
       systemPrompt: "You are a helpful chatbot that responds in the voice and tone of a pirate. Respond in 20 words or less."
+      auth: { strategy: owner, provider: userPools }
     )
     @aws_cognito_user_pools
 }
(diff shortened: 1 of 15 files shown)


palpatim (approved):
Discussed with @atierian. This seems to be the best of the available options, based on how we have to shuttle auth payloads between clients, AppSync, Lambdas, and tools.

It's a bit unfortunate that we have to have a special-case auth case for this directive, but it's mitigated by the fact that the Gen 2 API acts like a "normal" `authorization` modifier. That makes the GraphQL directive itself an implementation detail of the happy path; and an escape hatch for customers who use CDK rather than Gen 2.

Approved for both implementation and API bar raiser.

dpilch (approved):

PR #3013, opened 2024-11-12
feat(conversation): update directive input for tool definition for model list queries

atierian:
## Related PR
- https://github.com/aws-amplify/amplify-api-next/pull/390

## Problem

1. Customers wanting to use generated model queries as tools for conversation routes must know the name of the generated model query in advance.

```graphql
type Todo @model @auth(rules: [{ allow: owner }]) {
  content: String
  isDone: Boolean
}

type Mutation {
  chat(
    conversationId: ID!,
    content: [ContentBlockInput],
    aiContext: AWSJSON,
    toolConfiguration: ToolConfigurationInput
  ): ConversationMessage
  @conversation(
    aiModel: "anthropic.claude-3-haiku-20240307-v1:0",
    systemPrompt: "You are a helpful chatbot. Answer questions to the best of your ability.",
    tools: [
      {
        name: "listTodos",  # <--- must match query name
        description: "lists todos",
      },
    ],
    auth: { strategy: owner, provider: userPools },
  )
}
```

2. The name of a tool must be the name of the GraphQL query. This is an arbitrary requirement we currently enforce through the directive API that isn't necessary.

## Description of changes

Updates the `ToolMap` input in the `@conversation` directive definition
```diff
  input ToolMap {
-    name: String
-    description: String
+    name: String!
+    description: String!
+    queryName: String
+    modelName: String
+    modelOperation: ConversationToolModelOperation
  }

+  enum ConversationToolModelOperation {
+    list
+  }
```

As referenced in the [inline documentation](https://github.com/aws-amplify/amplify-category-api/pull/3013/files?diff=split&w=0#diff-f57f2ca47a2e1f04b43fff35aa6b420bd0f43cf11cc3c42c392a01a2ab7c17ceR33-R38), this is essentially a fake union type.

### New DX
```graphql
# model generated list query
@conversation(
  aiModel: "anthropic.claude-3-haiku-20240307-v1:0",
  systemPrompt: "You are a helpful chatbot. Answer questions to the best of your ability.",
  tools: [
    {
      name: "list_customers",
      description: "Provides data about the customer sending a message",
      modelName: "Customer",
      modelOperation: list,
    },
  ],
  auth: { strategy: owner, provider: userPools },
)

# custom query
@conversation(
  aiModel: "anthropic.claude-3-haiku-20240307-v1:0",
  systemPrompt: "You are a helpful chatbot. Answer questions to the best of your ability.",
  tools: [
[...]

PR #3037, opened 2024-11-19
fix(conversation): model list tool query name infinite recursion

atierian:
## Problem

The tool definition input for the `@conversation` directive allows customers to specify a `model` and `modelOperation`. For example:

```ts
const schema = a.schema({
  Appointment: a.model({
    time: a.timestamp(),
  }).authorization((allow) => [allow.authenticated()]),

  appointmentChat: a.conversation({
    aiModel: a.ai.model("Claude 3.5 Sonnet"),
    systemPrompt: `You are a friendly pet groomer who is helping with giving advice and selling clients on booking appointments.
    You like grooming pets in creative ways, such as technicolor dye jobs, fun shaped cuts, and nail polish.
    You should ask me more questions to learn about what I need in a grooming appointment.`,
    tools: [
      a.ai.dataTool({
        name: 'getAppointmentTimes',
        description: 'Searches for available appointments',
        model: a.ref('Appointment'),
        modelOperation: 'list',
      }),
    ],
  }).authorization((allow) => allow.owner()),
});
```

Model list query GraphQL inputs are always optional. We currently don't include any of those input types in the JSONSchema tool definition because there's a bug in the implementation that results in a blown stack due to infinite recursion. As a temporary workaround, we do a rudimentary check in the conversation transformer ([reference](https://github.com/aws-amplify/amplify-category-api/blob/b51f7908c86466bc1c666823c863ad38b7e118a6/packages/amplify-graphql-conversation-transformer/src/tools/process-tools.ts#L255-L257)) to ensure they're not included. This check depends, in part, on the name of the query field to determine if it's a model list query. 

https://github.com/aws-amplify/amplify-category-api/pull/3013 introduced a change in the `@conversation` directive API that (intentionally) allows the name of a tool to differ from the name of a query field.

The check is now operating on the tool name rather than the query field name as it should. This is problematic because the name of a model list query tool must start with `"list"` for the check to behave as expected.

Unfortunately this was missed because the test case that would've caught it has a tool named `list_customers`. ([reference](https://github.com/aws-amplify/amplify-category-api/pull/3037/files#diff-d2c0a1f0c50cc44895f2e2c9dfcf85424e98d4715cf79bffd4898635cec0a962L37))

## Description of changes
Determines whether a tool is a model list query tool based on the directive input.

A follow up PR will remove the need for this check all together. It will also add some E2E tests around model list query tools. However that's a large enough behavioral change that it needs to go out separately from this fix.

N/A

- [updated unit test case](https://github.com/aws-amplify/amplify-category-api/pull/3037/files#diff-d2c0a1f0c50cc44895f2e2c9dfcf85424e98d4715cf79bffd4898635cec0a962L37)
- [E2E test run](https://us-east-1.console.aws.amazon.com/codesuite/codebuild/594813022831/projects/amplify-category-api-e2e-workflow/batch/amplify-category-api-e2e-workflow:8756e2b4-f150-4a8f-880f-cb8b7f886b31?region=us-east-1)
- manual testing

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] E2E test run linked

### DIFF
--- a/packages/amplify-graphql-conversation-transformer/src/__tests__/schemas/conversation-route-model-query-tool-with-relationships.graphql
+++ b/packages/amplify-graphql-conversation-transformer/src/__tests__/schemas/conversation-route-model-query-tool-with-relationships.graphql
@@ -34,7 +34,7 @@ type Mutation {
     systemPrompt: "You are a helpful chatbot. Answer questions to the best of your ability.",
     tools: [
       {
-        name: "list_customers",
+        name: "random_name",
         description: "Provides data about the customer sending a message",
         modelName: "Customer",
         modelOperation: list,
--- a/packages/amplify-graphql-conversation-transformer/src/__tests__/__snapshots__/amplify-graphql-conversation-transformer.test.ts.snap
+++ b/packages/amplify-graphql-conversation-transformer/src/__tests__/__snapshots__/amplify-graphql-conversation-transformer.test.ts.snap
@@ -1545,7 +1545,7 @@ export function request(ctx) {
   const clientTools = args.toolConfiguration?.tools?.map((tool) => {
     return { ...tool.toolSpec };
   });
-  const dataTools = [{"name":"list_customers","description":"Provides data about the customer sending a message","inputSchema":{"json":{"type":"object","properties":{},"required":[]}},"graphqlRequestInputDescriptor":{"selectionSet":"items { name email activeCart { products { name price } customerId id createdAt updatedAt owner } orderHistory { items { products { name price } customerId id createdAt updatedAt owner } nextToken } id createdAt updatedAt owner } nextToken","propertyTypes":{},"queryName":"listCustomers"}}];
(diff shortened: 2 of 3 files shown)


packages/amplify-graphql-conversation-transformer/src/__tests__/schemas/conversation-route-model-query-tool-with-relationships.graphql
```diff
@@ -34,7 +34,7 @@ type Mutation {
     systemPrompt: "You are a helpful chatbot. Answer questions to the best of your ability.",
     tools: [
       {
-        name: "list_customers",
+        name: "random_name",
```
atierian, line 37: The name of this tool was changed because it was hiding the fact that we were (incorrectly) using the tool name to determine if a tool definition is a model list query tool. After changing the name, the test fails with the current implementation and works as expected with the changes introduced in this PR.

phani-srikar (approved):

Siqi-Shan (approved):

PR #3084, opened 2024-12-18
Gen1 migration

dpilch:
#### Description of changes

* Add transformer option `enableGen2Migration`.
    * Set dynamodb table data source removal policy to retain when true.
    * Output a mapping of model name to table name when true.
    * This will be exposed as a feature flag to gen 1 customers.

New behavior is not exposed through the construct because migration is not supported from construct to gen 2.

#### Issue #, if available

N/A

* Add transformer option `enableGen2Migration`.
    * Set dynamodb table data source removal policy to retain when true.
    * Output a mapping of model name to table name when true.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-category-api/package.json
+++ b/packages/amplify-category-api/package.json
@@ -103,7 +103,7 @@
     "coverageProvider": "v8",
     "coverageThreshold": {
       "global": {
-        "branches": 67,
+        "branches": 66,
         "functions": 42,
         "lines": 40
       }
--- a/packages/amplify-graphql-model-transformer/src/graphql-model-transformer.ts
+++ b/packages/amplify-graphql-model-transformer/src/graphql-model-transformer.ts
@@ -342,10 +342,9 @@ export class ModelTransformer extends TransformerModelBase implements Transforme
     });
     if (context.transformParameters.enableGen2Migration && context.transformParameters.enableTransformerCfnOutputs) {
       const { scope } = context.stackManager;
-      const env = context.synthParameters.amplifyEnvironmentName;
       new cdk.CfnOutput(cdk.Stack.of(scope), 'DataSourceMappingOutput', {
         value: cdk.Stack.of(scope).toJsonString(dataSourceMapping),
-        description: `Mapping of model name to data source table name for the ${env} environment.`,
+        description: 'Mapping of model name to data source table name.',
         exportName: cdk.Fn.join(':', [cdk.Aws.STACK_NAME, 'DataSourceMappingOutput']),
       });
     }
--- a/packages/amplify-graphql-model-transformer/src/graphql-model-transformer.ts
+++ b/packages/amplify-graphql-model-transformer/src/graphql-model-transformer.ts
@@ -342,9 +342,10 @@ export class ModelTransformer extends TransformerModelBase implements Transforme
     });
     if (context.transformParameters.enableGen2Migration && context.transformParameters.enableTransformerCfnOutputs) {
       const { scope } = context.stackManager;
+      const env = context.synthParameters.amplifyEnvironmentName;
       new cdk.CfnOutput(cdk.Stack.of(scope), 'DataSourceMappingOutput', {
         value: cdk.Stack.of(scope).toJsonString(dataSourceMapping),
-        description: 'Mapping of model name to data source table name.',
+        description: `Mapping of model name to data source table name for the ${env} environment.`,
         exportName: cdk.Fn.join(':', [cdk.Aws.STACK_NAME, 'DataSourceMappingOutput']),
       });
     }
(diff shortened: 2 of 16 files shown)


packages/amplify-graphql-model-transformer/src/graphql-model-transformer.ts
```diff
@@ -323,9 +324,30 @@ export class ModelTransformer extends TransformerModelBase implements Transforme
...
+        ([, datasource]) => datasource.ds.type === 'AMAZON_DYNAMODB',
+      );
+      ddbDatasources.forEach(([modelName, datasource]) => {
+        if (datasource.ds.dynamoDbConfig && !cdk.isResolvableObject(datasource.ds.dynamoDbConfig)) {
+          dataSourceMapping[modelName] = datasource.ds.dynamoDbConfig.tableName;
+        } else {
+          console.warn(
+            `Could not resolve table name for ${modelName}. DataSourceMappingOutput is incomplete. Please manually add ${modelName} to the mapping for your migration.`,
+          );
+        }
```
palpatim, line 340: Does this block mean that manually overridden table names aren't properly handled? In what cases would the warning be triggered? Is a warning sufficient? It's super easy to miss warnings in the output of a gen1 flow. Can we allow the customer to specify a manual configuration for the errored configs, and only throw an error if that manual specification isn't present?

packages/amplify-graphql-model-transformer/src/graphql-model-transformer.ts
```diff
@@ -323,9 +324,30 @@ export class ModelTransformer extends TransformerModelBase implements Transforme
...
+      });
     });
+    if (context.transformParameters.enableGen2Migration && context.transformParameters.enableTransformerCfnOutputs) {
+      const { scope } = context.stackManager;
+      new cdk.CfnOutput(cdk.Stack.of(scope), 'DataSourceMappingOutput', {
+        value: cdk.Stack.of(scope).toJsonString(dataSourceMapping),
+        description: 'Mapping of model name to data source table name.',
```
palpatim, line 347: nit: Is it possible to add the environment name to this description? (I'm not sure where we get that from, so it's a nice to have, not a must have)

packages/amplify-graphql-model-transformer/src/graphql-model-transformer.ts
dpilch: This was to make TypeScript happy. I haven't found a case where the else would actually execute.

```typescript
Property 'tableName' does not exist on type 'DynamoDBConfigProperty | IResolvable'.
  Property 'tableName' does not exist on type 'IResolvable'.
```

sundersc: How do we handle the cases where table name is changed using `override api`?

dpilch: This will correctly get the table name when using overrides. I just tested again to be sure.

```graphql
type Todo @model {
  id: ID!
  name: String!
  description: String
}
```

Deploy without override adds the correct mapping with the generated table name.

Then I added this override.

```typescript
// set custom table name
resources.models.Todo.modelDDBTable.addPropertyOverride(
  "TableName",
  "MyCustomTableName"
);
```

The table mapping output: `{"Todo":"MyCustomTableName"}`

packages/amplify-graphql-model-transformer/src/graphql-model-transformer.ts
```diff
@@ -323,9 +324,31 @@ export class ModelTransformer extends TransformerModelBase implements Transforme
...
+        if (datasource.ds.dynamoDbConfig && !cdk.isResolvableObject(datasource.ds.dynamoDbConfig)) {
+          dataSourceMapping[modelName] = datasource.ds.dynamoDbConfig.tableName;
+        } else {
+          console.warn(
+            `Could not resolve table name for ${modelName}. DataSourceMappingOutput is incomplete. Please manually add ${modelName} to the mapping for your migration.`,
+          );
+        }
+      });
     });
+    if (context.transformParameters.enableGen2Migration && context.transformParameters.enableTransformerCfnOutputs) {
```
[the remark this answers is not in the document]
dpilch: Good idea, I've added a check in `transformGraphQLSchemaV1` to prevent this.

packages/amplify-category-api/package.json
```diff
@@ -103,7 +103,7 @@
     "coverageProvider": "v8",
     "coverageThreshold": {
       "global": {
-        "branches": 67,
+        "branches": 66,
```
dpilch, line 106: Coverage decreased slightly because of the unreachable error message.

sundersc (approved):

palpatim (approved):

PR #3125, opened 2025-01-22
Add checks for parsing validate directive

bobbyu99:
#### Description of changes

- updated `@validate` to be repeatable in both definition and snapshot test
- added `validators.ts` that checks for invalid use of validate directive, including:
  - the value for `minLength` `maxLength` is non-negative integer
  - specified validation type is compatible with the field type (string validation type for string field, numeric validation type for numeric field)
  - each validation type can only be used once on the same field
- renamed `__tests__` correctly
- updated `jest.config.js` to ignore `types.ts` which has no executable codes, and to use globally defined coverage threshold

- unit tests for parsing directive inputs in `directive-input-parsing.test.ts`
  - length value passed into `minLength` `maxLength` is valid non-negative integer
  - numeric value passed into numeric validations is valid number
- unit tests for invalid directive use in `invalid-directiev-use.test.ts`
  - disallow validation on non-model type and allow on model type
  - disallow validation on list fields
  - disallow duplicate validation types on the same field
  - validation type is applied to compatible field type

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [ ] E2E test run linked


packages/amplify-graphql-validate-transformer/src/types.ts
```diff
@@ -0,0 +1,26 @@
+/* c8 ignore start */
+import { ObjectTypeDefinitionNode, FieldDefinitionNode } from 'graphql';
+/* c8 ignore end */
+
+// All supported validation types
```
palpatim, line 5: Here and throughout: since you're exporting these symbols, use JSDoc style comments (`/** ... */`) so they appear in IDE documentation

packages/amplify-graphql-validate-transformer/src/validators.ts
```diff
@@ -0,0 +1,98 @@
...
+
+const isLengthValidation = (type: ValidationType): boolean => {
+  return ['minLength', 'maxLength'].includes(type);
+};
+
+/**
+ * Validates that length validation values (minLength, maxLength) are valid non-negative integers.
+ */
+const validateLengthValue = (config: ValidateDirectiveConfiguration): void => {
+  const value = parseInt(config.value, 10);
```
palpatim, line 25: Recommend we use the `Number()` constructor rather than `parseInt`, since I assume we want to catch invalid constructions like `"10z.0"`. `parseInt("10z.0", 10)` returns `10`, but `Number("10z.0")` returns `NaN`, which I think is what we want?

packages/amplify-graphql-validate-transformer/src/__tests__/numeric-validation-types.test.ts
```diff
@@ -0,0 +1,116 @@
...
+  describe('Valid usage', () => {
+    test.each([
+      {
+        name: 'accepts -Infinity value',
+        schema: /* GraphQL */ `
+          type Post @model {
+            id: ID!
+            rating: Float! @validate(type: gt, value: "-Infinity")
+            score: Int! @validate(type: lt, value: "-Infinity")
+            counts: [Float]! @validate(type: gte, value: "-Infinity")
```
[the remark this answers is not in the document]
bobbyu99: List validation actually did not come up in either our design or API review. Let me chat with Josef to see his take on whether we want to support it.

packages/amplify-graphql-validate-transformer/src/__tests__/minmax-length.test.ts
```diff
@@ -0,0 +1,188 @@
...
+        schema: /* GraphQL */ `
+          type Post @model {
+            id: ID!
+            title: String! @validate(type: minLength, value: "0")
+            content: String! @validate(type: maxLength, value: "0")
+          }
+        `,
+      },
+      {
+        name: 'accepts length values of extremely large numbers beyond 64-bit range',
```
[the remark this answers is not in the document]
bobbyu99: Let me check what the maximum/minimum value our resolver can accept

packages/amplify-graphql-validate-transformer/src/__tests__/numeric-validation-types.test.ts
```diff
@@ -0,0 +1,116 @@
...
+          transformers: [new ModelTransformer(), transformer],
+        });
+      }).toThrow(error);
+    });
+  });
+
+  describe('Valid usage', () => {
+    test.each([
+      {
+        name: 'accepts -Infinity value',
```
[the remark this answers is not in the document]
bobbyu99: I will also double check if Infinity is accepted in resolver

packages/amplify-graphql-validate-transformer/src/__tests__/duplicate-validation-types.test.ts
```diff
@@ -0,0 +1,138 @@
...
+          }
+        `,
+      },
+    ])('$name', ({ schema }) => {
+      const out = testTransform({
+        schema,
+        transformers: [new ModelTransformer(), new ValidateTransformer()],
+      });
+      expect(out).toBeDefined();
+      expect(out.schema).toMatchSnapshot();
```
sundersc, line 135: What is the purpose of checking snapshot here? If we are looking for specific input types, can we narrow down to the specific type?

packages/amplify-graphql-validate-transformer/src/__tests__/minmax-length.test.ts
```diff
@@ -0,0 +1,188 @@
...
+          }
+        `,
+      },
+    ])('$name', ({ schema }) => {
+      const out = testTransform({
+        schema,
+        transformers: [new ModelTransformer(), new ValidateTransformer()],
+      });
+      expect(out).toBeDefined();
+      expect(out.schema).toMatchSnapshot();
```
sundersc, line 185: Same as above. Consider checking specific types. Comparing full snapshot for too many cases would significantly increase the line counts in snapshot files.

packages/amplify-graphql-validate-transformer/src/__tests__/numeric-validation-types.test.ts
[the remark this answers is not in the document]
bobbyu99: Decided to descope validation on list for now.

packages/amplify-graphql-validate-transformer/src/__tests__/duplicate-validation-types.test.ts
bobbyu99: Thanks for pointing this out. I was confused about when we should applied snapshot test. But it is clear to me now that snapshot should be used for testing schema generation. These test cases are for input parsing, so snapshot test is unnecessary.

packages/amplify-graphql-validate-transformer/src/validators.ts
bobbyu99: I noticed that in default transformer, we do the following ([link](https://github.com/aws-amplify/amplify-category-api/blob/d96f49a896b6130596c863ae99ac0ebcb86a9bba/packages/amplify-graphql-default-value-transformer/src/validators.ts#L70-L71))
```
const validateInt = (x: string): boolean => !Number.isNaN(parseInt(x, 10));
const validateFloat = (x: string): boolean => !Number.isNaN(parseFloat(x));
```

This is currently the only place where we parse `String` to `Number`. I suggest that we keep it consistent, so that it will not be a surprise to the customers that the parsing behavior of `@validate` is different from that of `@default`.

packages/amplify-graphql-validate-transformer/src/__tests__/numeric-validation-types.test.ts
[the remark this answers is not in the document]
bobbyu99: This falls under a rare use case. We will use `isNaN(parseFloat(x))` until a customer complains about is.

packages/amplify-graphql-validate-transformer/src/__tests__/minmax-length.test.ts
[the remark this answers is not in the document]
bobbyu99: This falls under a rare use case. We will use `isNaN(parseInt(x))` until a customer complains about is.

bobbyu99:
Accidentally triggerred an e2e run. But no e2e tests have been added for validate directive yet.

packages/amplify-graphql-validate-transformer/src/validators.ts
```diff
@@ -0,0 +1,155 @@
...
+  }
+};
+
+/**
+ * Validates that the field type is compatible with the validation type.
+ * @param fieldNode - The field definition node that the directive is applied to
+ * @param validationType - The validation type to validate against
+ */
+const validateTypeCompatibility = (fieldNode: FieldDefinitionNode, validationType: ValidationType): void => {
+  const baseType = getBaseType(fieldNode.type);
```
sundersc, line 89: What happens if the field is an array? It looks we are considering only the base type. IIRC, even for `[Int]` or `[Int!]!`, the base type will be `Int`. Is that's how the validation is expected to work here?

packages/amplify-graphql-validate-transformer/src/__tests__/invalid-directiev-use.test.ts
palpatim, line 1: nit for future PR: Rename this file to correct spelling of `directive`

palpatim (approved):
Approved with one spelling correction for a future PR

packages/amplify-graphql-validate-transformer/src/validators.ts
bobbyu99: Yes, we check that the field is not a list in one of the schema validation. Here is the order of validations: https://github.com/aws-amplify/amplify-category-api/pull/3125/files/2abf979bdeffba923a365e62d952c10e408301fb#diff-b3bd9c4ff73cfb587b969275f5369600d0710ef02a261ed072b2856b03a8eb65R26-R29. 
```
  validateModelType(parentNode);
  validateNoListFieldValidation(fieldNode);
  validateNoDuplicateTypes(fieldNode, directive, config.type as ValidationType);
  validateTypeCompatibility(fieldNode, config.type as ValidationType);
```

packages/amplify-graphql-validate-transformer/src/__tests__/invalid-directiev-use.test.ts
bobbyu99: thanks for the catch!

sundersc (approved):

PR #3126, opened 2025-01-24
fix(gen1): transform auth on fields of supported extended types

palpatim:
#### Description of changes

Transform `@auth`, `@http`, and `@function` directives on fields of supported extended types: `Query`, `Subscription`, `Mutation`.

Relevant API changes are to the `TransformerPluginProvider` API, where I added a new `fieldOfExtendedType` API. Functionally, this is the same as adding support for `ObjectTypeExtensionNode` to the existing `field` argument, but doing so would break any customers who implemented a custom transformer plugin. Adding the separate optional method allows existing customers to use as-is. 

That API change is mirrored in the provider implementations for the `@auth`, `@function`, and `@http` transformer plugins.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] [E2E test run](https://us-east-1.console.aws.amazon.com/codesuite/codebuild/594813022831/projects/amplify-category-api-e2e-workflow/batch/amplify-category-api-e2e-workflow:fded2b56-e02e-467e-83ce-ed361fb724a4?region=us-east-1) passes
  - E2E passes locally. Tagged release validates all behaviors.


packages/amplify-graphql-http-transformer/src/graphql-http-transformer.ts
```diff
@@ -182,6 +182,15 @@ export class HttpTransformer extends TransformerPluginBase {
     this.directiveList.push(args);
   };
 
+  fieldOfExtendedType = (
+    parent: ObjectTypeDefinitionNode | InterfaceTypeDefinitionNode | ObjectTypeExtensionNode,
```
sundersc, line 186: Don't we call this method only for `ObjectTypeExtensionNode` type from transformers? Can we get rid of `ObjectTypeDefinitionNode | InterfaceTypeDefinitionNode` here?

palpatim: D'oh! You are correct, we do. The `@function` and `@auth` transformers have the correct signature -- I forgot to remove the additional types from the `@http` transformer.

sundersc (approved):

Siqi-Shan (approved):

PR #3134, opened 2025-01-29
feat: add resolvers generation

bobbyu99:
#### Description of changes
- added `generateResolvers` method that generates resolver functions in CREATE and UPDATE resolver pipeline
- added `makeValidationSnippet` and `getValidationCheck` helpers for VTL code generation
- updated CREATE and UPDATE resolver pipeline in `transformer-core/src/transformer-context/resolver.ts`

- unit tests for `makeValidationSnippet` and `getValidationCheck` helpers
- unit tests for `generateResolvers`

#### Checklist
- [ ] PR description included
- [ ] `yarn test` passes
- [ ] E2E test run linked

### DIFF
--- a/packages/amplify-graphql-validate-transformer/src/vtl-generator.ts
+++ b/packages/amplify-graphql-validate-transformer/src/vtl-generator.ts
@@ -1,0 +1,49 @@
+import { printBlock, raw } from 'graphql-mapping-template';
+
+/**
+ * Generates a VTL snippet for field validation.
+ * @param fieldName - The name of the field to validate
+ * @param validationType - The type of validation to perform
+ * @param validationValue - The value to validate against
+ * @param errorMessage - The error message to display if validation fails
+ * @returns A VTL code block that performs the validation check and throws an error if validation fails
+ */
+export const makeValidationSnippet = (fieldName: string, validationType: string, validationValue: string, errorMessage: string): string => {
+  const validationCheck = getValidationCheck(fieldName, validationType, validationValue);
+
+  return printBlock(`Validating "${fieldName}" with type "${validationType}" and value "${validationValue}"`)(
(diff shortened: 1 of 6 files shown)


packages/amplify-graphql-validate-transformer/src/vtl-generator.ts
```diff
@@ -0,0 +1,49 @@
...
+    #end`),
+  );
+};
+
+const getValidationCheck = (fieldName: string, validationType: string, value: string): string => {
+  const fieldRef = `$ctx.args.input.${fieldName}`;
+
+  switch (validationType.toLowerCase()) {
+    case 'gt':
+      return `#set($validationFailed = $${fieldRef} <= ${value})`;
```
palpatim, line 29: Here and throughout this file: Why are we evaluating negative assertions rather than positive ones? E.g., why do we define `lt` as being "not greater than or equal to" as opposed to just "less than"?

packages/graphql-mapping-template/api-extractor.json
palpatim, line 1: As above, whence this file?

packages/amplify-graphql-directives/api-extractor.json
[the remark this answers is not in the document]
bobbyu99: Looks like I accidentally commit the temporary files from running `yarn extract-api`

packages/amplify-graphql-validate-transformer/src/vtl-generator.ts
bobbyu99: I didn't think about why negative assertion when implementing this. But after thinking it through, I am leaning towards positive assertion as it explicitly states what we expect to be true.

palpatim (approved):

PR #3148, opened 2025-02-05
chore: bump cdk version to 2.177.0

Siqi-Shan:
#### Description of changes

Bumps `aws-cdk` and `aws-cdk-lib` dependencies versions to `2.177.0`, and feature flag `@aws-cdk/aws-iam:oidcRejectUnauthorizedConnections` set to true, refer to https://github.com/aws/aws-cdk/issues/32920

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [ ] E2E test run linked


packages/amplify-graphql-api-construct/src/amplify-graphql-api.ts
```diff
@@ -146,6 +146,10 @@ export class AmplifyGraphqlApi extends Construct {
     super(scope, id);
     this.stack = Stack.of(scope);
 
+    if (!FeatureFlags.of(this).isEnabled('@aws-cdk/aws-iam:oidcRejectUnauthorizedConnections')) {
+      this.node.setContext('@aws-cdk/aws-iam:oidcRejectUnauthorizedConnections', true);
+    }
```
sobolk, line 151: is it possible to detect that the flag was explicitly set to `false` and honor that? Default to `true` if it's `undefined` ?

We might want to let customers opt out from this somehow?

Siqi-Shan: Thanks for suggestions, got some thinkings around this for customer provided `cdk.json` and `FeatureFlags.of(this).isEnabled(...)`:
- no feature flag is presented: return `false`
- feature flag is set to `true`: return `true`
- feature flag is set to `false`: will return `false`
- i.e. `FeatureFlags.of(this).isEnabled(...)` will not distinguish if the feature flag is not provided or set to `false`

The way I can think to achieve this is to use `construct.node.tryGetContext`:
- the construct will try to get context value first (since feature context is just context value), and if no feature flag is presented there, `undefined` will be returned, while the construct will set feature flag to default `true`
- if the feature flag is presented, we continue with `FeatureFlags.of(this).isEnabled(...)` and honor `false` value there

I'm not sure if this is the recommended approach, as [CDK notes](https://github.com/aws-amplify/amplify-category-api/security/dependabot/162) mentioned "should not disallow this", but I personally think honor explicit `false` is the right path here

sobolk: > I'm not sure if this is the recommended approach

It is not recommended. But in case somebody relies on that behavior they may want a way to keep old behavior.

The default (setting this to true if not set at all) should follow the recommendation.
Opt-out (setting explicitly to false) is a way for customer to say that they consciously choose to live dangerously. Presumably with a reason.

Siqi-Shan: Sorry for confusions in prev comment. I double checked the API provided by `FeatureFlags` and did some experiments with a sample app. Turns out the API will return `undefined` if a feature flag is not presented in `cdk.json`, and that effectively provided us a way to honor the provided value (if feature flag is presented).

```typescript
isEnabled(featureFlag: string): boolean | undefined;
```

I agree to preserve the `opt-out` option here to avoid possible regressions in customer's usage, will have this revised

palpatim (approved):
LGTM, contingent on verification that our feature flag processing works as expected

PR #3150, opened 2025-02-06
test: evaluate mapping template e2e

bobbyu99:
#### Description of changes

This PR adds evaluate template tests (using `aws appsync evaluate-mapping-template`), targeting `makeValidationSnippet` specifically to make sure handling of input validation is correct and that the generated logic performs as advertised. There will be another PR for more e2e tests with actual queries. 

Changes:
- [`validate-transformer-evaluate-mapping.test.ts`](https://github.com/aws-amplify/amplify-category-api/pull/3150/files#diff-bf58adf16697b3ef40e85f2d1bae852e1f284de0cc1ddfff063da4d0f8599101)
  - Error message and string validation threshold parsing to escape quotes properly (see below for more details)
  - Numeric validations (like large positive/negative values and values slightly above/below thresholds)
  - String validations (like special characters, unicode characters, and regex patterns)
- [`vtl-generator.ts`](https://github.com/aws-amplify/amplify-category-api/pull/3150/files#diff-6a7a4947c26116510c5dc1a6a160dbad7e9faf73e063947db0aa898d17769ed5)
  - In https://github.com/aws-amplify/amplify-category-api/pull/3149#issue-2833754243, we escape single quote in custom error message such that VTL function outputs the error message exactly as it's provided by the customer. I determined that it is necessary to also escape single quote in string validation threshold (the value to validate against for `startsWith`, `endsWith`, `matches`) such that validation is performed for exactly the the value provided by the customer. 
  - Note that this does not have to do with Java vs. JS regex differences, but purely to avoid malformed VTL string syntax as I am wrapping the `value` in single quotes within the VTL snippet—like `#set($validationPassed = $util.matches('${value}', ${fieldRef}))`
  - Snapshots updated for using single quotes
- Export `makeValidationSnippet` from validate transformer so we can test it in e2e test
- Add a few helper functions to run evaluate template tests

[E2E](https://tiny.amazon.com/csvkcwqk/IsenLink)

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] E2E test run linked


packages/amplify-graphql-api-construct-tests/src/utils/validate-transformer-helper.ts
```diff
@@ -0,0 +1,205 @@
...
+/**
+ * Evaluates a mapping template with the given context
+ * @param templateName The name of the template to evaluate
+ * @param contextName The name of the context to use
+ * @param directory The directory containing the template files
+ * @returns The result of the evaluation
+ */
+export const evaluateTemplate = async (templateName: string, contextName: string, directory: string): Promise<any> => {
+  const { stdout, stderr } = await execAsync(
+    `aws appsync evaluate-mapping-template --template file://${templateName} --context file://${contextName}`,
```
palpatim, line 162: Several problems with this pattern:
1. `execAsync` spawns a shell process. If the values contained in `templateName` or `contextName` are [unsanitized](https://nodejs.org/api/child_process.html#child_processexeccommand-options-callback:~:text=COPY-,Never%20pass%20unsanitized%20user%20input%20to%20this%20function,-.%20Any%20input%20containing), then they can execute processes with the permissions of the test user. If we decide to `exec`, use the [`execFile`](https://nodejs.org/api/child_process.html#child_processexecfilefile-args-options-callback) flavor, which doesn't spawn a shell. However...
2. ...even better, use the AppSync SDK, not the command line. That is our preferred pattern throughout the tests (with the exception of things like `cdk` and `amplify` CLI operations), and avoids process spawns and the associated IPC oddness.

palpatim (changes requested):

palpatim (dismissed):

palpatim (approved):

Siqi-Shan (approved):

PR #3197, opened 2025-02-28
fix: use aws owned key as default for ddb sse

sundersc:
#### Description of changes

Change the DynamoDB tables SSE config to use the AWS owned key instead of AWS managed key by default.

**Reason:** AWS managed keys incur additional charges and it is not the best option to be the default.

> **Note:** When SSE is undefined, it doesn't mean SSE is disabled, it means DynamoDB will use AWS owned key.

NA

#### Issue #, if available
NA

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] E2E test run linked

### DIFF
--- a/packages/amplify-graphql-api-construct-tests/src/__tests__/backends/amplify-table/simple-todo/updateIndex/app.ts
+++ b/packages/amplify-graphql-api-construct-tests/src/__tests__/backends/amplify-table/simple-todo/updateIndex/app.ts
@@ -49,7 +49,7 @@ todoTable.setGlobalSecondaryIndexProvisionedThroughput('byName2', {
   writeCapacityUnits: 4,
 });
 todoTable.pointInTimeRecoveryEnabled = true;
-todoTable.sseSpecification = { sseEnabled: false };
+todoTable.sseSpecification = { sseEnabled: true };
 todoTable.streamSpecification = { streamViewType: StreamViewType.KEYS_ONLY };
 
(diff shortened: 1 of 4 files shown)


packages/amplify-graphql-api-construct-tests/src/__tests__/backends/amplify-table/simple-todo/updateIndex/app.ts
```diff
@@ -49,7 +49,7 @@ todoTable.setGlobalSecondaryIndexProvisionedThroughput('byName2', {
   writeCapacityUnits: 4,
 });
 todoTable.pointInTimeRecoveryEnabled = true;
-todoTable.sseSpecification = { sseEnabled: false };
+todoTable.sseSpecification = { sseEnabled: true };
```
palpatim, line 52: Please add a comment--at least a doc callout--for future developers who might get confused by the API here.

palpatim (commented):
Doc comment request but functionally LGTM

packages/amplify-graphql-api-construct-tests/src/__tests__/backends/amplify-table/simple-todo/updateIndex/app.ts
sundersc: Added the comment to `amplify-table-1.test.ts` file.

palpatim (approved):

dpilch (approved):

PR #3232, opened 2025-03-27
feat: upgrade lambda runtimes to node 22

dpilch:
#### Description of changes

Upgrade lambda runtimes to node 22.

The transformers cannot be upgraded to a new major version because the next major version is already in use in gen 2. `@aws-amplify/amplify-category-api` is set to upgrade to the next major version. This is the entry point for gen 1 and is not used in gen 2.

Lambda runtimes set to Node 22.

#### Issue #, if available

N/A

* Unit tests
* E2E tests: https://us-east-1.console.aws.amazon.com/codesuite/codebuild/594813022831/projects/amplify-category-api-e2e-workflow/batch/amplify-category-api-e2e-workflow:b5962487-3363-4fb4-b499-0899b2e3b315?region=us-east-1

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] E2E test run linked


packages/amplify-category-api/src/provider-utils/awscloudformation/pipeline-with-awaiter.ts
```diff
@@ -26,7 +26,7 @@ export type GitHubSourceActionInfo = {
   tokenSecretArn: string;
 };
 
-const lambdaRuntimeNodeVersion = lambda.Runtime.NODEJS_18_X;
+const lambdaRuntimeNodeVersion = new lambda.Runtime('nodejs22.x', lambda.RuntimeFamily.NODEJS, { supportsInlineCode: true });
```
dpilch, line 29: `NODEJS_22_X` static member is not available. Define the runtime through the class constructor as is recommended in the docs.

sobolk (approved):

sobolk:
@dpilch by looking at changes - it doesn't seem category api nor transformers need major version.

If I understand correctly:
1. Lambdas will be recreated - they're internal part of our implementation. No action is required from customer.
2. The "container" templates are used to generate new resources added to a project, i.e. no impact for existing project?

The major version on CLI side was driven by the fact that we use Node to run scripts like overrides, custom resources or hooks. Lambda changes there were for completeness but they themselves did not warrant MV.

dpilch:
1. Yes
2. Yes

Makes sense, we can keep on the same major version then.

Issue #2929, opened 2024-10-07
IAM not authorized for schema with only custom operations


palpatim:
### Environment information

```plain text
System:
  OS: macOS 15.0.1
  CPU: (10) arm64 Apple M1 Pro
  Memory: 121.84 MB / 32.00 GB
  Shell: /bin/zsh
Binaries:
  Node: 18.20.4 - ~/.local/share/mise/installs/node/18/bin/node
  Yarn: 1.22.22 - ~/.local/share/mise/installs/node/18/bin/yarn
  npm: 10.7.0 - ~/.local/share/mise/installs/node/18/bin/npm
  pnpm: undefined - undefined
NPM Packages:
  @aws-amplify/auth-construct: 1.3.1
  @aws-amplify/backend: 1.3.0
  @aws-amplify/backend-auth: 1.2.0
  @aws-amplify/backend-cli: 1.2.8
  @aws-amplify/backend-data: 1.1.4
  @aws-amplify/backend-deployer: 1.1.4
  @aws-amplify/backend-function: 1.5.0
  @aws-amplify/backend-output-schemas: 1.2.0
  @aws-amplify/backend-output-storage: 1.1.2
  @aws-amplify/backend-secret: 1.1.3
  @aws-amplify/backend-storage: 1.2.0
  @aws-amplify/cli-core: 1.1.3
  @aws-amplify/client-config: 1.3.2
  @aws-amplify/deployed-backend-client: 1.4.1
  @aws-amplify/form-generator: 1.0.3
  @aws-amplify/model-generator: 1.0.8
  @aws-amplify/platform-core: 1.1.0
  @aws-amplify/plugin-types: 1.3.0
  @aws-amplify/sandbox: 1.2.2
  @aws-amplify/schema-generator: 1.2.4
  aws-amplify: 6.6.2
  aws-cdk: 2.160.0
  aws-cdk-lib: 2.160.0
  typescript: 5.6.2
AWS environment variables:
  AWS_PROFILE = personal
  AWS_DEFAULT_REGION = us-west-2
  AWS_SCHMELTE_ACCOUNT = 779656175277
  AWS_STS_REGIONAL_ENDPOINTS = regional
  AWS_NODEJS_CONNECTION_REUSE_ENABLED = 1
  AWS_SDK_LOAD_CONFIG = 1
CDK environment variables:
  CDK_DEFAULT_ACCOUNT = 779656175277
  CDK_DEFAULT_REGION = us-west-2
```

[...]

PR #3270, opened 2025-04-23
fix: add missing aws_iam directive on custom types

svidgen:
#### Description of changes

Adds `@aws_iam` directive to non-model types to support IAM auth from functions and other AWS services.

#### Issue #, if available

1. https://github.com/aws-amplify/amplify-category-api/issues/2929

1. Updated tests
2. Red/Green access test on [EventBridge test app](https://github.com/svidgen/amplify-gen2-node-playground/tree/event-bridge-testing). (Removed the dummy model workaround, observed failure, updated to local build including fix, observed success.)
3. Confirmed async function also still works as expected, per concerns from the linked issue.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] E2E test run linked ([here](https://tiny.amazon.com/igtrhb10/IsenLink))


svidgen:
E2E's and PR workflow are still failing in early stages. This has notably been occurring on prior PR's as well and warrants some investigation, which I'll perform prior to continuing with this PR.

packages/amplify-graphql-auth-transformer/API.md
```diff
@@ -117,6 +117,8 @@ export class AuthTransformer extends TransformerAuthBase implements TransformerA
     // (undocumented)
     addFieldsToObject: (ctx: TransformerTransformSchemaStepContextProvider, modelName: string, ownerFields: Array<string>) => void;
     // (undocumented)
+    addIamAuthToCustomTypes: (ctx: TransformerTransformSchemaStepContextProvider) => void;
+    // (undocumented)
```
sobolk, line 121: It seems this could be private.
(on the other hand, we never paid a lot of attention to transformers public apis...).

sobolk (approved):

packages/amplify-graphql-auth-transformer/API.md
svidgen: Yeah. I haven't formed a strong opinion here yet. I'm just following the established pattern for now. 🤷

sobolk: I would recommend adopting "if something can be private it should be". that creates less liability going forward. (we've been doing this for a while on backend side).
For transformers it would be more to form a habit (or maybe stop the bleeding) rather than fix them. Their apis are already overexposed.

sobolk: > Their apis are already overexposed

This might be a sign that all these transformers packages could have been directories in single package.

svidgen: I might to agree. And, it all seems pretty chaotic to me at the moment. But, there's also a ton of *intrinsic* complexity in the problem space. So, I'm attempting to respect what came before and keep as many existing fences erected until I understand them better and have more solid opinions.

Amplifiyer (approved):

PR #3389, opened 2025-12-30
chore: add a timeout on the entire CLI invocation

rix0rrr:
In ca-central-1 our canary is taking more than an hour to finih. It's not producing no output for 5 minutes because the "no output timeout" would have caught that.

Add another "total duration timeout", set to 30 minutes, to see what it's doing as it's timing out.


rix0rrr:
I'm YOLOing it.

PR #3395, opened 2026-01-13
fix: api canary workflow

svidgen:
#### Description of changes

This workflow is failing in CI: https://tiny.amazon.com/4yawr9g3/IsenLink

When run *locally*, I get TS errors (don't seem to trigger overall failure) and a `jest` process that gets launched in *watch* mode (never exits). Unclear to me from the logging in CB whether this is the real issue here, but this change causes the process to exit successfully on my machine.

Hoping it does the same in CB.

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes
- [ ] E2E test run linked


svidgen:
⏳ - [PR Checks](https://tiny.amazon.com/15x5pjlwm/IsenLink)
⏳ - [E2E Tests](https://tiny.amazon.com/an46dmgm/IsenLink)

client-test-apps/js/api-model-relationship-app/package.json
```diff
@@ -48,7 +48,7 @@
     "test:execute": "cross-env TEST_HARNESS_STAGE=execute yarn test",
     "test:watch": "cross-env TEST_HARNESS_STAGE=watch yarn test",
     "test:teardown": "cross-env TEST_HARNESS_STAGE=teardown yarn test",
-    "test:ci": "yarn test --testResultsProcessor=\"jest-junit\""
+    "test:ci": "CI=true yarn test --testResultsProcessor=\"jest-junit\""
```
iliapolo: This makes sense, and looks like exactly what was missing to correctly simulate the CB run locally. But then do we really need `watchForFileChanges: false` you made above?

Issue #3409, opened 2026-02-10
Remove unnecessary VPC endpoints created with SQL schemas


iliapolo:
### Environment information

```plain text
System:
  OS: macOS 26.2
  CPU: (12) arm64 Apple M3 Pro
  Memory: 2.37 GB / 36.00 GB
  Shell: /bin/zsh
Binaries:
  Node: 22.16.0 - /Users/epolon/.nvm/versions/node/v22.16.0/bin/node
  Yarn: 1.22.22 - /Users/epolon/.nvm/versions/node/v22.16.0/bin/yarn
  npm: 10.9.2 - /Users/epolon/.nvm/versions/node/v22.16.0/bin/npm
  pnpm: 10.29.2 - /Users/epolon/.nvm/versions/node/v22.16.0/bin/pnpm
NPM Packages:
  @aws-amplify/auth-construct: 1.11.0
  @aws-amplify/backend: 1.20.0
  @aws-amplify/backend-ai: Not Found
  @aws-amplify/backend-auth: 1.9.1
  @aws-amplify/backend-cli: 1.8.2
  @aws-amplify/backend-data: 1.6.3
  @aws-amplify/backend-deployer: 2.1.5
  @aws-amplify/backend-function: 1.16.0
  @aws-amplify/backend-output-schemas: 1.8.0
  @aws-amplify/backend-output-storage: 1.3.3
  @aws-amplify/backend-secret: 1.4.2
  @aws-amplify/backend-storage: 1.4.3
  @aws-amplify/cli-core: 2.2.3
  @aws-amplify/client-config: 1.10.0
  @aws-amplify/data-construct: 1.17.0
  @aws-amplify/data-schema: 1.23.0
  @aws-amplify/deployed-backend-client: 1.8.1
  @aws-amplify/form-generator: 1.2.6
  @aws-amplify/model-generator: 1.2.2
  @aws-amplify/platform-core: 1.10.4
  @aws-amplify/plugin-types: 1.11.2
  @aws-amplify/sandbox: 2.1.4
  @aws-amplify/schema-generator: 1.4.1
  @aws-cdk/toolkit-lib: 1.6.1
  aws-amplify: 6.16.2
  aws-cdk-lib: 2.234.1
  typescript: 5.9.3
AWS environment variables:
No CDK environment variables
```

### Data packages

```plain text
amplify-rds-gen2@1.0.0 /Users/epolon/dev/src/local/amplify-rds-gen2
[...]

PR #3449, opened 2026-03-26
feat: add opt-in minimizeRdsVpcEndpoints to the SQL data source strategy

Simone319:
#### Description of changes

Adds an opt-in **`minimizeRdsVpcEndpoints`** field to the SQL data source strategy (`SQLLambdaModelDataSourceStrategy`).

When a SQL (RDS) data source is installed into a VPC, the transformer provisions a set of interface VPC endpoints so the SQL Lambda can reach AWS services. Historically it always provisioned **five** endpoints — `ssm`, `ssmmessages`, `ec2`, `ec2messages`, `kms` — but at runtime the SQL Lambda only consumes the **`ssm`** endpoint (to read the database connection secret). The other four are unused and incur ongoing cost. `minimizeRdsVpcEndpoints` lets a customer opt into provisioning only the `ssm` endpoint:

| `minimizeRdsVpcEndpoints` | Interface VPC endpoints provisioned |
| --- | --- |
| `false` (default) | `ssm`, `ssmmessages`, `ec2`, `ec2messages`, `kms` (5) |
| `true` | `ssm` (1) |

The field lives next to `vpcConfiguration` on the SQL strategy and only takes effect when `vpcConfiguration` is set; it is a no-op for SQL data sources that are not installed into a VPC.

```ts
const strategy: SQLLambdaModelDataSourceStrategy = {
  name: 'MySqlStrategy',
  dbType: 'MYSQL',
  dbConnectionConfig: { /* ... */ },
  vpcConfiguration: { /* vpcId, securityGroupIds, subnetAvailabilityZoneConfig */ },
  minimizeRdsVpcEndpoints: true, // opt in: provision only the `ssm` endpoint
};
```

##### Why scope it to the SQL data source strategy (not a global flag)?

VPC-endpoint provisioning is a property of an individual SQL (RDS) data source — it is only meaningful for a SQL strategy that declares a `vpcConfiguration`. A backend can define multiple data sources, so a global `TransformParameter` / `translationBehavior` flag would apply indiscriminately and could not express "minimize endpoints for *this* RDS data source but not that one." Placing the field on `SQLLambdaModelDataSourceStrategy` keeps the option next to the `vpcConfiguration` it modifies and scopes it precisely to the resource it affects. An earlier revision of this PR plumbed the flag through `TransformParameters` / `translationBehavior`; this revision **removes that global plumbing** in favor of the data-source-scoped field.

##### Why does the default stay `false`? (breaking-change rationale)

The default remains `false` (all five endpoints) deliberately. Removing endpoints from an already-deployed API is a **destructive** CloudFormation change: on the next deploy CFN deletes the four now-unreferenced `AWS::EC2::VPCEndpoint` resources. Defaulting to `true`, or minimizing unconditionally, would silently delete those endpoints from every existing RDS-in-VPC stack on redeploy. Gating the behavior behind an explicit opt-in means existing stacks are untouched until the customer knowingly enables the optimization.

##### Coordinated changes

This is **1 of 3** coordinated repository changes that surface `minimizeRdsVpcEndpoints` end-to-end:

1. **`aws-amplify/amplify-category-api`** (this PR) — adds the field to `SQLLambdaModelDataSourceStrategy` and rewires the RDS transformer consumer.
2. **`aws-amplify/amplify-data`** — surfaces the field in `data-schema-types`.
3. **`aws-amplify/amplify-backend`** — surfaces the field through `backend-data`.

##### Summary of code changes

- Add `minimizeRdsVpcEndpoints?: boolean` (`@default false`) to `SQLLambdaModelDataSourceStrategy` — both the public construct type (`amplify-graphql-api-construct`) and the internal transformer interface (`amplify-graphql-transformer-interfaces`).
- Rewire `RdsModelResourceGenerator` to read `strategy.minimizeRdsVpcEndpoints` (instead of `context.transformParameters.minimizeRdsVpcEndpoints`) and thread it into `getSsmEndpoint` / `createRdsLambda`, which provision `['ssm']` when enabled and the full five-endpoint set otherwise.
- Remove the earlier global plumbing: `minimizeRdsVpcEndpoints` is deleted from `TransformParameters`, `defaultTransformParameters`, `TranslationBehavior`, `PartialTranslationBehavior`, and `defaultTranslationBehavior`.
- Regenerate the `amplify-graphql-api-construct` `.jsii` / `API.md` for the moved public API surface (the diff is limited to the `minimizeRdsVpcEndpoints` move).

No CDK/CFN template *parameters* are added. When `minimizeRdsVpcEndpoints: true`, the number of provisioned [`AWS::EC2::VPCEndpoint`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ec2-vpcendpoint.html) resources for a VPC-bound SQL data source drops from five to one (`ssm` only). With the default (`false`), generated resources are unchanged.

#### Issue #, if available

Related to #3409

- `amplify-graphql-model-transformer` unit tests — `amplify-sql-resource-generator.test.ts`: **12/12 pass**, including:
  - *"provisions all five VPC endpoints by default (minimizeRdsVpcEndpoints defaults to false)"*
  - *"provisions only the ssm VPC endpoint when minimizeRdsVpcEndpoints is enabled"*
- `amplify-graphql-api-construct` functional tests — `sql-model-definition.test.ts`: **7/7 pass**, including *"ssm as credential store with minimizeRdsVpcEndpoints enabled"* (asserts the strategy-scoped field, not `translationBehavior`).
- `tsc` builds pass for the affected packages (`amplify-graphql-transformer-interfaces`, `amplify-graphql-transformer-core`, `amplify-graphql-transformer-test-utils`, `amplify-graphql-model-transformer`); `jsii` regenerates cleanly for `amplify-graphql-api-construct`.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [ ] E2E test run linked


sarayev:
Nice cleanup — thanks for tracking this down! Removing those four unused VPC endpoints is a real cost win for customers, and the root-cause writeup in #3409 makes it easy to see why only `ssm` is actually needed. The diff is tight and the unit test update is spot-on.

I’ve got two things I’d like us to sort out before merging — one about communicating this to existing users, and one about the e2e tests. Details inline. 🙏

packages/amplify-graphql-model-transformer/src/resolvers/rds/resolver.ts
```diff
@@ -98,8 +98,8 @@ export const getSsmEndpoint = (scope: Construct, resourceNames: SQLLambdaResourc
     return Fn.join('', ['ssm.', Fn.ref('AWS::Region'), '.amazonaws.com']);
   }
 
-  // Although the Lambda function will only invoke SSM directly, internally the SDK makes calls to other services as well
-  const services = ['ssm', 'ssmmessages', 'ec2', 'ec2messages', 'kms'];
+  // Only the SSM VPC endpoint is needed for Lambda to access Parameter Store
+  const services = ['ssm'];
```
sarayev: One thing worth calling out: for anyone who already has an RDS-in-VPC API deployed, the next `amplify push` / `cdk deploy` after they pick up this change will have CloudFormation **delete** these four endpoints from their live stack. Functionally that’s fine since nothing was using them — but it does mean an upgrade silently modifies their infrastructure.

Could we add a CHANGELOG / release-note entry flagging this? Folks running cost-anomaly alerts or compliance tooling on their VPCs will see four resources disappear and may wonder why. A one-liner like *“RDS-in-VPC deployments will have four unused VPC endpoints (`ssmmessages`, `ec2`, `ec2messages`, `kms`) removed on next deploy — this is expected and safe”* would save someone a confused investigation later.

packages/amplify-e2e-tests/src/__tests__/rds-pg-array-objects.test.ts
```diff
@@ -137,6 +137,7 @@ describe('RDS Model Directive', () => {
     expect(rdsLambdaFunction.Properties.VpcConfig.SecurityGroupIds.length).toBeGreaterThan(0);
 
     expect(getResource(resources, `${resourceNames.sqlVpcEndpointPrefix}ssm`, CDK_VPC_ENDPOINT_TYPE)).toBeDefined();
+    // TODO(#3449): re-enable the reduced VPC-endpoint assertion (only 'ssm' expected) once graphql-model-transformer >=3.x ships in the released @aws-amplify/cli-internal that this Gen1 amplify-push e2e installs. The reduction is validated by the unit test sql-model-definition.test.ts (endpoints.length === 1).
```
sarayev: I think these e2e assertions are a bit of a time bomb. They still expect all 5 endpoints (`toBeDefined()`), and they only pass today because the installed CLI pulls `@aws-amplify/graphql-model-transformer` from npm at `2.14.2`, which still emits 5. Once this change ships and the published package updates, these will start failing on their own without anyone touching them.

The TODO points at `graphql-model-transformer >=3.x`, but I couldn’t find a timeline for a 3.x release — so this could sit as latent CI breakage for a while. Could we either:
- flip the four assertions to `toBeUndefined()` now (with a short note on the Gen1 CLI version lag), or
- gate them on the installed transformer version if that’s detectable?

Either way, a concrete tracking issue would be better than a version that may not land soon.

svidgen (approved):
Looks good to me if @sarayev is also happy.

packages/amplify-graphql-model-transformer/src/resolvers/rds/resolver.ts
Simone319: Implemented as feature flag instead to avoid breaking existing user experience.

packages/amplify-e2e-tests/src/__tests__/rds-pg-array-objects.test.ts
Simone319: The feature flag approach no longer require the e2e trick.

sarayev (commented):
LGTM as well 👍

PR #3468, opened 2026-04-13
fix: stabilize E2E tests — exclude broken regions, fix Node 22 compatibility

sarayev:
## Summary

Stabilizes E2E test suite by excluding regions with infrastructure issues and fixing Node 22 compatibility problems. Batch 10 result: **190✅ / 1❌** (only `cleanup_e2e_resources` utility failed).

### Region Exclusions

**Exclude `me-south-1` (BAH) and `me-central-1` (DXB)**
- These regions are currently unavailable for E2E test execution.
- Removed from `scripts/e2e-test-regions.json` and `packages/amplify-e2e-core/src/configure/index.ts`.

**Exclude `sa-east-1` (GRU)**
- The `amplify push` command is erroring in this region, causing consistent test failures.
- Removed from `scripts/e2e-test-regions.json` and `packages/amplify-e2e-core/src/configure/index.ts`.

### Node 22 Compatibility Fix

**Fix `sql_generate_unauth` failure (`packages/amplify-e2e-core/src/categories/api.ts`)**
- Changed `.expect()` to `.wait()` for the `generateUnauthSQL` error message assertion.
- **Root cause**: Node 22 emits a punycode deprecation warning (`[DEP0040]`) as the first line of output before the actual error message. `.expect()` uses `shift: true` which only matches the next line — so it matched the deprecation warning instead of the expected error. `.wait()` uses `shift: false` and scans all lines until a match is found.

### Region-Pinned Test

**Pin `searchable-previous-deployment-had-node-to-node` to `us-west-1` (`scripts/split-e2e-tests.ts`)**
- Added to `FORCE_REGION_MAP` so it always runs in `us-west-1`.
- **Root cause**: The test was landing on `ap-northeast-3` (Osaka) via round-robin, where `t2.small.elasticsearch` instance type is deprecated/unavailable.
- Regenerated `codebuild_specs/e2e_workflow.yml` to include `CLI_REGION: us-west-1`.

### Other

- Reverted nexpect output to show last 10 lines (was temporarily set to full output for debugging).
- Added `amplify --version` logging in `packages/amplify-e2e-core/src/categories/function.ts` for debugging.

## Test Results

**Batch 10** (`fix/exclude-bah-dxb-regions`): **190✅ / 1❌**
- Only failure: `cleanup_e2e_resources` (utility task, not a real test)
- All previously-failing tests now pass with these fixes


sarayev:
A successful e2e run can be found [here](https://tiny.amazon.com/1eomwmehb/IsenLink), where only `cleanup-e2e-resources` utility is failing.

svidgen (approved):

adrianjoshua-strutt (approved):

PR #3441, opened 2026-03-19
fix: dep updates

svidgen:
- fast-xml-parser ^5.5.2 -> ^5.5.7 (CVE-2026-26278, entity bypass, DoS, stack overflow)
- Added minimatch ^3.1.5 resolution (ReDoS across nx, api-extractor, etc.)
- Added lodash-es ^4.17.23 resolution (prototype pollution)

Addresses alerts: #295, #294, #302, #307-310, #236, #273-275, #282-285, #290, #292

Note: ajv ^8.18.0 resolution attempted but reverted - breaks commitlint strict mode schema validation in commit-msg hook.

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes
- [ ] E2E test run linked
- [ ] Tests are [changed or added](https://github.com/aws-amplify/amplify-cli/blob/master/CONTRIBUTING.md#tests)
- [ ] Relevant documentation is changed or added (and PR referenced)
- [ ] New AWS SDK calls or CloudFormation actions have been added to relevant test and service IAM policies
- [ ] Any CDK or CloudFormation parameter changes are called out explicitly

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #3474, opened 2026-04-23
fix: pin fast-xml-parser to ~5.6.0 to fix &#xD; entity validation bug

adrianjoshua-strutt:
## Problem

`npx ampx pipeline-deploy` fails for all customers in Amplify Hosting with:
```
[EntityReplacer] Invalid character '#' in entity name: "#xD"
```

## What happened

The @nodable/entities package was extracted from fast-xml-parser into a new repo ([nodable/val-parsers](https://github.com/nodable/val-parsers)) about two weeks ago. Version 2.0.0 introduced a validateEntityName() function that throws when an entity name starts with #. This breaks the AWS SDK, which calls parser.addEntity("#xD", "\r") to handle XML numeric character references like &#xD; (carriage return) in STS responses.

fast-xml-parser 5.7.0 upgraded its dependency from @nodable/entities ^1.1.0 to ^2.1.0, pulling in the broken validation. Any customer whose npm install resolved to fxp >= 5.7.0 after ~April 17 hits this on every deploy.

## Root Cause

- `@nodable/entities@2.0.0` added validation that throws on `#` prefix in entity names
- `fast-xml-parser >=5.7.0` depends on `@nodable/entities ^2.1.0` (broken)
- `fast-xml-parser 5.6.0` depends on `@nodable/entities ^1.1.0` (works)
- `fast-xml-parser <=5.5.12` uses inline EntitiesParser with no validation (works)

The actual root cause will be fixed by https://github.com/nodable/val-parsers/pull/3 allowing use to unpin the version. 

## Fix

1. Pin `fast-xml-parser` to `~5.6.0` in both construct packages and root resolution
2. Pin `@nodable/entities` to `^1.0.0` globally to prevent 2.x from being pulled in
3. Add `@nodable/entities` to `bundledDependencies` in both construct packages

## Verification

- Reproduced the exact error using the customer reproduction app (fxp 5.7.1 + @nodable/entities 2.1.0)
- Verified fix resolves it (fxp 5.6.0 + @nodable/entities 1.1.0)
- XML with `&#xD;` and `&#13;` parses successfully after fix


svidgen (approved):

Simone319:
Wouldn't the issue be fixed by this change? https://github.com/aws-amplify/amplify-category-api/pull/3441/changes

Also, the PR title seems to be wrong as we're pinning to 5.5.12 instead of 5.6.0.

adrianjoshua-strutt:
@Simone319 You are correct. We already pinned the version. I have updated the original issue and will close this ticket.

PR #3475, opened 2026-04-28
feat(amplify-category-api): keep datastore fields when disabling datastore

cadivus:
Needed to e.g. filter for soft deleted data when disabling datastore

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes
- [ ] E2E test run linked


packages/amplify-category-api/src/__tests__/provider-utils/awscloudformation/utils/preserve-sync-fields.test.ts
```diff
@@ -0,0 +1,301 @@
...
+    ObjectTypeDefinition: (node: {
+      name: { value: string };
+      fields?: ReadonlyArray<{
+        name: { value: string };
+        type: { kind: string; name?: { value: string } };
+      }>;
+    }) => {
+      if (node.name.value !== typeName) return undefined;
+      for (const field of node.fields ?? []) {
+        const scalar = field.type.kind === 'NamedType' && field.type.name ? field.type.name.value : '';
```
adrianjoshua-strutt: The code only checks field.type.kind === 'NamedType' when detecting existing sync fields. If someone has _version: Int! in their schema, the `NonNullType` wrapper means it won't be recognized. We should do a recursive check here.

adrianjoshua-strutt (changes requested):

packages/amplify-category-api/src/__tests__/provider-utils/awscloudformation/utils/preserve-sync-fields.test.ts
cadivus: Does the fix work for you?

PR #3491, opened 2026-06-22
fix: stagger e2e batch fan-out into waves to avoid CodeBuild batch-orchestration fault

Simone319:
## Problem
The `amplify-category-api-e2e-workflow` CodeBuild batch deterministically FAULTs with `Internal Service Error` ("CodeBuild is experiencing issues") at the `IN_PROGRESS` batch-orchestration phase, ~22s after `publish_to_local_registry` completes. All downstream test groups are force-STOPPED. This has blocked e2e since ~May 22, 2026.

## Root cause
The generated batch build-graph fans out into ~184 test jobs that **all depend solely on `publish_to_local_registry`**, so they are released in a single simultaneous wave. CodeBuild's batch orchestrator faults on that burst.

This is **not** a concurrency/quota issue (every sampled build had QUEUED=0s and provisioned immediately; `maximumBuildsAllowed=300 > 184`) and **not** caused by test growth (the fan-out has been ~180-193 since Nov 2024, and was higher — 205-229 — in 2024 with no faults). Evidence points to a CodeBuild service-side change ~May 21-22, 2026 that reduced batch-orchestration tolerance: the identical 193-group batch ran healthy on May 19 and faulted on May 22 with no repo change. The pr-workflow (only ~11 groups) is unaffected.

## Fix
Stagger the fan-out into sequential waves in `scripts/split-e2e-tests.ts` so the orchestrator never sees a single >~50-job release burst:
- New constant `WAVE_SIZE = 46`.
- Wave 0 (first 46 jobs) depends on `publish_to_local_registry`; each later job `i` depends on job `i - WAVE_SIZE` (strictly-decreasing index → acyclic; max single-event release width = WAVE_SIZE).
- `cleanup_e2e_resources` now depends on the final wave's leaf jobs so it still runs after all tests.
- Regenerated `codebuild_specs/e2e_workflow.yml`.

**No test logic, test isolation, or job count changes** — only `depend-on` edges. Total jobs unchanged (192); max simultaneous release width 174 → 46.

## Verification
Triggered an e2e batch on this branch (`amplify-category-api-e2e-workflow:a6fe9799-...`): after `publish_to_local_registry` succeeded, wave 1 released ~46-49 jobs and the batch stayed `IN_PROGRESS` with **no FAULT** — clearing the +22s death window that killed every prior unstaggered batch, and holding healthy through the run.

## Tradeoff
Waves serialize, adding some wall-clock latency — an accepted tradeoff to eliminate the deterministic orchestration FAULT.

## Note
The underlying trigger is a CodeBuild service-side regression (~May 22, 2026); this change is a defensive workaround that makes the batch resilient to the lower orchestration ceiling. Worth a CodeBuild support ticket separately.

### DIFF
--- a/codebuild_specs/e2e_workflow.yml
+++ b/codebuild_specs/e2e_workflow.yml
@@ -532,7 +532,8 @@ batch:
           TEST_SUITE: src/__tests__/rds-mysql-multi-auth-1.test.ts
           CLI_REGION: sa-east-1
       depend-on:
-        - publish_to_local_registry
+        - >-
+          custom_transformers_invalid_input_arguments_schema_data_access_patterns_schema_predictions_function_10
     - identifier: rds_mysql_model_v2
       buildspec: codebuild_specs/run_e2e_tests.yml
       env:
@@ -542,7 +543,8 @@ batch:
           TEST_SUITE: src/__tests__/rds-mysql-model-v2.test.ts
           CLI_REGION: us-east-1
       depend-on:
-        - publish_to_local_registry
+        - >-
+          api_7_schema_function_2_global_sandbox_lambda_conflict_handler_index_with_stack_mappings
     - identifier: rds_mysql_field_auth_userpool_static_dynamic
       buildspec: codebuild_specs/run_e2e_tests.yml
       env:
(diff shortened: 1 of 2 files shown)


Simone319:
Closing as the fix is applied in https://github.com/aws-amplify/amplify-category-api/pull/3497.

PR #3456, opened 2026-04-01
fix(searchable): enforce TLS 1.2 on Elasticsearch domains (v1 transformer)

sarayev:
## Description

The v1 `graphql-elasticsearch-transformer` creates Elasticsearch domains without any `DomainEndpointOptions` — no `EnforceHTTPS`, no `TLSSecurityPolicy`. AWS is deprecating TLS 1.0/1.1 support on April 20, 2026.

Adds `DomainEndpointOptions` with `TLSSecurityPolicy: Policy-Min-TLS-1-2-2019-07` to the domain. Does NOT add `EnforceHTTPS` to avoid breaking customers currently using HTTP.

Uses `as any` cast because `cloudform-types@4.2.0` lacks the `DomainEndpointOptions` type, but CloudFormation supports it.

Related: aws-amplify/amplify-cli#14329

### Changes
- `packages/graphql-elasticsearch-transformer/src/resources.ts` — Added `DomainEndpointOptions` with `TLSSecurityPolicy` to `makeElasticsearchDomain()`
- `packages/graphql-elasticsearch-transformer/src/__tests__/SearchableModelTransformer.test.ts` — Added TLS 1.2 assertion test

### CDK / CloudFormation Parameters Changed
`AWS::Elasticsearch::Domain` → `DomainEndpointOptions.TLSSecurityPolicy` set to `Policy-Min-TLS-1-2-2019-07` (previously no `DomainEndpointOptions` at all)

### Validation
- Unit tests: 7/7 pass
- E2E: [batch 406490ab](https://us-east-1.console.aws.amazon.com/codesuite/codebuild/594813022831/projects/amplify-category-api-e2e-workflow/batch/amplify-category-api-e2e-workflow:406490ab-0d9d-4145-b1f0-e7b728838512?region=us-east-1) — 69/82 passed, all 3 searchable tests passed, failures are pre-existing (auth/schema)
- Companion V2 fix: category-api/opensearch-tls-1-2-support (PR against main)

### Checklist
- [x] `yarn test` passes
- [x] Tests are changed or added
- [x] CDK/CloudFormation parameter changes called out
- [x] E2E test run linked

PR #3396, opened 2026-01-14
chore: update node python libraries and windows container

ShadowCat567:
#### Description of changes

`node-pty` updated to latest in preparation for node 24
`node-gyp` updated to more recent version also for compatibility with newer node versions
Windows container updated to be compatible with newer version of Visual Studio

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [ ] E2E test run linked
- [ ] Tests are [changed or added](https://github.com/aws-amplify/amplify-cli/blob/master/CONTRIBUTING.md#tests)
- [ ] Relevant documentation is changed or added (and PR referenced)
- [ ] New AWS SDK calls or CloudFormation actions have been added to relevant test and service IAM policies
- [ ] Any CDK or CloudFormation parameter changes are called out explicitly

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #3494, opened 2026-06-26
fix(amplify-category-api): migrate container templates off bitnami ECR Public + enforce TLS 1.2 on v1 searchable

sarayev:
#### Description of changes

This branch fixes the live breakage caused by Bitnami images being removed from Amazon ECR Public Gallery, fixes a pre-existing container-template regression that predates the removal, and completes the TLS 1.0 → 1.2 enforcement for the v1 (Elasticsearch) searchable transformer.

##### Bitnami ECR Public removal — container templates

Bitnami container images were permanently removed from ECR Public on June 10, 2026, so any `public.ecr.aws/bitnami/*` pull now fails. All four customer-facing container templates scaffolded by the API category referenced Bitnami base images, which breaks `amplify add api` (container) docker builds for every customer on a cold cache.

Per the [AWS guidance](https://aws.amazon.com/blogs/containers/bitnami-image-removal-from-ecr-public/), the blog's primary mitigation (mirror to a private ECR repo) is not appropriate here because these templates are generated into each customer's own project — we cannot embed a customer-specific account/registry URI into scaffolding. The blog's recommended long-term path is to move to an equivalent official base image. These templates were migrated to the Docker Official Images mirror published on ECR Public, preserving `node`/`npm`/`python`/`pip` behavior with no customer configuration:

- `public.ecr.aws/bitnami/node:22` → `public.ecr.aws/docker/library/node:22` (3 templates)
- `public.ecr.aws/bitnami/python:3.12-prod` → `public.ecr.aws/docker/library/python:3.12` (1 template)

Both a node template (graphql-express) and the python template (dockercompose-rest-express) were validated with real `docker build` + `docker run`: builds exit 0, dependencies install on the new bases, and the containers start cleanly.

##### Container regression introduced by #3436

PR #3436 ("chore: update outdated container template dependencies") changed the python base image from the valid tag `public.ecr.aws/bitnami/python:3.8-prod-debian-10` to `public.ecr.aws/bitnami/python:3.12-prod`. The bare `3.12-prod` tag does not exist in the Bitnami python repo — Bitnami `-prod` python tags always carry an OS suffix such as `-prod-debian-12` — so the python template build failed with a manifest-not-found error even *before* the ECR Public removal. The migration above resolves this regression as well, since `public.ecr.aws/docker/library/python:3.12` is a valid, existing tag. (#3396 was also reviewed and touches only Node version bumps / CI config — it does not modify any Dockerfile or container template.)

##### TLS 1.2 enforcement on v1 (Elasticsearch) searchable

PR #3456 set `DomainEndpointOptions.TLSSecurityPolicy = 'Policy-Min-TLS-1-2-2019-07'` on the v1 Elasticsearch domain but did not set `EnforceHTTPS`. On `AWS::Elasticsearch::Domain`, the minimum TLS policy only governs the HTTPS endpoint, so without `EnforceHTTPS: true` the policy is not applied and the domain keeps reporting `Policy-Min-TLS-1-0-2019-07` — which is why amplify-cli e2e still produced a `Policy-Min-TLS-1-0` CREATE_FAILED for v1 searchable after #3456. The companion v2 (OpenSearch) fix in #3472 already pairs `enforceHttps: true` with `tlsSecurityPolicy: TLS_1_2` and is complete; only the v1 path was left incomplete.

This branch adds `EnforceHTTPS: true` to the v1 domain's `DomainEndpointOptions` so the TLS 1.2 minimum is actually enforced, matching the v2 behavior. The existing unit test (which incorrectly asserted `EnforceHTTPS` was `undefined`) was updated to assert it is now `true`.

- `AWS::Elasticsearch::Domain` `DomainEndpointOptions.EnforceHTTPS` set to `true` on the v1 searchable domain (paired with the existing `TLSSecurityPolicy: Policy-Min-TLS-1-2-2019-07`): https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-elasticsearch-domain-domainendpointoptions.html

#### Issue #, if available

N/A — addresses the Bitnami ECR Public removal and follow-up to #3456 / #3472.

- Built and ran both container templates on the new official base images:
  - `graphql-express` (node): `docker build` exits 0; container logs `Listening on localhost:3000/graphql`.
  - `dockercompose-rest-express/python`: `docker build` exits 0; Flask dev server starts on `0.0.0.0:5000`.
- Confirmed `public.ecr.aws/bitnami/node:22` no longer resolves (manifest inspect fails) while `public.ecr.aws/docker/library/{node:22,python:3.12}` resolve.
- `jest SearchableModelTransformer` in `graphql-elasticsearch-transformer`: 7 passed, 7 snapshots passed, `resources.ts` at 100% coverage. Confirmed the compiled `lib/resources.js` emits both `EnforceHTTPS: true` and `TLSSecurityPolicy: 'Policy-Min-TLS-1-2-2019-07'`.
- Cloud e2e was intentionally NOT run (handled separately).

#### Checklist

- [x] PR description included
- [x] `yarn test` passes (scoped to the touched `graphql-elasticsearch-transformer` package)
- [ ] Relevant documentation is changed or added (and PR referenced) — no docs reference these templates on this branch

### DIFF
--- a/packages/amplify-category-api/resources/awscloudformation/container-templates/dockercompose-rest-express/express/Dockerfile
+++ b/packages/amplify-category-api/resources/awscloudformation/container-templates/dockercompose-rest-express/express/Dockerfile
@@ -1,4 +1,4 @@
-FROM public.ecr.aws/bitnami/node:22
+FROM public.ecr.aws/docker/library/node:22
 
 ENV PORT=8080
 EXPOSE 8080
--- a/packages/amplify-category-api/resources/awscloudformation/container-templates/dockercompose-rest-express/python/Dockerfile
+++ b/packages/amplify-category-api/resources/awscloudformation/container-templates/dockercompose-rest-express/python/Dockerfile
@@ -1,5 +1,5 @@
 # set base image (host OS)
-FROM public.ecr.aws/bitnami/python:3.12-prod
+FROM public.ecr.aws/docker/library/python:3.12
 
 # set the working directory in the container
 WORKDIR /code
--- a/packages/amplify-category-api/resources/awscloudformation/container-templates/dockerfile-rest-express/Dockerfile
+++ b/packages/amplify-category-api/resources/awscloudformation/container-templates/dockerfile-rest-express/Dockerfile
@@ -1,4 +1,4 @@
-FROM public.ecr.aws/bitnami/node:22
+FROM public.ecr.aws/docker/library/node:22
 
 ENV PORT=8080
 EXPOSE 8080
--- a/packages/amplify-category-api/resources/awscloudformation/container-templates/graphql-express/Dockerfile
+++ b/packages/amplify-category-api/resources/awscloudformation/container-templates/graphql-express/Dockerfile
@@ -1,4 +1,4 @@
-FROM public.ecr.aws/bitnami/node:22
+FROM public.ecr.aws/docker/library/node:22
 
 WORKDIR /usr/src/app
 COPY package*.json ./
--- a/packages/amplify-e2e-core/global-setup.js
+++ b/packages/amplify-e2e-core/global-setup.js
@@ -1,0 +1,4 @@
+// Allow people to use `amplify-category-api-e2e-core/global-setup` as a jest globalSetup.
+const globalSetup = require('./lib/jest-global-setup');
+
+module.exports = globalSetup.default || globalSetup;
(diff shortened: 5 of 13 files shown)


sarayev:
Superseded by full split: #3496 (bitnami ECR removal), #3498 (v1 searchable TLS 1.2 EnforceHTTPS), #3499 (Gen1 e2e placeholder + harness logging) — plus a forthcoming @function-migration PR. Closing this mixed draft.

PR #3495, opened 2026-06-26
fix(graphql-generation-transformer): grant InvokeModel on inference profile and underlying models

sarayev:
#### Description of changes

The `@generation` transformer creates an AppSync→Bedrock IAM role that grants `bedrock:InvokeModel`. Previously the role scoped the action to a single regional foundation-model ARN:

```
arn:<partition>:bedrock:<region>::foundation-model/<modelId>
```

For a **cross-region inference profile** model id (prefixed `us.`, `eu.`, or `apac.` — e.g. `us.anthropic.claude-haiku-4-5-20251001-v1:0`) that ARN is incorrect. Invoking an inference profile requires `bedrock:InvokeModel` on the **inference-profile ARN** *and* on the **underlying regional foundation-model ARNs** the profile routes to. Because the transformer only granted the (malformed) foundation-model ARN, customers using a cross-region profile id with `@generation` hit runtime **AccessDenied**.

(The `@conversation` path is unaffected — it invokes Bedrock through `@aws-amplify/ai-constructs`.)

##### Fix

When the model id matches `/^(us|eu|apac)\./`, the role now grants `bedrock:InvokeModel` on both:

- the inference-profile ARN: `arn:<partition>:bedrock:<region>:<account>:inference-profile/<modelId>`
- the underlying foundation-model ARN with a wildcard region: `arn:<partition>:bedrock:*::foundation-model/<foundationModelId>`, where `<foundationModelId>` is the profile id with the region prefix stripped. A wildcard region is used because a cross-region profile may dispatch to any of several regions.

Plain foundation-model ids keep the existing single-ARN behavior. All ARNs are built from CDK `partition`/`region`/`account` tokens — nothing is hardcoded.

##### Tests

Added unit/snapshot tests in the generation transformer package proving:
- a plain FM id grants only the foundation-model ARN (unchanged behavior), and
- an inference-profile id grants both the profile ARN and the wildcard-region foundation-model ARN.

These tests are offline (no AWS). Full suite: **16 passed, 16 total** (`jest`).

None. Only the resource list of an existing inline IAM policy statement changes.

#### Issue #, if available

N/A

- `jest` (offline) — 16 passed, 16 total, including 2 new tests + 2 new snapshots.
- `tsc --noEmit` — passes (exit 0).

#### Checklist

- [x] PR description included
- [x] `yarn test` passes (generation transformer package: 16/16)
- [ ] E2E test run linked
- [ ] Relevant documentation is changed or added (no docs file exists for this package)

### DIFF
--- a/packages/amplify-graphql-generation-transformer/src/__tests__/__snapshots__/amplify-graphql-generation-transformer.test.ts.snap
+++ b/packages/amplify-graphql-generation-transformer/src/__tests__/__snapshots__/amplify-graphql-generation-transformer.test.ts.snap
@@ -148,6 +148,62 @@ function createUserAgent(request) {
 }
 `;
 
+exports[`generation route bedrock InvokeModel IAM resources cross-region inference profile id grants the profile ARN and wildcard-region foundation-model ARN 1`] = `
+[
+  {
+    "Fn::Join": [
+      "",
+      [
+        "arn:",
+        {
+          "Ref": "AWS::Partition",
+        },
+        ":bedrock:",
+        {
+          "Ref": "AWS::Region",
+        },
+        ":",
+        {
+          "Ref": "AWS::AccountId",
+        },
+        ":inference-profile/us.anthropic.claude-haiku-4-5-20251001-v1:0",
+      ],
+    ],
+  },
+  {
+    "Fn::Join": [
+      "",
+      [
+        "arn:",
+        {
+          "Ref": "AWS::Partition",
+        },
+        ":bedrock:*::foundation-model/anthropic.claude-haiku-4-5-20251001-v1:0",
+      ],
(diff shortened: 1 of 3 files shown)


sarayev:
Closing as superseded. This PR added IAM for cross-region Bedrock **inference profiles** to support the @generation e2e using a Claude inference-profile model. The e2e fix landed differently: #3500 switched @generation to **amazon.nova-lite-v1:0**, a plain ON_DEMAND foundation model that needs no inference-profile IAM. The generation shard now passes (validated in batch d7085183). The @conversation path uses ai-constructs, which already handles the inference-profile IAM (no AccessDenied observed). So this inference-profile IAM change is no longer needed. Closing.

PR #3499, opened 2026-06-26
test(amplify-category-api-e2e-core): ensure Gen1 placeholder app + improve harness failure logging

sarayev:
## Description

Two Gen1 e2e harness improvements in `amplify-e2e-core`.

#### Ensure Gen1 placeholder app before `amplify init`

Adds an idempotent, best-effort run-start step that guarantees the Gen1 deprecation-bypass placeholder app (`DoNotDeleteAppToBypassGen1Deprecation`) and its `test` backend environment exist in each e2e shard's account + region before any `amplify init` runs. This self-heals the Gen1 end-of-life gate: if a prior run's cleanup deleted the placeholder app, the suite would otherwise be blocked.

The new `ensureGen1PlaceholderApp(region)` helper uses `@aws-sdk/client-amplify`: it paginates `ListApps`, creates the app via `CreateApp` only when missing, and creates the `test` environment via `CreateBackendEnvironment` only when `ListBackendEnvironments` shows it absent. It defaults region to `process.env.CLI_REGION`, relies on the ambient e2e credentials, and swallows/logs all errors so a healthy run is never broken. It is wired as a Jest `globalSetup` hook in both Gen1 e2e packages (`amplify-e2e-tests` and `graphql-transformers-e2e-tests`) so it runs once per shard, via standalone modules that avoid the heavy e2e-core barrel (and its native `node-pty` dependency).

#### Improve harness failure logging

`initJSProjectWithProfile` now emits an explicit `initJSProjectWithProfile SUCCESS` / `FAILED: <err>` instead of the misleading unconditional `Result of initJSProjectWithProfile: null` on success.

The nspawn `run()` exit handler now dumps the full captured output plus the signal alongside the exit code on a non-zero exit, instead of only the last 10 lines (`lastScreen`) which could be empty for push/add-function failures and hid the real error. Timeout branches and the quiet success path are unchanged.

## Testing

`tsc` of `amplify-e2e-core` passes. Logging changes are diagnostic-only with no behavior change.

Split out from #3494 (PR-6 of the full split).

### DIFF
--- a/packages/amplify-e2e-core/global-setup.js
+++ b/packages/amplify-e2e-core/global-setup.js
@@ -1,0 +1,4 @@
+// Allow people to use `amplify-category-api-e2e-core/global-setup` as a jest globalSetup.
+const globalSetup = require('./lib/jest-global-setup');
+
+module.exports = globalSetup.default || globalSetup;
--- a/packages/amplify-e2e-core/package.json
+++ b/packages/amplify-e2e-core/package.json
@@ -22,6 +22,7 @@
     "clean": "rimraf ./lib"
   },
   "dependencies": {
+    "@aws-sdk/client-amplify": "~3.600.0",
     "@aws-sdk/client-amplifybackend": "~3.600.0",
     "@aws-sdk/client-appsync": "~3.600.0",
     "@aws-sdk/client-cloudformation": "~3.600.0",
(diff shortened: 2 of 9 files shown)


Simone319:
Why is https://github.com/aws-amplify/amplify-category-api/pull/3499 and https://github.com/aws-amplify/amplify-category-api/pull/3497 both implementing the ensureGen1PlaceholderApp? Look like they are merging into 2 different branches. Is there any mechanism to keep the branches in sync?

Simone319 (approved):

PR #3510, opened 2026-07-08
fix(schema-generator): sanitize database identifiers in TypeScript output path

svidgen:
## Summary

Sanitizes database column names and enum type names in the TypeScript data schema output path, matching the existing `cleanMappedName` behavior in the GraphQL-SDL output path.

## Problem

The TypeScript schema generator passed raw database column names and enum type names directly to `ts.factory.createIdentifier()`, which emits them as bare JavaScript identifiers without any sanitization. Names containing special characters (parentheses, brackets, quotes, etc.) could produce invalid or unsafe TypeScript output.

## Changes

- Add `sanitizeIdentifier()` helper that strips leading non-alphabetic characters and replaces non-alphanumeric characters with underscores (matching `cleanMappedName` from the GraphQL-SDL path)
- Use `createStringLiteral` instead of `createIdentifier` for field property names in `createProperty()`
- Use `createStringLiteral` instead of `createIdentifier` for enum type property names in `createEnums()`
- Sanitize enum type names in `a.ref()` calls for consistency between the enum definition and its references
- Fix enum deduplication logic to work with `StringLiteral` property names (uses `.text` instead of `.escapedText`)

## Defense in Depth

This is a defense-in-depth measure. The TS output path should apply the same sanitization as the GraphQL-SDL path to ensure consistent, safe output regardless of database column naming.

## Testing

- Added 5 new unit tests verifying:
  - Column names with shell metacharacters are sanitized
  - Column names with brackets/special chars are sanitized  
  - Enum type names with special characters are sanitized
  - Fallback names are generated for all-special-character column names
  - Property keys are emitted as string literals
- Updated 16 existing snapshots to reflect the new quoted property key format
- All 100 tests in the package pass

### DIFF
--- a/packages/amplify-graphql-schema-generator/src/ts-schema-generator/helpers.ts
+++ b/packages/amplify-graphql-schema-generator/src/ts-schema-generator/helpers.ts
@@ -21,6 +21,25 @@ const GQL_TYPESCRIPT_DATA_SCHEMA_TYPE_MAP = {
 };
 
 /**
+ * Sanitizes a database identifier (column name, enum name) to produce a safe TypeScript property name.
+ * Matches the behavior of `cleanMappedName` in the GraphQL-SDL output path.
+ * - If the name has no alphabetic characters, a fallback is generated.
+ * - Leading non-alphabetic characters are stripped.
+ * - Non-alphanumeric characters (except underscores) are replaced with underscores.
+ */
+const sanitizeIdentifier = (name: string, isField: boolean = false): string => {
+  if (!name?.match(/[a-zA-Z]/)) {
+    const suffix = name?.replace(/[^0-9]+/g, '') ?? '';
+    return isField ? `field${suffix}` : `Model${suffix}`;
+  }
+
+  return name
+    .replace(/^[^a-zA-Z]+/, '')
+    .replace(/[^a-zA-Z0-9_]+/g, '_')
+    .trim();
+};
+
+/**
  * Creates a typescript data schema property from internal SQL schema representation
  * Example typescript data schema property output: `id: a.string().required()`
  * @param field SQL IR field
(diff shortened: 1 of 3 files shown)


svidgen:
## E2E & PR Checks Results ✅

Both batches completed successfully:

**E2E batch** (`amplify-category-api-e2e-workflow:fec71fde-851d-46f5-ad25-8c7e92db76f5`): **195/196 passed**
- Only persistent failure: `cleanup_e2e_resources` — this is a post-test infrastructure cleanup job, not an actual test. All 195 real test builds passed.
- [CodeBuild batch](https://tiny.amazon.com/i0d7kn4n/isenamazfede)

**PR checks batch** (`amplify-category-api-pr-workflow:fff7e3bd-b94c-4947-9ce9-18e35d6b16bd`): **11/11 passed** ✅
- [CodeBuild batch](https://tiny.amazon.com/18ksvk7p6/isenamazfede)

Ready for merge.

packages/amplify-graphql-schema-generator/src/ts-schema-generator/helpers.ts
```diff
@@ -20,6 +20,25 @@ const GQL_TYPESCRIPT_DATA_SCHEMA_TYPE_MAP = {
...
+ * Matches the behavior of `cleanMappedName` in the GraphQL-SDL output path.
+ * - If the name has no alphabetic characters, a fallback is generated.
+ * - Leading non-alphabetic characters are stripped.
+ * - Non-alphanumeric characters (except underscores) are replaced with underscores.
+ */
+const sanitizeIdentifier = (name: string, isField: boolean = false): string => {
+  if (!name?.match(/[a-zA-Z]/)) {
+    const suffix = name?.replace(/[^0-9]+/g, '') ?? '';
+    return isField ? `field${suffix}` : `Model${suffix}`;
+  }
```
Simone319, line 34: Even though unlikely, if multiple fields contain no letter, and no number, they will result in the same name `field`. Shall we add a randomized suffix?

svidgen: I think that equates to a bad state. Randomization doesn't give an anchor point to code against.

Simone319: Fair point — randomization is the wrong call; a non-deterministic name would churn on every regen and give consumers nothing stable to code against. Withdrawing that suggestion.

That said, the underlying collision is still real: two all-special-character columns (e.g. () and @#$) both collapse to field, producing duplicate property keys. Since a column with zero alphanumeric characters is arguably a bad-state input anyway, a good follow-up would be to error with a clear message (pointing the user at an explicit rename/@mapsTo) rather than silently coercing. Fine to leave that for a separate change — flagging it so we don't lose the thread.

Simone319 (approved):

PR #3517, opened 2026-07-28
fix(graphql-transformer-core): silence CfnResource#addDependency deprecation warning on aws-cdk-lib >= 2.262.0

Simone319:
### Problem

On `aws-cdk-lib >= 2.262.0`, Amplify's transformer packages emit a jsii deprecation warning — once per resource dependency edge — during `synth`/deploy:

```
[WARNING] aws-cdk-lib.CfnResource#addDependency is deprecated. Use addResourceDependency instead. This API will be removed in the next major release.
```

[aws/aws-cdk#38314](https://github.com/aws/aws-cdk/pull/38314) (shipped in `aws-cdk-lib` 2.262.0) renamed the public `CfnResource#addDependency` method to `addResourceDependency` and marked the old name `@deprecated`. Amplify calls `addDependency` internally, so any consumer whose resolved `aws-cdk-lib` is `>= 2.262.0` sees the warning repeated many times — cosmetic (deploys succeed) but it drowns out genuine warnings.

### Why not a plain rename

This repo pins `aws-cdk-lib@2.260.0`, where `addResourceDependency` does **not** exist publicly (only an `@internal` `_addResourceDependency`) — a plain rename fails to compile (TS2551). Bumping the peer-dependency floor to `>= 2.262.0` would be a breaking change for consumers pinned below that. So this uses a version-tolerant runtime feature-detect instead.

### Fix

A small helper `addCfnResourceDependency(source, target)` in `packages/amplify-graphql-transformer-core/src/utils/cfn-dependency.ts` that calls `addResourceDependency` when it exists (CDK `>= 2.262.0`) and falls back to `addDependency` otherwise (CDK `<= 2.261.0`). Wired into the 4 `CfnResource` call sites. The unrelated construct-node `.node.addDependency` call is left untouched. The `as`-cast is contained entirely within the helper.

### Validation

- CDK 2.260.0 (pinned): compiles; `amplify-graphql-transformer-core` unit tests pass (incl. 3 dedicated `cfn-dependency.test.ts` cases); all snapshots unchanged (identical CloudFormation `DependsOn`).
- CDK 2.262.1: synth harness shows the deprecation warning count go 1 → 0, with byte-identical `DependsOn` output.

### Scope

This PR is scoped to the `addCfnResourceDependency` shim only. Two changes that previously rode along have been split out:
- `getGlobalSecondaryIndexes` accessor centralization → #3511
- `cdk init` tsc-strip fail-fast guard → #3520

### DIFF
--- a/packages/amplify-graphql-model-transformer/src/resources/dynamo-model-resource-generator.ts
+++ b/packages/amplify-graphql-model-transformer/src/resources/dynamo-model-resource-generator.ts
@@ -2,7 +2,7 @@ import * as cdk from 'aws-cdk-lib';
 import { TransformerContextProvider } from '@aws-amplify/graphql-transformer-interfaces';
 import { ModelResourceIDs, ResourceConstants, SyncResourceIDs } from 'graphql-transformer-common';
 import { ObjectTypeDefinitionNode } from 'graphql';
-import { SyncUtils, setResourceName } from '@aws-amplify/graphql-transformer-core';
+import { SyncUtils, setResourceName, addCfnResourceDependency } from '@aws-amplify/graphql-transformer-core';
 import { AttributeType, CfnTable, ITable, StreamViewType, Table, TableEncryption } from 'aws-cdk-lib/aws-dynamodb';
 import { CfnDataSource } from 'aws-cdk-lib/aws-appsync';
 import * as iam from 'aws-cdk-lib/aws-iam';
@@ -139,7 +139,7 @@ export class DynamoModelResourceGenerator extends ModelResourceGenerator {
     );
 
     const cfnDataSource = dataSource.node.defaultChild as CfnDataSource;
-    cfnDataSource.addDependency(role.node.defaultChild as CfnRole);
+    addCfnResourceDependency(cfnDataSource, role.node.defaultChild as CfnRole);
 
     if (context.isProjectUsingDataStore()) {
       // eslint-disable-next-line @typescript-eslint/no-explicit-any
--- a/packages/amplify-graphql-transformer-core/API.md
+++ b/packages/amplify-graphql-transformer-core/API.md
@@ -105,6 +105,9 @@ import { Visibility } from 'aws-cdk-lib/aws-appsync';
 import { VTLRuntimeTemplate } from '@aws-amplify/graphql-transformer-interfaces';
 
 // @public (undocumented)
+export const addCfnResourceDependency: (source: CfnResource, target: CfnResource) => void;
+
+// @public (undocumented)
 export const APICategory = "api";
 
 // @public (undocumented)
--- a/packages/amplify-graphql-transformer-core/src/__tests__/cfn-dependency.test.ts
+++ b/packages/amplify-graphql-transformer-core/src/__tests__/cfn-dependency.test.ts
@@ -1,0 +1,40 @@
+import { App, Stack, CfnResource } from 'aws-cdk-lib';
+import { Template } from 'aws-cdk-lib/assertions';
+import { addCfnResourceDependency } from '../utils/cfn-dependency';
+
+const makePair = () => {
+  const stack = new Stack(new App(), 'TestStack');
+  const target = new CfnResource(stack, 'Target', { type: 'AWS::IAM::Role', properties: {} });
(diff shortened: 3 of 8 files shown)


packages/amplify-graphql-transformer-core/src/utils/schema-utils.ts
```diff
@@ -6,15 +6,26 @@ import { ListValueNode, ObjectTypeDefinitionNode, StringValueNode, TypeNode } fr
...
+ * Reads the list of global secondary indexes tracked on a DynamoDB L2 `Table` construct.
+ *
+ * aws-cdk-lib 2.260 renamed the private `globalSecondaryIndexes` array on the DynamoDB L2 `Table` to
+ * `_globalSecondaryIndexes` (now an `ArrayBox` exposing the same `find`/`some`/`length` surface and the
+ * same `{ indexName, keySchema }` element shape); Amplify's managed-table construct keeps the public
+ * `globalSecondaryIndexes` array. This reads whichever exists (covers standard `Table` + Amplify managed
+ * table). Centralized so a future CDK rename is a one-line fix.
+ */
+export const getGlobalSecondaryIndexes = (table: any): any => table['_globalSecondaryIndexes'] ?? table.globalSecondaryIndexes;
+
```
sarayev: Question: This PR's title/description scope it to the `addCfnResourceDependency` CDK deprecation shim, but it also dedupes the `getGlobalSecondaryIndexes` accessor across three packages (looks like a follow-up to #3505). Is this change relevant to / intentionally part of this PR, or should it be split into its own PR so the CDK-dependency change can be reviewed and land independently?

sarayev (commented):
Overall LGTM, left some questions about test coverage and some changes that seemed unrelated to the addDependency warning changes.

packages/amplify-graphql-api-construct-tests/src/commands.ts
```diff
@@ -91,6 +91,12 @@ const removeWholeProjectTypecheckFromSynth = (projectPath: string): void => {
   }
   const appWithoutTypecheck = cdkJson.app.replace(/^\s*npx\s+tsc\s*&&\s*/, '');
   if (appWithoutTypecheck === cdkJson.app) {
+    if (cdkJson.app.includes('tsc')) {
```
sarayev: Question: This adds a fail-fast guard in `removeWholeProjectTypecheckFromSynth`, and the commit message notes it addresses review feedback on #3519 (a different, already-merged PR). Is this intentionally bundled here? If it's independent, splitting it out would let #3519's outstanding feedback land immediately rather than being gated behind this PR.

packages/amplify-graphql-transformer-core/src/utils/schema-utils.ts
```diff
@@ -6,15 +6,26 @@ import { ListValueNode, ObjectTypeDefinitionNode, StringValueNode, TypeNode } fr
 import { ModelResourceIDs, getBaseType } from 'graphql-transformer-common';
 import { getModelDataSourceStrategy } from './model-datasource-strategy-utils';
 
+/**
+ * Reads the list of global secondary indexes tracked on a DynamoDB L2 `Table` construct.
+ *
```
sarayev: Question: `getGlobalSecondaryIndexes` is now exported as public API but is only exercised indirectly through consumer snapshot tests. Given `addCfnResourceDependency` got dedicated unit tests in this same PR, do you think direct unit tests are warranted here too (e.g. `_globalSecondaryIndexes` set / only public `globalSecondaryIndexes` / neither)? Happy to defer if you feel the snapshot coverage is sufficient.

packages/amplify-graphql-transformer-core/src/utils/schema-utils.ts
Simone319: Thanks for the review! You were right that this PR had grown beyond the `addDependency` deprecation shim. I've split it into three: this PR (#3517) is now just the CDK deprecation shim + its tests; the GSI-accessor centralization moved to #3511 (with the direct unit tests you suggested); and the cdk init tsc-strip fail-fast guard moved to #3520. Each can now be reviewed and land independently.

sarayev (approved):

Issue #3489, opened 2026-06-11
Gen1→Gen2 migration: @primaryKey ignored for imported DynamoDB tables — deploy fails with "Imported table properties did not match the expected table properties"


francescoRobin:
## Description

When migrating a Gen 1 app to Gen 2 with `migratedAmplifyGen1DynamoDbTableMappings`, any model using a custom `@primaryKey` (e.g. `userId: String! @primaryKey(sortKeyFields: ["typeStartDate"])`) fails to deploy. The `Custom::ImportedAmplifyDynamoDBTable` synthesizes with the default `id` HASH key, so the import validation in the table manager rejects it:

```
Imported table properties did not match the expected table properties.
AttributeDefinitions does not match the expected value.
Imported Value: [{"AttributeName":"startDate","AttributeType":"S"},{"AttributeName":"type","AttributeType":"S"},{"AttributeName":"typeStartDate","AttributeType":"S"},{"AttributeName":"userId","AttributeType":"S"}]
Expected: [{"AttributeType":"S","AttributeName":"id"}]
KeySchema does not match the expected value.
Imported Value: [{"AttributeName":"typeStartDate","KeyType":"RANGE"},{"AttributeName":"userId","KeyType":"HASH"}]
Expected: [{"AttributeName":"id","KeyType":"HASH"}]
```

In our migration (23 models), the 22 models with the default `id` key imported fine; the single model with a custom `@primaryKey` failed every deploy.

## Reproduction

1. Gen 1 app with a model using a custom primary key:

```graphql
type HealthRecord @model {
  userId: String! @primaryKey(sortKeyFields: ["typeStartDate"])
  typeStartDate: String!
  type: String! @index(name: "MetricTypeIndex", sortKeyFields: ["startDate"])
  startDate: AWSDateTime!
}
```

2. Run the `gen2-migration` flow (`assess` → `lock` → `generate`), keeping the generated `migratedAmplifyGen1DynamoDbTableMappings` so the Gen 1 tables are reused.
3. Deploy the Gen 2 branch (`ampx pipeline-deploy`).
4. The model's nested stack fails with the error above; the root stack ends in `ROLLBACK_COMPLETE`.

## Root cause

`replaceDdbPrimaryKey()` in `amplify-graphql-index-transformer/src/resolvers/resolvers.ts` gates its CFN-level override on `isAmplifyDynamoDbModelDataSourceStrategy(strategy)`:

```ts
[...]

PR #3523, opened 2026-08-24
fix(graphql-model-transformer): derive imported DynamoDB table key schema from @primaryKey

sharonyajain:
## Problem

When migrating a Gen 1 backend to Gen 2 with `migratedAmplifyGen1DynamoDbTableMappings`, any model whose schema declares a custom `@primaryKey` (partition key other than `id`) fails deployment. The `Custom::ImportedAmplifyDynamoDBTable` resource is synthesized with an *expected* key schema of `id`/`HASH`, so the TableManager import validation rejects the (correct) live Gen 1 table:

```
Received response status [FAILED] from custom resource. Message returned:
Imported table properties did not match the expected table properties.
```

The branch stack then rolls back (`ROLLBACK_COMPLETE`); because it is a failed CREATE, the stack must be deleted manually before any retry.

Reported in aws-amplify/amplify-category-api#3489 (surfaced downstream as aws-amplify/amplify-backend#3281).

## Why it matters

This blocks Gen 1 → Gen 2 migration for any customer whose data model uses a custom primary key — a common pattern — mid-migration, with a rough manual-cleanup recovery. In a reported 19-model schema, 18 tables (default `id` keys, several with GSIs) migrated fine; the single model with a custom `@primaryKey` was the only failure.

## Fix (symptom → root cause → change)

- **Symptom:** the imported table's expected `keySchema`/`attributeDefinitions` are `id`/`HASH` regardless of the model's `@primaryKey`.
- **Root cause:** `AmplifyDynamoModelResourceGenerator.createModelTable` instantiates `AmplifyDynamoDBTable` with a hardcoded `partitionKey: { name: 'id', type: STRING }` for **every** table, including imported ones. For owned tables the `@primaryKey` transformer (`replaceDdbPrimaryKey` in the index transformer) corrects the key schema downstream via `cfnTable.addPropertyOverride(...)`. That correction does not reach the imported-table validation path, which consumes the initial construct properties directly (`import-table.ts` → `getExpectedKeySchema` returns `createTableInput.KeySchema` verbatim). GSIs are unaffected because they are already derived from the model.
- **Change:** in `createModelTable`, **for imported tables only**, derive the partition key — and the sort key (single or composite) — from the model's `@primaryKey` via `getPrimaryKeyFieldNodes`, mirroring how GSIs are already derived from the model. Composite sort keys reuse the same `ModelCompositeKeySeparator()` join as the transformer's `getSortKeyName`. Owned tables are untouched (still `id`, corrected downstream exactly as before).

## Tests

Added three regression tests in `amplify-dynamodb-table-generator.test.ts`, each asserting the synthesized `Custom::ImportedAmplifyDynamoDBTable` properties:

- **custom `@primaryKey` (partition key ≠ `id`)** → `keySchema`/`attributeDefinitions` use `userAuthenticating`, not `id`.
- **custom `@primaryKey` with a sort key** → `keySchema` is `[HASH, RANGE]` on the model fields (this case caught a gap in the first iteration and drove the sort-key handling).
- **imported table without a custom `@primaryKey`** → still `id`/`HASH` (no regression).

Verified these tests **fail on the base code** (imported table synthesizes `id`/`HASH`) and pass with the fix — confirmed reproduction, not just static analysis.

## Manual verification

- `amplify-dynamodb-table-generator.test.ts`: 7/7 pass.
- Full `@aws-amplify/graphql-model-transformer` suite: 12 suites / 203 tests / 87 snapshots pass — **no snapshot drift** (the change is gated on `isTableImported`).
- `@aws-amplify/graphql-index-transformer` suite: 93 tests pass.
- `tsc --noEmit` on `graphql-model-transformer`: clean.

## Screenshots

N/A — no user-visible UI change (GraphQL transformer synthesis only).

### DIFF
--- a/packages/amplify-graphql-model-transformer/src/__tests__/amplify-dynamodb-table-generator.test.ts
+++ b/packages/amplify-graphql-model-transformer/src/__tests__/amplify-dynamodb-table-generator.test.ts
@@ -244,6 +244,110 @@ describe('ModelTransformer:', () => {
     validateModelSchema(parse(out.schema));
   });
 
+  it('derives the imported table key schema from a custom @primaryKey with a composite sort key', async () => {
+    const validSchema = `
+      type Event @model {
+        tenantId: ID! @primaryKey(sortKeyFields: ["region", "createdAt"])
+        region: String!
+        createdAt: String!
+        name: String
+      }
+    `;
+    const out = testTransform({
+      schema: validSchema,
+      transformers: [new ModelTransformer(), new PrimaryKeyTransformer()],
+      dataSourceStrategies: {
+        Event: {
+          dbType: 'DYNAMODB' as const,
+          provisionStrategy: 'IMPORTED_AMPLIFY_TABLE' as const,
+          tableName: 'Event-myApiId-myEnv',
+        },
+      },
+    });
+    const table = out.stacks['Event'].Resources?.EventTable;
+    expect(table).toBeDefined();
+    expect(table.Type).toBe(CUSTOM_IMPORTED_DDB_CFN_TYPE);
+    // A composite sort key is stored as a single string attribute named by joining the fields.
+    expect(table.Properties.keySchema).toEqual([
+      { attributeName: 'tenantId', keyType: 'HASH' },
+      { attributeName: 'region#createdAt', keyType: 'RANGE' },
+    ]);
+    expect(table.Properties.attributeDefinitions).toEqual(
+      expect.arrayContaining([
+        { attributeName: 'tenantId', attributeType: 'S' },
+        { attributeName: 'region#createdAt', attributeType: 'S' },
+      ]),
+    );
+    validateModelSchema(parse(out.schema));
+  });
+
+  it('derives a numeric attribute type for an integer @primaryKey / sort key on an imported table', async () => {
+    const validSchema = `
+      type Reading @model {
+        sensorId: Int! @primaryKey(sortKeyFields: ["ts"])
+        ts: Int!
+        value: Float
+      }
+    `;
+    const out = testTransform({
+      schema: validSchema,
+      transformers: [new ModelTransformer(), new PrimaryKeyTransformer()],
+      dataSourceStrategies: {
+        Reading: {
+          dbType: 'DYNAMODB' as const,
+          provisionStrategy: 'IMPORTED_AMPLIFY_TABLE' as const,
+          tableName: 'Reading-myApiId-myEnv',
+        },
+      },
+    });
+    const table = out.stacks['Reading'].Resources?.ReadingTable;
+    expect(table).toBeDefined();
+    expect(table.Type).toBe(CUSTOM_IMPORTED_DDB_CFN_TYPE);
+    expect(table.Properties.keySchema).toEqual([
+      { attributeName: 'sensorId', keyType: 'HASH' },
+      { attributeName: 'ts', keyType: 'RANGE' },
+    ]);
+    expect(table.Properties.attributeDefinitions).toEqual(
+      expect.arrayContaining([
+        { attributeName: 'sensorId', attributeType: 'N' },
+        { attributeName: 'ts', attributeType: 'N' },
+      ]),
+    );
+    validateModelSchema(parse(out.schema));
+  });
+
+  it('derives a string attribute type for an enum-typed @primaryKey on an imported table', async () => {
+    // Enum-backed key fields are stored as strings. `attributeTypeFromScalar` throws on non-scalar
+    // (enum) types, so the generator must treat enums as `S` the way the index transformer does.
+    const validSchema = `
+      enum Status {
+        ACTIVE
+        ARCHIVED
+      }
+      type Record @model {
+        status: Status! @primaryKey
+        label: String
+      }
+    `;
+    const out = testTransform({
+      schema: validSchema,
+      transformers: [new ModelTransformer(), new PrimaryKeyTransformer()],
+      dataSourceStrategies: {
+        Record: {
+          dbType: 'DYNAMODB' as const,
+          provisionStrategy: 'IMPORTED_AMPLIFY_TABLE' as const,
+          tableName: 'Record-myApiId-myEnv',
+        },
+      },
+    });
+    const table = out.stacks['Record'].Resources?.RecordTable;
+    expect(table).toBeDefined();
+    expect(table.Type).toBe(CUSTOM_IMPORTED_DDB_CFN_TYPE);
+    expect(table.Properties.keySchema).toEqual([{ attributeName: 'status', keyType: 'HASH' }]);
+    expect(table.Properties.attributeDefinitions).toEqual([{ attributeName: 'status', attributeType: 'S' }]);
+    validateModelSchema(parse(out.schema));
+  });
+
   it('keeps the default id key schema for an imported table without a custom @primaryKey', async () => {
     const validSchema = `
       type Note @model {
--- a/packages/amplify-graphql-model-transformer/src/__tests__/amplify-dynamodb-table-generator.test.ts
+++ b/packages/amplify-graphql-model-transformer/src/__tests__/amplify-dynamodb-table-generator.test.ts
@@ -348,6 +348,33 @@ describe('ModelTransformer:', () => {
     validateModelSchema(parse(out.schema));
   });
 
+  it('derives the implicit id key schema for an imported table whose model declares no id field', async () => {
+    // With no @primaryKey (and no explicit id), getPrimaryKeyFieldNodes falls back to the implicit
+    // id: ID! field, so the imported table must still get an `id`/HASH key schema.
+    const validSchema = `
+      type Note @model {
+        content: String
+      }
+    `;
+    const out = testTransform({
+      schema: validSchema,
+      transformers: [new ModelTransformer(), new PrimaryKeyTransformer()],
+      dataSourceStrategies: {
+        Note: {
+          dbType: 'DYNAMODB' as const,
+          provisionStrategy: 'IMPORTED_AMPLIFY_TABLE' as const,
+          tableName: 'Note-myApiId-myEnv',
+        },
+      },
+    });
+    const table = out.stacks['Note'].Resources?.NoteTable;
+    expect(table).toBeDefined();
+    expect(table.Type).toBe(CUSTOM_IMPORTED_DDB_CFN_TYPE);
+    expect(table.Properties.keySchema).toEqual([{ attributeName: 'id', keyType: 'HASH' }]);
+    expect(table.Properties.attributeDefinitions).toEqual([{ attributeName: 'id', attributeType: 'S' }]);
+    validateModelSchema(parse(out.schema));
+  });
+
   it('keeps the default id key schema for an imported table without a custom @primaryKey', async () => {
     const validSchema = `
       type Note @model {
--- a/packages/amplify-graphql-model-transformer/src/__tests__/amplify-dynamodb-table-generator.test.ts
+++ b/packages/amplify-graphql-model-transformer/src/__tests__/amplify-dynamodb-table-generator.test.ts
@@ -7,6 +7,7 @@ import {
 import { parse } from 'graphql';
 import { ModelTransformer } from '../graphql-model-transformer';
 import { SearchableModelTransformer } from '@aws-amplify/graphql-searchable-transformer';
+import { PrimaryKeyTransformer } from '@aws-amplify/graphql-index-transformer';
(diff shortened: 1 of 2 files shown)


soberm (commented):
Reviewed the fix for deriving imported DynamoDB table key schema from `@primaryKey`. The core fix is correct and safely scoped to the imported-table path — no blocking issues. Leaving a few suggestions (test coverage for composite/numeric/enum keys) and nitpicks (DRY/typing).

packages/amplify-graphql-model-transformer/src/__tests__/amplify-dynamodb-table-generator.test.ts
```diff
@@ -171,4 +172,101 @@ describe('ModelTransformer:', () => {
...
+    expect(table).toBeDefined();
+    expect(table.Type).toBe(CUSTOM_IMPORTED_DDB_CFN_TYPE);
+    expect(table.Properties.isImported).toBe(true);
+    // Key schema and attribute definitions must reflect the custom @primaryKey, not `id`.
+    expect(table.Properties.keySchema).toEqual([{ attributeName: 'userAuthenticating', keyType: 'HASH' }]);
+    expect(table.Properties.attributeDefinitions).toEqual([{ attributeName: 'userAuthenticating', attributeType: 'S' }]);
+    validateModelSchema(parse(out.schema));
+  });
+
+  it('derives the imported table key schema from a custom @primaryKey with a sort key', async () => {
```
soberm, line 212: **Suggestion:** No test covers a *composite* (multi-field) sort key on an imported table. The new tests cover no sort key and a single-field sort key, but none exercise `sortKeyFieldNodes.length > 1` (in `amplify-dynamo-model-resource-generator.ts` ~L208–216), which joins field names with `ModelResourceIDs.ModelCompositeKeySeparator()` into one STRING attribute. Multi-field composite sort keys (old `@key(fields: [...])`-style schemas) are common in real Gen 1 apps — arguably more common than the single-sort-key case that is tested. Suggest adding a 4th test like `@primaryKey(sortKeyFields: ["status", "createdAt"])` asserting the `keySchema` RANGE attribute name is `status#createdAt`.

packages/amplify-graphql-model-transformer/src/resources/amplify-dynamodb-table/amplify-dynamo-model-resource-generator.ts
```diff
@@ -172,6 +176,46 @@ export class AmplifyDynamoModelResourceGenerator extends DynamoModelResourceGene
...
+    let partitionKey = {
+      name: 'id',
+      type: AttributeType.STRING,
+    };
+    let sortKey: { name: string; type: AttributeType } | undefined;
+    if (isTableImported) {
+      const [primaryKeyFieldNode, ...sortKeyFieldNodes] = getPrimaryKeyFieldNodes(def);
+      partitionKey = {
+        name: primaryKeyFieldNode.name.value,
+        type: attributeTypeFromScalar(primaryKeyFieldNode.type) === 'N' ? AttributeType.NUMBER : AttributeType.STRING,
```
soberm: **Suggestion:** The numeric (`Int`/`Float`) primary/sort key branch is untested. All three new schemas use `ID`/`String` fields, so `attributeTypeFromScalar(...) === 'N' ? AttributeType.NUMBER : AttributeType.STRING` never exercises the NUMBER branch. A wrong `'S'` vs `'N'` here reproduces exactly the class of bug this PR fixes, for numeric-keyed models. Worth a test with e.g. `@primaryKey` on an `Int` field.

packages/amplify-graphql-model-transformer/src/resources/amplify-dynamodb-table/amplify-dynamo-model-resource-generator.ts
```diff
@@ -172,6 +176,46 @@ export class AmplifyDynamoModelResourceGenerator extends DynamoModelResourceGene
...
+      };
```
soberm: **Suggestion:** (intended location: ~L199–206, the `attributeTypeFromScalar` calls) `attributeTypeFromScalar` throws for enum/Boolean-typed keys. It only handles `String`/`ID`/`Int`/`Float` and throws `Expected scalar and got <Type>` otherwise (explicitly for `Boolean`). But `graphql-primary-key-transformer.ts`'s `validate()` permits `@primaryKey`/sort keys on **enum** fields (`isScalarOrEnum`). So importing a model with an enum-typed key now hits an uncaught synth-time exception with a message that won't point users back to "your primary key type isn't supported for imported tables." Not a regression (pre-fix it failed at deploy time), but a real gap — consider mapping enums to `AttributeType.STRING` explicitly, or wrapping the call with a clearer error.

packages/amplify-graphql-model-transformer/src/resources/amplify-dynamodb-table/amplify-dynamo-model-resource-generator.ts
```diff
@@ -172,6 +176,46 @@ export class AmplifyDynamoModelResourceGenerator extends DynamoModelResourceGene
...
+        sortKey = {
+          name: sortKeyFieldNodes[0].name.value,
+          type: attributeTypeFromScalar(sortKeyFieldNodes[0].type) === 'N' ? AttributeType.NUMBER : AttributeType.STRING,
+        };
+      } else if (sortKeyFieldNodes.length > 1) {
+        // Composite sort keys are stored as a single string attribute whose name is the sort key
+        // field names joined by the model composite key separator (matches `getSortKeyName` in the
+        // index transformer's `replaceDdbPrimaryKey`).
+        sortKey = {
+          name: sortKeyFieldNodes.map((node) => node.name.value).join(ModelResourceIDs.ModelCompositeKeySeparator()),
```
soberm: **Nitpick:** Reinvents an existing helper. `ModelResourceIDs.ModelCompositeAttributeName(names)` (in `graphql-transformer-common`) already does this exact join and is used for the same purpose elsewhere (e.g. `dynamodbUtils.ts`'s `applyCompositeSortKey`). Calling it directly would be more DRY and make the "matches `getSortKeyName`" intent more self-evident than re-deriving the join inline.

packages/amplify-graphql-model-transformer/src/resources/amplify-dynamodb-table/amplify-dynamo-model-resource-generator.ts
```diff
@@ -172,6 +176,46 @@ export class AmplifyDynamoModelResourceGenerator extends DynamoModelResourceGene
...
+      const [primaryKeyFieldNode, ...sortKeyFieldNodes] = getPrimaryKeyFieldNodes(def);
+      partitionKey = {
+        name: primaryKeyFieldNode.name.value,
+        type: attributeTypeFromScalar(primaryKeyFieldNode.type) === 'N' ? AttributeType.NUMBER : AttributeType.STRING,
+      };
+      if (sortKeyFieldNodes.length === 1) {
+        // A single sort key field maps directly to a sort key attribute of the field's scalar type.
+        sortKey = {
+          name: sortKeyFieldNodes[0].name.value,
+          type: attributeTypeFromScalar(sortKeyFieldNodes[0].type) === 'N' ? AttributeType.NUMBER : AttributeType.STRING,
```
soberm: **Nitpick:** Duplicated ternary. `attributeTypeFromScalar(x) === 'N' ? AttributeType.NUMBER : AttributeType.STRING` appears twice verbatim (here and ~L200). A tiny local helper (`const toAttributeType = (t) => ...`) would remove the duplication and give a single place to address the enum/Boolean edge case above.

packages/amplify-graphql-model-transformer/src/resources/amplify-dynamodb-table/amplify-dynamo-model-resource-generator.ts
```diff
@@ -172,6 +176,46 @@ export class AmplifyDynamoModelResourceGenerator extends DynamoModelResourceGene
...
+    // validation, which consumes the initial construct properties directly. If we left the key
+    // schema hardcoded to `id`, any model declaring a custom `@primaryKey` (partition key other
+    // than `id`, and/or a sort key) would fail import with "Imported table properties did not
+    // match the expected table properties". So for imported tables we derive the key schema from
+    // the model definition up front, mirroring how GSIs are already derived from the model.
+    let partitionKey = {
+      name: 'id',
+      type: AttributeType.STRING,
+    };
+    let sortKey: { name: string; type: AttributeType } | undefined;
```
soberm: **Nitpick:** Inline structural type instead of CDK's `Attribute`. `let sortKey: { name: string; type: AttributeType } | undefined;` — `aws-cdk-lib/aws-dynamodb` already exports an `Attribute` interface with exactly this shape (used in `amplify-dynamodb-table-construct/index.ts`). Importing and using `Attribute` would be a touch more self-documenting.

packages/amplify-graphql-model-transformer/src/__tests__/amplify-dynamodb-table-generator.test.ts
sharonyajain: Fixed in fc1261fb3: added a regression test `derives the imported table key schema from a custom @primaryKey with a composite sort key` that exercises the `sortKeyFieldNodes.length > 1` branch and asserts the joined `region#createdAt` RANGE attribute.

packages/amplify-graphql-model-transformer/src/resources/amplify-dynamodb-table/amplify-dynamo-model-resource-generator.ts
sharonyajain: Fixed in fc1261fb3 — added `derives a numeric attribute type for an integer @primaryKey / sort key on an imported table` (an `Int!` partition key + `Int!` sort key), asserting `attributeType: "N"` for both, so the NUMBER branch is now covered.

packages/amplify-graphql-model-transformer/src/resources/amplify-dynamodb-table/amplify-dynamo-model-resource-generator.ts
sharonyajain: Thank you for pointing this out. This was a real correctness bug. `attributeTypeFromScalar` throws on non-scalar (enum) types, and an enum `@primaryKey` is valid (the PrimaryKeyTransformer allows `isScalarOrEnum`), so synth would have crashed. Fixed in fc1261fb3 with an enum-aware `keyAttributeType` helper that returns `S` for enum-typed keys (mirroring the index transformer's `attributeTypeFromType`), plus a regression test `derives a string attribute type for an enum-typed @primaryKey on an imported table`. Boolean keys are already rejected upstream by the PrimaryKeyTransformer, so they cannot reach this path.

packages/amplify-graphql-model-transformer/src/resources/amplify-dynamodb-table/amplify-dynamo-model-resource-generator.ts
sharonyajain: Fixed in fc1261fb3 — now calls `ModelResourceIDs.ModelCompositeAttributeName(...)` (identical `join(ModelCompositeKeySeparator())`, so the attribute name is unchanged) instead of the inline join.

packages/amplify-graphql-model-transformer/src/resources/amplify-dynamodb-table/amplify-dynamo-model-resource-generator.ts
sharonyajain: Fixed in fc1261fb3 — the duplicated `attributeTypeFromScalar(...) === "N" ? ... : ...` is now a single local `keyAttributeType(type)` helper used for both the partition and sort key (and it also carries the enum handling from the comment above).

packages/amplify-graphql-model-transformer/src/resources/amplify-dynamodb-table/amplify-dynamo-model-resource-generator.ts
sharonyajain: Fixed in fc1261fb3 — `partitionKey` and `sortKey` are now typed as the CDK `Attribute` interface from `aws-cdk-lib/aws-dynamodb` instead of the inline structural type.

sharonyajain:
Addressed the latest review batch in `2b27d4201`. Dispositions:

**Applied**

- Extracted the key derivation into a private `getImportedTableKeySchema` method plus a `keyAttributeType` helper; `createModelTable` is now a single call. No behavior change.
- Added `derives the implicit id key schema for an imported table whose model declares no id field` (a `type Note @model { content: String }` with no `id` and no `@primaryKey`), covering the `getImplicitlyDefinedIdField` fallback. File now has 11 tests; full `graphql-model-transformer` suite is 207 tests / 87 snapshots green, `tsc` clean.

soberm (approved):

Issue #3162, opened 2025-02-12
Amplify AI Access Denied: Missing bedrock:InvokeModelWithResponseStream Permission


mnai01:
### Environment information

```plain text
System:
  OS: Windows 10 10.0.19045
  CPU: (8) x64 Intel(R) Core(TM) i7-4770K CPU @ 3.50GHz
  Memory: 10.04 GB / 31.95 GB
Binaries:
  Node: 20.11.1 - F:\Program Files\nodejs\node.EXE     
  Yarn: undefined - undefined
  npm: 10.2.4 - F:\Program Files\nodejs\npm.CMD        
  pnpm: undefined - undefined
NPM Packages:
  @aws-amplify/auth-construct: 1.6.0
  @aws-amplify/backend: 1.14.0
  @aws-amplify/backend-auth: 1.5.0
  @aws-amplify/backend-cli: 1.4.9
  @aws-amplify/backend-data: 1.4.0
  @aws-amplify/backend-deployer: 1.1.16
  @aws-amplify/backend-function: 1.12.1
  @aws-amplify/backend-output-schemas: 1.4.0
  @aws-amplify/backend-output-storage: 1.1.4
  @aws-amplify/backend-secret: 1.1.6
  @aws-amplify/backend-storage: 1.2.4
  @aws-amplify/cli-core: 1.2.3
  @aws-amplify/client-config: 1.5.6
  @aws-amplify/deployed-backend-client: 1.5.0
  @aws-amplify/form-generator: 1.0.3
  @aws-amplify/model-generator: 1.0.12
  @aws-amplify/platform-core: 1.6.1
  @aws-amplify/plugin-types: 1.8.0
  @aws-amplify/sandbox: 1.2.11
  @aws-amplify/schema-generator: 1.2.7
  aws-amplify: 6.12.3
  aws-cdk: 2.178.1
  aws-cdk-lib: 2.178.1
  typescript: 5.7.3
No AWS environment variables
No CDK environment variables
```

### Data packages

```plain text
chrome-extension-boilerplate-react@5.0.4 F:\Programming\Javascript\React\chrome-extension-boilerplate-react
├─┬ @aws-amplify/backend-cli@1.4.9
│ └─┬ @aws-amplify/schema-generator@1.2.7
│   └── @aws-amplify/graphql-schema-generator@0.11.7
└─┬ @aws-amplify/backend@1.14.0
  └─┬ @aws-amplify/backend-data@1.4.0
[...]

PR #3530, opened 2026-09-10
fix(generation-transformer): grant inference-profile IAM for cross-region Bedrock models

### LINKED ISSUE #3502: a.generation() Bedrock IAM role missing inference-profile permission (newer Claude models fail with AccessDeniedException)
## Describe the bug

When using `a.generation()` with newer Claude models (e.g. `Claude Haiku 4.5`, `Claude Sonnet 4`), the auto-generated AppSync IAM role only grants `bedrock:InvokeModel` on the **foundation model ARN**, but AWS routes requests through a **cross-region inference profile** (`global.*` or `us.*` prefix). This results in an `AccessDeniedException` at runtime.

## Error message

```
User: arn:aws:sts::263556453558:assumed-role/GenerationBedrockDat.../APPSYNC_ASSUME_ROLE
is not authorized to perform: bedrock:InvokeModel on resource:
arn:aws:bedrock:us-west-2:263556453558:inference-profile/global.anthropic.claude-haiku-4-5-20251001-v1:0
because no identity-based policy allows the bedrock:InvokeModel action
```

## Root cause

In `grapqhl-generation-transformer.js`, `createBedrockDataSourceRole()` only adds a policy for the foundation model ARN:

```js
resources: [`arn:${partition}:bedrock:${region}::foundation-model/${bedrockModelId}`]
```

It does not include the account-scoped inference profile ARN (`arn:aws:bedrock:*:ACCOUNT:inference-profile/*`), which AWS requires when routing through cross-region inference.

This is the same root cause as #3162 (which affects `a.conversation()` + `InvokeModelWithResponseStream`) but affects `a.generation()` + `InvokeModel`.

## Workaround (CDK escape hatch in `backend.ts`)

```ts
import { Role } from "aws-cdk-lib/aws-iam"

backend.data.node.findAll().forEach((construct) => {
  if (
    construct instanceof Role &&
    construct.node.id.startsWith("GenerationBedrockDataSource") &&
    construct.node.id.endsWith("IAMRole")
  ) {
    construct.addToPolicy(new PolicyStatement({
      effect: Effect.ALLOW,
      actions: ["bedrock:InvokeModel"],
      resources: [
        `arn:aws:bedrock:*:${backend.stack.account}:inference-profile/*`,
        `arn:aws:bedrock:*::foundation-model/anthropic.*`,
      ],
    }))
  }
})
```

## Expected behavior

The auto-generated role should include both foundation model and inference profile ARNs, consistent with what is needed for current Bedrock model routing behavior.

## Related issues

- #3162
- aws-amplify/docs#8121
- #3218

sharonyajain:
#### Description of changes

When a `@generation` route uses one of the newer Claude models (e.g. `Claude Haiku 4.5`, `Claude Sonnet 4.5`), the auto-generated AppSync Bedrock IAM role only granted `bedrock:InvokeModel` on the **foundation-model ARN**. These models resolve (via `@aws-amplify/data-schema`'s `supportedModelsLookup`) to a **cross-region inference profile** id such as `global.anthropic.claude-haiku-4-5-...`, which AWS routes through an **inference-profile ARN** that the generated policy never listed. The result is an `AccessDeniedException` at runtime:

```
User: arn:aws:sts::<acct>:assumed-role/GenerationBedrockDat.../APPSYNC_ASSUME_ROLE
is not authorized to perform: bedrock:InvokeModel on resource:
arn:aws:bedrock:us-west-2:<acct>:inference-profile/global.anthropic.claude-haiku-4-5-20251001-v1:0
because no identity-based policy allows the bedrock:InvokeModel action
```

**Root cause:** `createBedrockDataSourceRole()` in `grapqhl-generation-transformer.ts` built a single policy statement whose only resource was `arn:${partition}:bedrock:${region}::foundation-model/${bedrockModelId}`, interpolating the raw model id. For an inference-profile id, that ARN neither *is* nor *covers* the `inference-profile/*` ARN the request is actually made against.

**Fix:** a new exported helper `bedrockInvokeModelResources(partition, region, account, bedrockModelId)` computes the correct resource list:
- **Plain foundation model** (e.g. `anthropic.claude-3-haiku-...`) → foundation-model ARN only. **Unchanged behavior.**
- **Cross-region inference profile** (prefix `global.` / `us.` / `eu.` / `apac.` / `ap.`) → the account-scoped `inference-profile/*` ARN **plus** the underlying regional foundation-model ARN; for `global.`, the partition-wide foundation-model ARN is added too (global profiles can route to any region).

This mirrors the three-part grant that `@aws-amplify/ai-constructs` already uses for the `a.conversation()` path (the fix for the analogous `InvokeModelWithResponseStream` denial). The `a.generation()` path was missing the equivalent grant.

> Note: this supersedes and extends the earlier attempt in #3495, which was closed as superseded when #3500 switched the `@generation` E2E test model to an ON_DEMAND foundation model (greening the shard without fixing the product code). This change also covers the `global.` prefix that #3495's `us.`/`eu.`/`apac.` regex did not — and `global.` is exactly what the newer Claude 4.5/4.6 models resolve to.

None. This only widens the resource list on an existing inline IAM policy statement of the auto-generated Bedrock data source role; no CDK/CFN parameters are added or changed.

#### Issue #, if available

Fixes #3502
Related: #3162 (same root cause on the `a.conversation()` path)

- **Unit tests** (`bedrock-model-resources.test.ts`, added): assert the exact ARN list for a plain foundation model, each cross-region prefix (`global.`/`us.`/`eu.`/`apac.`/`ap.`), and a non-default partition. Verified passing locally.
- **Integration test** (added to `amplify-graphql-generation-transformer.test.ts`): transforms a `@generation` route whose `aiModel` is a `global.` inference profile and asserts the generated IAM role's policy statements include the `inference-profile/...` resource. This fails against the pre-fix single-ARN code, so it is a genuine regression guard.
- The plain-foundation-model paths are asserted to be byte-for-byte unchanged.

#### Checklist

- [x] PR description included
- [ ] `yarn test` passes 
- [ ] E2E test run linked


svidgen (approved):
Looks reasonable. It would be nice to add some e2e coverage as well; but probably not critical.

Did you do happen to do any manually testing?

sharonyajain:
Yes. I ran a live end-to-end repro on a sandbox to confirm both the failure and the fix.

**Repro of the bug (against current `main`)** — deployed a minimal `a.generation({ aiModel: a.ai.model('Claude Haiku 4.5') })` backend on `@aws-amplify/backend@1.25.0`, which bundles the unfixed generation-transformer. `Claude Haiku 4.5` resolves to `global.anthropic.claude-haiku-4-5-20251001-v1:0` — a cross-region inference profile — but the deployed data-source role grants `bedrock:InvokeModel` only on `arn:<partition>:bedrock:<region>::foundation-model/global.anthropic.claude-haiku-4-5-20251001-v1:0` (the inference-profile model id stuffed into a `foundation-model/` ARN). An authenticated `generateHaiku` query then fails:

> `errorType: "AccessDeniedException"` — "The model is disabled or this generation route is missing a necessary identity-based policy."

because AppSync actually invokes the `inference-profile/*` ARN, which the foundation-model-only grant doesn't cover.

**After this change**, `bedrockInvokeModelResources` grants the account-scoped `inference-profile/*` ARN plus the regional (and, for `global.`, partition-wide) `foundation-model/*` ARNs — matching the three-part grant `@aws-amplify/ai-constructs` already uses for `a.conversation()` (see #3162). The coverage is also locked in by the integration test in this PR, which fails against the pre-fix single-ARN code.

